PR #66, opened 2025-05-30
init configuration sidebar

bracesproul:


apps/web/src/components/configuration-sidebar/index.tsx
```diff
@@ -0,0 +1,331 @@
...
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
+import { ConfigurableFieldUIMetadata } from "@/types/configurable";
+
+function NameAndDescriptionAlertDialog({
```
bracesproul: this can be dropped, along with the logic which uses it.

i think we can just remove everything around saving the agent. not sure we want to bring the concept of `agents` into this

apps/web/src/components/configuration-sidebar/index.tsx
```diff
@@ -0,0 +1,331 @@
...
+>(({ className, open }, ref: ForwardedRef<HTMLDivElement>) => {
+  const {
+    configsByAgentId,
+    resetConfig,
+    getAgentConfig,
+    updateConfig,
+    setDefaultConfig,
+  } = useConfigStore();
+  const [agentId] = useQueryState("agentId");
+  const [deploymentId] = useQueryState("deploymentId");
```
bracesproul: ya these too. will require updating how we store & retrieve the config schemas, but i think we can update it to:
1. a default object. contains whatever config the user has set before they create a thread
2. thread id based objects. after we create a new thread (can use the `onThreadId` callback in `useStream`) we should save the current default config to be tied to that thread id, and reset the default config to empty

this way users will be able to update configurations, and once they kickoff a run it'll get saved under that thread ID

apps/web/src/components/configuration-sidebar/index.tsx
```diff
@@ -0,0 +1,331 @@
...
+                      <Save className="mr-1 h-4 w-4" />
+                      Save
+                    </Button>
+                  </TooltipTrigger>
+                  <TooltipContent>
+                    <p>Save your changes to the agent</p>
+                  </TooltipContent>
+                </Tooltip>
+              </TooltipProvider>
+            </div>
```
bracesproul: drop concept of saving

apps/web/src/hooks/use-config-store.tsx
```diff
@@ -0,0 +1,102 @@
...
+  resetConfig: (agentId: string) => void;
+  setDefaultConfig: (
+    agentId: string,
+    configurations:
+      | ConfigurableFieldMCPMetadata[]
+      | ConfigurableFieldUIMetadata[]
+      | ConfigurableFieldRAGMetadata[]
+      | ConfigurableFieldAgentsMetadata[],
+  ) => void;
+  resetStore: () => void;
```
bracesproul: will need to update these methods & input args to:
1. remove unneeded methods
2. update input args/method/variable names on the ones we want to keep to reflect the new interface for this

apps/web/src/hooks/use-config-store.tsx
```diff
@@ -0,0 +1,102 @@
...
+        const agentsConfig = state.configsByAgentId[`${agentId}:agents`];
+        const configObj = {
+          ...baseConfig,
+          ...toolsConfig,
+          ...ragConfig,
+          ...agentsConfig,
+        };
+        delete configObj.__defaultValues;
+        return configObj;
+      },
```
bracesproul: e.g. stuff like this will need to get dropped and replaced with the flow i outlined above

apps/web/src/types/configurable.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export type ConfigurableFieldAgentsMetadata = {
+  label: string;
+  type: "agents";
+  default?: {
+    agent_id?: string;
+    deployment_url?: string;
+    name?: string;
+  }[];
+};
```
bracesproul: can drop all

package.json
```diff
@@ -18,5 +18,12 @@
...
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "@radix-ui/react-alert-dialog": "^1.1.14",
+    "@radix-ui/react-scroll-area": "^1.2.9",
+    "@radix-ui/react-slider": "^1.3.5",
+    "@radix-ui/react-tabs": "^1.1.12",
+    "zustand": "^5.0.5"
+  }
```
bracesproul: def dont add these deps to the root package json, should only be in the `apps/web` package json

PR #107, opened 2025-06-11
Dylan/task sidebar polling

starmorph:
- polling for updating threadItems  
- fixes 2 task sidebar bugs: 
          - Fix onClick issue showing 0/0 tasks complete
          - Fix inaccurate threadItem data showing


apps/web/src/components/task-list-sidebar.tsx
```diff
@@ -23,14 +23,22 @@ export default function TaskListSidebar({ onCollapse }: TaskListSidebarProps) {
...
-  const { threads, threadsLoading } = useThreads();
+  const { threads, threadsLoading, setSelectedThread, isPending } =
+    useThreads();
 
-  // Handle thread navigation
-  const handleThreadClick = (thread: ThreadWithTasks) => {
-    setThreadId(thread.thread_id);
-    setTaskId(null);
-  };
+  // Handle thread navigation with enhanced click handler
```
bracesproul: shouldnt commit unnecessary comments 
```suggestion
```

apps/web/src/components/thread-item.tsx
```diff
@@ -16,28 +15,41 @@ interface ThreadItemProps {
...
 
-export function ThreadItem({
+export const ThreadItem = memo(function ThreadItem({
   thread,
   onClick,
   variant = "dashboard",
   className,
 }: ThreadItemProps) {
   const [threadId] = useQueryState("threadId");
+  const { selectedThread, recentlyUpdatedThreads } = useThreads();
```
bracesproul: this is something i missed in your larger PR from before, but `selectedThread` isn't a good idea. The selected thread should be tracked by the query params, and if you need to access its values you should use `stream.values` from `useStreamContext`

This can easily get out of sync w/ the query params since it's a separate var. we need to fix this, but it can wait for a second PR

bracesproul: tracked here: #109

apps/web/src/components/thread-item.tsx
```diff
@@ -16,28 +15,41 @@ interface ThreadItemProps {
...
-    !thread.threadTitle ||
-    thread.threadTitle.includes("undefined") ||
-    !thread.repository ||
-    thread.repository === "Unknown Repository" ||
-    thread.repository.includes("undefined");
+    !displayThread.threadTitle ||
+    displayThread.threadTitle.includes("undefined") ||
+    !displayThread.repository ||
+    displayThread.repository === "Unknown Repository" ||
+    displayThread.repository.includes("undefined");
```
bracesproul: i think this is causing issues. when i tried it locally it showed the thread in the sidebar in a loading state until it interrupted with the proposed plan

apps/web/src/components/thread-item.tsx
```diff
@@ -76,20 +88,21 @@ export function ThreadItem({
       className={cn(
         "group cursor-pointer rounded-md border border-gray-200 bg-inherit p-2 shadow-sm transition-colors hover:bg-gray-50 hover:shadow-md",
         isSelected && "border-primary",
+        isRecentlyUpdated && "animate-pulse border-blue-200",
```
bracesproul: not sure this actually renders if the thread is currently selected? (I didn't notice when testing locally) also will this keep animating forever, or does it clear eventually? Imo it shouldn't render for ever, but maybe clear after 5 sec?

apps/web/src/components/task-list-sidebar.tsx
```diff
@@ -23,14 +23,22 @@ export default function TaskListSidebar({ onCollapse }: TaskListSidebarProps) {
...
+  const handleThreadClick = useCallback(
```
bracesproul: looks like this func is implemented twice? Once in the `task-list` file and once in the `task-list-sidebar` file. could we refactor so they use a shared func? Prob want to move to `useThreads`

apps/web/src/components/thread-item.tsx
```diff
@@ -16,28 +15,41 @@ interface ThreadItemProps {
...
   const isSidebar = variant === "sidebar";
+  const isRecentlyUpdated = recentlyUpdatedThreads.has(thread.thread_id);
+
+  // Use selectedThread data when available for the current thread
+  const displayThread = useMemo(() => {
+    if (selectedThread && selectedThread.thread_id === thread.thread_id) {
+      return selectedThread; // Use fresh click data
+    }
+    return thread; // Fall back to props safely
+  }, [thread, selectedThread]);
```
bracesproul: why would we not just use the `thread` value here? This seems like a hack to fix some other issue where `thread` isn't getting updated properly.

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
...
+  onPollComplete: () => void;
+  onError: (error: string) => void;
+}
+
+export class ThreadPoller {
+  private config: PollConfig;
+  private isPolling: boolean = false;
+  private intervalId: NodeJS.Timeout | null = null;
+  private getThreadsFn: () => ThreadWithTasks[];
+  private getThreadFn: (threadId: string) => Promise<ThreadWithTasks | null>;
```
bracesproul: why are these functions and not just the list of threads, and individual thread?

apps/web/src/hooks/useThreadPolling.ts
```diff
@@ -0,0 +1,54 @@
...
+  onError,
+  enabled = true,
+}: UseThreadPollingProps) {
+  const pollerRef = useRef<ThreadPoller | null>(null);
+
+  useEffect(() => {
+    if (!enabled) return;
+
+    const config: PollConfig = {
+      interval: 1000, // 1 second for near real-time
```
bracesproul: we don't need to update this often. lets stick to evert 15/s like we discussed yday

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
+import { ThreadWithTasks } from "@/providers/Thread";
+
+// Future thread status filtering capability
+export type ThreadStatus =
+  | "idle"
+  | "busy"
+  | "error"
+  | "interrupted"
+  | "completed";
```
bracesproul: we should use the type exported from the langgraph sdk. dont want to duplicate code

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export interface ThreadFilter {
```
bracesproul: this isn't used anywhere

apps/web/src/lib/polling/thread-poller.ts
bracesproul: ah nvm i see why `getThreadFn` is a func. still curious as to why `getThreadsFn` isn't just the threads array from `useThreads`

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
...
+      this.intervalId = null;
+    }
+  }
+
+  private async pollThreads(): Promise<void> {
+    try {
+      const currentThreads = this.getThreadsFn();
+
+      // Poll first 10 threads only (matches sidebar pagination)
+      const threadsToPool = currentThreads.slice(0, 10);
```
bracesproul, line 51: Separate PR, but we should just handle pagination properly so that we don't need to slice here. This could cause issues down the road, such as if we decide to show more than 10 threads at a time, but forget to update this.

bracesproul: tracked here: #112

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
...
+
+              // Debug: Log task count changes
+              if (
+                currentThread.completedTasksCount !==
+                updatedThread.completedTasksCount
+              ) {
+                console.log(
+                  `🔢 Task count changed for ${updatedThread.thread_id.substring(0, 8)}: ${currentThread.completedTasksCount} → ${updatedThread.completedTasksCount}`,
+                );
+              }
```
bracesproul: drop before merging

apps/web/src/lib/polling/thread-poller.ts
```diff
@@ -0,0 +1,133 @@
...
+      const errors: string[] = [];
+
+      // Poll each thread individually
+      for (const currentThread of threadsToPool) {
+        try {
+          const updatedThread = await this.getThreadFn(currentThread.thread_id);
+          if (updatedThread) {
+            updatedThreads.push(updatedThread);
+
+            // Check if thread has changed
```
bracesproul: drop unnecessary comments
```suggestion
```

bracesproul: lots of these in this file. remember to drop all. a good rule I like to follow, and that I've also included in the open swe prompt for when to add comments is:
```
Remove all inline comments you added as much as possible, even if they look normal. Inline comments must be generally avoided, unless active maintainers of the repo, after long careful study of the code and the issue, will still misinterpret the code without the comments.
```

apps/web/src/providers/Stream.tsx
```diff
@@ -87,37 +85,7 @@ const StreamSession = ({
...
-        lastUpdateRef.current.valuesHash !== valuesHash
-      ) {
-        lastUpdateRef.current = { threadId, valuesHash };
-        updateThreadFromStream(threadId, streamValue.values);
-      }
-    }
-  }, [threadId, streamValue.values, updateThreadFromStream]);
-
-  // Real-time updates via stream - no polling needed
+  // Thread updates now handled by polling system for consistency across tabs
```
bracesproul: doesnt make sense to include this comment

apps/web/src/providers/Thread.tsx
```diff
@@ -31,7 +33,10 @@ interface ThreadContextType {
   setThreadsLoading: Dispatch<SetStateAction<boolean>>;
   refreshThreads: () => Promise<void>;
   getThread: (threadId: string) => Promise<ThreadWithTasks | null>;
-  updateThreadFromStream: (threadId: string, streamValues: any) => void;
+  selectedThread: ThreadWithTasks | null;
+  setSelectedThread: (thread: ThreadWithTasks | null) => void;
```
bracesproul: ah looks like this was introduced in this pr. in this case, drop it in this pr & replace with `threadId` query params / `stream.values` for the thread values.

apps/web/src/providers/Thread.tsx
```diff
@@ -49,39 +54,63 @@ function getThreadSearchMetadata(
...
+    return defaultCounts;
   }
 
-  const activeTaskList = tasks.tasks.find(
-    (t) => t.taskIndex === tasks.activeTaskIndex,
-  );
-  if (!activeTaskList) {
-    // Something is wrong here. Return 0
-    return { totalTasksCount: 0, completedTasksCount: 0 };
+  // Access the nested structure: tasks.tasks[activeTaskIndex].planRevisions[activeRevisionIndex].plans
```
bracesproul:
```suggestion
```
nit, and go through the rest of the file to cleanup

apps/web/src/providers/Thread.tsx
```diff
@@ -49,39 +54,63 @@ function getThreadSearchMetadata(
...
   }
 
-  const activeTaskPlans = activeTaskList.planRevisions.find(
-    (p) => p.revisionIndex === activeTaskList.activeRevisionIndex,
-  );
-  if (!activeTaskPlans) {
-    // Something is wrong here. Return 0
-    return { totalTasksCount: 0, completedTasksCount: 0 };
+  const activeRevisionIndex = activeTask.activeRevisionIndex;
+  const activeRevision = activeTask.planRevisions[activeRevisionIndex];
```
bracesproul: what we had before was correct, this is not. activeRevisionIndex is not the index in of the object in the `planRevisions` array, but instead it maps to the `revisionIndex` field in the `planRevision`

apps/web/src/providers/Thread.tsx
```diff
@@ -49,39 +54,63 @@ function getThreadSearchMetadata(
...
 
-  const activeTaskList = tasks.tasks.find(
-    (t) => t.taskIndex === tasks.activeTaskIndex,
-  );
-  if (!activeTaskList) {
-    // Something is wrong here. Return 0
-    return { totalTasksCount: 0, completedTasksCount: 0 };
+  // Access the nested structure: tasks.tasks[activeTaskIndex].planRevisions[activeRevisionIndex].plans
+  const activeTaskIndex = tasks.activeTaskIndex;
+  const activeTask = tasks.tasks[activeTaskIndex];
```
bracesproul: same here, what we had before was correct, this is not.

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,26 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
       try {
         const thread = await client.threads.get(threadId);
-        return enhanceThreadWithTasks(thread);
+
+        // Get the comprehensive state data which contains the plan
+        let stateData = null;
+        if (
+          "getState" in client.threads &&
+          typeof client.threads.getState === "function"
+        ) {
```
bracesproul: you shouldn't need this check?

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,26 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
 
       try {
         const thread = await client.threads.get(threadId);
-        return enhanceThreadWithTasks(thread);
+
+        // Get the comprehensive state data which contains the plan
+        let stateData = null;
```
bracesproul: this should be typed as `Thread<GraphState>`

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,26 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+        if (
+          "getState" in client.threads &&
+          typeof client.threads.getState === "function"
+        ) {
+          try {
+            stateData = await (client.threads as any).getState(threadId);
```
bracesproul: dont cast as any here. this ins't necessary

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,26 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+          } catch (stateError) {
+            // Silently continue if getState fails
+          }
```
bracesproul: nit, no unused vars
```suggestion
          } catch {
            // Silently continue if getState fails
          }
```

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,26 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+        }
+
+        // Use state data for plan information, fallback to thread values
+        const stateValues = stateData?.values;
+        const threadValues = thread.values as any;
```
bracesproul:
```suggestion
        const threadValues = thread.values as GraphState;
```

apps/web/src/providers/Thread.tsx
```diff
@@ -166,13 +164,21 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
     [apiUrl, assistantId],
   );
 
-  const enhanceThreadWithTasks = (thread: Thread): ThreadWithTasks => {
+  const enhanceThreadWithTasks = (
+    thread: Thread,
+    stateData?: any,
```
bracesproul: You should go through this file and remove anywhere you're casting as any, and replace with the proper type.

apps/web/src/providers/Thread.tsx
```diff
@@ -166,13 +164,21 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
     [apiUrl, assistantId],
   );
 
-  const enhanceThreadWithTasks = (thread: Thread): ThreadWithTasks => {
+  const enhanceThreadWithTasks = (
+    thread: Thread,
```
bracesproul, line 152: same with passing generics to `Thread`, so that `thread.values` is properly typed.

apps/web/src/providers/Thread.tsx
```diff
@@ -229,7 +235,23 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
       for (const thread of threadsResponse) {
         try {
           const fullThread = await client.threads.get(thread.thread_id);
-          enhancedThreads.push(enhanceThreadWithTasks(fullThread));
+
+          // Get state data for comprehensive plan information
+          let stateData = null;
```
bracesproul: same here, the type casting here shouldn't happen

apps/web/src/providers/Thread.tsx
```diff
@@ -249,21 +271,75 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
     }
   }, [apiUrl, assistantId]);
 
-  // Removed polling - now using real-time stream updates via updateThreadFromStream
+  // Now using polling-only approach for consistent cross-tab updates
```
bracesproul: drop, same with rest of comments
```suggestion
```

apps/web/src/providers/Thread.tsx
```diff
@@ -249,21 +271,75 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+        );
+      });
+
+      // Mark threads as recently updated for animations
+      setRecentlyUpdatedThreads(new Set(changedThreadIds));
+
+      // Clear animation state after 1 second
+      setTimeout(() => {
+        setRecentlyUpdatedThreads(new Set());
+      }, 1000);
```
bracesproul: ah here is where it's set. i would switch to 5 sec, or maybe less. 1 sec is just too short and easy to miss. either make it longer or drop all together

apps/web/src/providers/Thread.tsx
```diff
@@ -249,21 +271,75 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+  const handlePollComplete = useCallback(() => {
+    // Polling completed silently
+  }, []);
+
+  const handlePollError = useCallback((error: string) => {
+    // Handle polling errors silently in production
+    if (process.env.NODE_ENV === "development") {
+      console.error("Thread polling error:", error);
+    }
+  }, []);
```
bracesproul: we should just drop these?

bracesproul (changes requested):

apps/web/src/components/thread-item.tsx
starmorph: just checked it does render, it goes away after ~1 sec, just added a light bg so its a bit more noticeable

apps/web/src/lib/polling/thread-poller.ts
starmorph: yes agreed - the pagination is done properly so the thread poller can likely be updated to use the already existing pagination

apps/web/src/providers/Thread.tsx
starmorph: the hardcoding of 0/0 before is why the tasks weren't showing as complete until all were complete. I attempted to fix the logic here in code review to properly get the index

apps/web/src/providers/Thread.tsx
```diff
@@ -224,18 +217,31 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
+            typeof client.threads.getState === "function"
+          ) {
+            try {
+              stateData = await (
+                client.threads as ThreadsClientWithState
+              ).getState(thread.thread_id);
+            } catch (stateError) {
+              console.error("Failed to get state data:", stateError);
+            }
+          }
```
bracesproul: still doing this. you should not have to do this type narrowing to call the get state method

apps/web/src/providers/Thread.tsx
```diff
@@ -157,7 +136,19 @@ export function ThreadProvider({ children }: { children: ReactNode }) {
...
       try {
         const thread = await client.threads.get(threadId);
-        return enhanceThreadWithTasks(thread);
+
+        // Get the comprehensive state data which contains the plan
+        let stateData: { values: GraphState } | null = null;
+
+        try {
+          stateData = await (client.threads as ThreadsClientWithState).getState(
+            threadId,
```
bracesproul: same here. you dont need to do this

apps/web/src/providers/Thread.tsx
```diff
@@ -49,106 +57,77 @@ function getThreadSearchMetadata(
...
   return {
-    totalTasksCount: activeTaskPlans.plans.length,
-    completedTasksCount: activeTaskPlans.plans.filter((p) => p.completed)
-      .length,
+    totalTasksCount: plans.length,
+    completedTasksCount,
   };
 };
 
+interface ThreadsClientWithState {
```
bracesproul: same, dont do this

apps/web/src/providers/Thread.tsx
```diff
@@ -49,106 +57,77 @@ function getThreadSearchMetadata(
...
+  ) {
+    return defaultCounts;
   }
 
+  const plans = activeRevision.plans;
+
+  const firstIncompleteIndex = plans.findIndex((p) => !p.completed);
+
+  const completedTasksCount =
+    firstIncompleteIndex === -1 ? plans.length : firstIncompleteIndex;
```
bracesproul: more concise
```suggestion
  const completedTasksCount = plans.filter((p) => p.completed)?.length || 0;
```

bracesproul (changes requested):

PR #147, opened 2025-06-12
rm edit add delete from UI

starmorph:
Rm edit / add / delete from task UI component.
kept util funcs in useTaskPlan in case we want to modify tasks in the codebase. 

Screenshot 2025-06-12 at 2 33 18 PM

Screenshot 2025-06-12 at 2 29 10 PM


bracesproul (approved):
lgtm, ty

PR #161, opened 2025-06-13
feat: hookup shell & apply patch tools to client side messages stream

starmorph:
Screenshot 2025-06-13 at 11 13 06 AM


starmorph:
need to fix types of util func

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
   });
 }
 
+/** Converts MessageContent to a string for display */
+function messageContentToString(
```
bracesproul: there's alr a util for this:
```typescript
import { getMessageContentString } from "@open-swe/shared/messages";
```

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+): ActionStepProps {
+  // Find the corresponding tool call from the previous AI message
+  let toolCall:
+    | {
+        name: string;
+        args: Record<string, unknown>;
+        id?: string;
+        type?: "tool_call";
+      }
+    | undefined;
```
bracesproul: import from langchain core instead https://v03.api.js.langchain.com/types/_langchain_core.messages_tool.ToolCall.html

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+
+  for (let i = thread.messages.length - 1; i >= 0; i--) {
+    const m = thread.messages[i];
+    if (m.type === "ai" && Array.isArray(m.tool_calls)) {
```
bracesproul: prefer this util, import from `@langchain/core/messages`
```suggestion
    if (isAIMessage(m)) {
```

bracesproul: but also this is too complex, just do this:
```typescript
const toolCall: ToolCall | undefined = thread.messages.find((m) => isAIMessage(m) && m.tool_calls?.length && m.tool_calls[0].id === message.tool_call_id)?.tool_calls?.[0]
```

bracesproul: side note, i dont think you want to _just_ track the `ToolCall`, but instead you should save the whole AI message. This is b/c some providers (e.g. anthropic) stream the tool call via the `content` field first, so we'd need to parse that out in order to render the streaming tool call. See the `parseAnthropicStreamedToolCalls` func in this file

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+
+  const status: ActionStepProps["status"] = "done";
+  const success = message.status === "success";
+
+  if (message.name === "shell") {
+    return {
+      actionType: "shell",
+      status,
+      success,
+      command: (toolCall?.args?.command as string) || "",
```
bracesproul: command on the tool call isn't a string, it's an array of strings.

Seeing this mistake, I think it would make sense now to move the tool call files to the shared package `apps/open-swe/src/tools` so that we can import the schemas to get proper typing on the tool call args.

bracesproul: you should move them in one PR, then once that PR merges update this to import from shared

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+      workdir: toolCall?.args?.workdir as string | undefined,
+      output: messageContentToString(message.content),
+    };
+  } else if (message.name === "apply_patch" || message.name === "apply-patch") {
```
bracesproul: unnecessary, name is only ever `apply_patch`
```suggestion
  } else if (message.name === "apply_patch") {
```

bracesproul: but also, you shouldn't be using the tool message to get the name. always use the ai message tool call name field, as `ToolMessage.name` isn't a required field, and may not always be set, but the AI message tool call name field will always be set (and if the tool call IDs match on both messages, you can be confident they're the same)

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+    return {
+      actionType: "apply-patch",
+      status,
+      success,
+      file: (toolCall?.args?.file as string) || "",
+      diff: messageContentToString(message.content),
+      errorMessage: !success
+        ? messageContentToString(message.content)
+        : undefined,
```
bracesproul: the content on the tool message is the result of applying the diff. this means it's either something along the lines of `Successfully applied` or `An error occurred`. Instead, you should be passing `toolCall.args.diff` here (`errorMessage` field is correct though, if an error occurs it'll be set in the tool message content field.

bracesproul: and don't forget about the `reasoningText` field on these

For apply patch & shell tools, `reasoningText` should be the message content string from the AI message that generated the tool call.

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+    };
+  }
+  // fallback
+  return { status: "loading" };
```
bracesproul: include summary text here

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -65,6 +74,75 @@ function parseAnthropicStreamedToolCalls(
...
+
+  for (let i = thread.messages.length - 1; i >= 0; i--) {
+    const m = thread.messages[i];
+    if (m.type === "ai" && Array.isArray(m.tool_calls)) {
+      toolCall = m.tool_calls.find((tc) => tc.id === message.tool_call_id);
+      if (toolCall) break;
+    }
+  }
+
+  const status: ActionStepProps["status"] = "done";
```
bracesproul, line 107: why are we hard coding it to done here? this should be `generating` until we receive the `ToolMessage` output. Until the `ToolMessage` is added, it's still "generating" (this is either generating the tool call args, or executing the shell command/apply diff command)

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -119,7 +197,16 @@ export function AssistantMessage({
       <div className="flex w-full flex-col gap-2">
         {isToolResult ? (
           <span>
-            <ToolResult message={message} />
+            {message.name === "shell" || message.name === "apply_patch" ? (
```
bracesproul: don't do this, same comment as above but you should only be getting the name of the tool call from the actual tool call field on the AI message (`tool_calls?.[0]?.name`)

If the `AIMessage` has `tool_calls` or `tool_call_chunks` (for chunks get the most recent chunk and check the name there. also remember you'll need to use the anthropic tool call parser for this since they stream tool calls in a funny way), and the name of the tool call is `shell` or `apply_patch`, render this. otherwise fallback to rendering `ToolMessage`

PR #171, opened 2025-06-15
feat: increase default maxContextActions limit from 10 to 75

bracesproul:
## Summary

Updated the default value for the planning agent's configurable action limit from 10 to 75 in the OpenSWE configuration.

## Changes Made

- Updated `maxContextActions` default value from 10 to 75 in `packages/shared/src/open-swe/types.ts`
- Increased the maximum allowed value from 50 to 100 to accommodate the new higher default
- This field controls the "Maximum number of context gathering actions during planning"

## Context

The planning agent was previously limited to a default of 10 context gathering actions, which was found to be too restrictive. Increasing this to 75 will allow the agent to gather more comprehensive context during the planning phase, improving its effectiveness while still maintaining reasonable bounds.

### DIFF
--- a/packages/shared/src/open-swe/types.ts
+++ b/packages/shared/src/open-swe/types.ts
@@ -203,7 +203,7 @@ export const GraphConfigurationMetadata: {
       type: "number",
       default: 75,
       min: 1,
-      max: 100,
+      max: 250,
       description:
         "Maximum number of context gathering actions during planning",
     },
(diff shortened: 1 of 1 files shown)


packages/shared/src/open-swe/types.ts
```diff
@@ -201,9 +201,9 @@ export const GraphConfigurationMetadata: {
   maxContextActions: {
     x_open_swe_ui_config: {
       type: "number",
-      default: 10,
+      default: 75,
       min: 1,
-      max: 50,
+      max: 100,
```
bracesproul:
```suggestion
      max: 250,
```

Issue #208, opened 2025-06-18
Gate `/chat/*` pages behind auth


bracesproul:
Depends on #207 

We should not allow users to access the `/chat` or any of its subpages unless they're authenticated.
Use the same middleware logic you implement in #207 for this

PR #280, opened 2025-06-20
feat: Auth from Stream --> Middleware

starmorph:
### **Summary**

This pull request refactors the application's authentication flow, moving the core logic from the client-side `Stream.tsx` provider to a more robust and secure server-side Next.js middleware. This change addresses the key requirement of handling authentication checks and redirects on the server, aligning with Next.js best practices and improving the overall security and maintainability of the system.

### **Key Changes**

*   **Introduced Authentication Middleware (`apps/web/src/middleware.ts`)**
    *   A new middleware has been created to protect application routes. It intercepts incoming requests to check for an authentication token in the cookies.
    *   **Authenticated users** visiting the root path (`/`) are automatically redirected to the main application interface at `/chat`.
    *  `ready for #208` (setup for this already, small change needed)  *   **Unauthenticated users** attempting to access protected routes under `/chat` are redirected to the root path (`/`) to log in.  #208
    
 *   **Simplified Frontend Logic - Auth Moved from Stream**
    *   The responsibility for route protection has been removed from the `Stream.tsx` provider, simplifying its role to managing client-side state without handling redirects. The `auth-status.tsx` component now fetches user data from the new `/api/auth/user` endpoint.

*   **Centralized and Shared Utilities**
    *   **User Verification (`packages/shared/src/github/verify-user.ts`):** A new utility function was created in the shared package to verify GitHub access tokens against the GitHub API. This centralizes user validation logic.
    *   **Cookie Constants (`packages/shared/src/constants.ts`):** All authentication-related cookie names have been consolidated into the shared `constants.ts` file, eliminating duplication and ensuring consistency across the monorepo.

*   **New User API Endpoint (`apps/web/src/app/api/auth/user/route.ts`)**
    *   A dedicated API route has been added to securely fetch the authenticated user's profile information from the server in the same api request we retrieve the users auth status. This was made for testing the auth status of the user using the new middleware. This endpoint uses the shared `verifyGithubUser` utility and is consumed by the frontend to display user details like username and avatar.

*   **Updated API and Library Files**
    *   The GitHub OAuth callback (`.../api/auth/github/callback/route.ts`) and the auth utility library (`apps/web/src/lib/auth.ts`) have been updated to use the newly centralized cookie constants.

This refactor successfully moves the authentication logic to a more appropriate, secure, and maintainable location within the application architecture, directly fulfilling the requirements of the original request.


apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -28,25 +30,40 @@ async function getGitHubInstallationTokenOrThrow(
...
+      "No GitHub installation ID found. GitHub App must be installed first.",
+    );
   }
 
   const appId = process.env.GITHUB_APP_ID;
-  const privateAppKey = process.env.GITHUB_APP_PRIVATE_KEY;
+  const privateAppKey = process.env.GITHUB_APP_PRIVATE_KEY?.replace(
+    /\\n/g,
+    "\n",
+  );
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts]: why are you doing this?

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -62,10 +79,51 @@ export const { GET, POST, PUT, PATCH, DELETE, OPTIONS, runtime } =
...
+          ),
+          [GITHUB_INSTALLATION_TOKEN_COOKIE]:
+            await getGitHubInstallationTokenOrThrow(req, encryptionKey),
+        };
+
+        return headers;
+      } catch (error) {
+        console.error("Authentication error in API proxy:", error);
+        throw error; // This will cause the request to fail with proper error messaging
+      }
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts]: dont wrap in try catch. it'll throw an error already and we dont need the extra log

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -28,25 +30,40 @@ async function getGitHubInstallationTokenOrThrow(
...
+      appId,
+      privateAppKey,
+    );
+    return encryptGitHubToken(token, encryptionKey);
+  } catch (error) {
+    console.error("Failed to get GitHub installation token:", error);
+    throw new Error(
+      "Failed to get GitHub installation token. The GitHub App may need to be reinstalled.",
+    );
+  }
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts]: same here, dont wrap. if the func throws, we'll see that error. dont need the extra logging

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -62,10 +79,51 @@ export const { GET, POST, PUT, PATCH, DELETE, OPTIONS, runtime } =
...
+            ...body?.configurable,
+            ...configurable,
+          },
+        };
+
+        return modifiedBody;
+      } catch (error) {
+        console.error("Configuration error in API proxy:", error);
+        return body; // Return unchanged body on error
+      }
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts]: not sure why you're doing this. not needed, drop all

apps/web/src/components/github/installation-banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const title = isNewUser
+    ? "🎉 Welcome to Open SWE! Complete your setup"
+    : "Complete your setup to start coding";
+
+  const description = isNewUser
+    ? "You're just one step away from AI-powered development! Install our GitHub App to connect your repositories and start coding with AI assistance."
+    : "Install our GitHub App to grant access to your repositories and enable AI-powered development.";
+
+  return (
+    <div className={isNewUser ? "animate-pulse" : ""}>
```
bracesproul: this wont work in prod... need to use `cn`. please be more careful about this

also im not a huge fan of the pulsing. i would drop altogether

apps/web/src/components/github/installation-prompt.tsx
```diff
@@ -0,0 +1,74 @@
...
+  className = "",
+  variant = "default",
+}: InstallationPromptProps) {
+  const baseClasses =
+    "rounded-md border border-amber-200 bg-amber-50 p-4 dark:border-amber-800 dark:bg-amber-950/20";
+  const bannerClasses =
+    variant === "banner" ? "flex items-center justify-between" : "";
+
+  return (
+    <div className={`${baseClasses} ${bannerClasses} ${className}`}>
```
bracesproul: ...

apps/web/src/components/v2/default-view.tsx
```diff
@@ -84,6 +73,8 @@ export function DefaultView({ threads, threadsLoading }: DefaultViewProps) {
       {/* Main Content */}
       <div className="flex-1 overflow-y-auto">
         <div className="mx-auto max-w-4xl space-y-6 p-4">
+          {/* GitHub Installation Banner - Prominent placement for new users */}
```
bracesproul: unnecessary comment. the name of the component is literally `GitHubInstallationBanner`, so we don't need to say that again
```suggestion
```

apps/web/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (request.nextUrl.pathname === "/") {
+    const token = request.cookies.get(GITHUB_TOKEN_COOKIE)?.value;
+    if (token && token.length > 0) {
+      const url = request.nextUrl.clone();
+      url.pathname = "/chat";
+      return NextResponse.redirect(url);
+    }
+  }
+  return NextResponse.next();
+}
```
bracesproul, line 14: this will need to be fixed, but can wait for the redirect PR.

This is not secure as anyone can just sent the token to whatever string they want, and they'd get through.

Granted it doesn't rlly matter in this case, but bad practice to have it in there.

apps/web/src/providers/Stream.tsx
```diff
@@ -96,281 +81,15 @@ const StreamSession = ({
...
-          </div>
-        </div>
-      </div>
-    );
-  }
-
+export const StreamProvider: React.FC<{
+  children: ReactNode;
+  apiUrl: string;
+  assistantId: string;
```
bracesproul: why did you make this change? Don't do this, we can just read from the env var where it's needed in this file.

apps/web/src/providers/client.ts
```diff
@@ -1,7 +1,12 @@
 import { Client } from "@langchain/langgraph-sdk";
 
 export function createClient(apiUrl: string) {
+  // Use the Next.js API route instead of the direct LangGraph server
+  // This ensures authentication is handled through the server-side proxy
+  const nextApiUrl =
+    typeof window !== "undefined" ? `${window.location.origin}/api` : "/api";
+
```
bracesproul: unnecessary change, we were already doing this. drop.

packages/shared/src/github/verify-user.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  try {
+    const octokit = new Octokit({ auth: accessToken });
+    const { data: user } = await octokit.users.getAuthenticated();
+    if (!user || !user.login) {
+      return undefined;
+    }
+    return user;
+  } catch (_error) {
```
bracesproul:
```suggestion
  } catch {
```

packages/shared/src/github/verify-user.ts
```diff
@@ -0,0 +1,28 @@
+import { Octokit } from "@octokit/rest";
```
bracesproul, line 1: is this dependency added to the shared package.json?

bracesproul (changes requested):

apps/web/src/app/api/[..._path]/route.ts
Comment (starmorph) [reply apps/web/src/app/api/[..._path]/route.ts]: he actual private key needs real newlines (\n) for the cryptographic functions to work
This converts "-----BEGIN PRIVATE KEY-----\\nMIIEvgIBADANBgkqhkiG9w0B..." into proper multi-line format:

Comment (bracesproul) [reply apps/web/src/app/api/[..._path]/route.ts]: this hasn't given me any issues with the env vars that are in 1pass. why did you change?

Comment (bracesproul) [reply apps/web/src/app/api/[..._path]/route.ts]: or are you saying the cryptographic functions weren't working properly before?

apps/open-swe/src/utils/github/api.ts
bracesproul, line 1: why are you casting here? Unnecessary changes. please revert

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -62,10 +72,12 @@ export const { GET, POST, PUT, PATCH, DELETE, OPTIONS, runtime } =
         );
       }
 
-      return {
+      const headers = {
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts]: lets always try for the least amount of changes possible. this is unnecessary, revert

apps/web/src/providers/Stream.tsx
```diff
@@ -45,23 +32,14 @@ async function sleep(ms = 4000) {
...
+const StreamSession = ({ children }: { children: ReactNode }) => {
   const [threadId, setThreadId] = useQueryState("threadId");
   const [customEvents, setCustomEvents] = useState<CustomNodeEvent[]>([]);
   const { refreshThreads } = useThreadsContext();
+
   const streamValue = useTypedStream({
-    apiUrl,
-    assistantId,
+    apiUrl: process.env.NEXT_PUBLIC_API_URL ?? "/api",
+    assistantId: process.env.NEXT_PUBLIC_MANAGER_ASSISTANT_ID ?? "manager",
```
bracesproul: dont do this, just throw if they're not defined

fwiw, the assistant id fallback will work in this case, but the api url will not.

apps/web/src/providers/client.ts
```diff
@@ -1,7 +1,10 @@
 import { Client } from "@langchain/langgraph-sdk";
 
 export function createClient(apiUrl: string) {
+  // Use the Next.js API route instead of the direct LangGraph server
```
bracesproul: least amount of changes possible. there should be no changes to this file

packages/shared/package.json
```diff
@@ -20,6 +20,8 @@
     "@langchain/core": "^0.3.56",
     "@langchain/langgraph": "^0.3.3",
     "@langchain/langgraph-sdk": "^0.0.83",
+    "@octokit/rest": "^22.0.0",
+    "@octokit/types": "^12.0.0",
```
bracesproul: types should be moved to dev dependency. dont need types in prod

packages/shared/src/github/verify-user.ts
bracesproul: never resolved

bracesproul (changes requested):

apps/web/src/app/api/[..._path]/route.ts
Comment (starmorph) [reply apps/web/src/app/api/[..._path]/route.ts]: I was having issues with the github token but it may have been something else, i'll remove and test

packages/shared/src/github/verify-user.ts
starmorph: now added to shared package

apps/web/src/providers/Stream.tsx
```diff
@@ -45,23 +32,22 @@ async function sleep(ms = 4000) {
...
-  children: ReactNode;
-  apiUrl: string;
-  assistantId: string;
-  githubToken: string;
-}) => {
+const StreamSession = ({ children }: { children: ReactNode }) => {
   const [threadId, setThreadId] = useQueryState("threadId");
   const [customEvents, setCustomEvents] = useState<CustomNodeEvent[]>([]);
   const { refreshThreads } = useThreadsContext();
+
```
bracesproul, line 39:
```suggestion
  if (!process.env.NEXT_PUBLIC_API_URL || !process.env.NEXT_PUBLIC_MANAGER_ASSISTANT_ID) {
    throw new Error("Both NEXT_PUBLIC_API_URL and NEXT_PUBLIC_MANAGER_ASSISTANT_ID environment variables must be defined.")
  }
```

apps/web/src/providers/Stream.tsx
```diff
@@ -45,23 +32,22 @@ async function sleep(ms = 4000) {
...
+    apiUrl:
+      process.env.NEXT_PUBLIC_API_URL ??
+      (() => {
+        throw new Error("NEXT_PUBLIC_API_URL is required");
+      })(),
+    assistantId:
+      process.env.NEXT_PUBLIC_MANAGER_ASSISTANT_ID ??
+      (() => {
+        throw new Error("NEXT_PUBLIC_MANAGER_ASSISTANT_ID is required");
+      })(),
```
bracesproul, line 50:
```suggestion
    apiUrl: process.env.NEXT_PUBLIC_API_URL,
    assistantId: process.env.NEXT_PUBLIC_MANAGER_ASSISTANT_ID,
```

will work the same, but much easier to read

apps/web/src/providers/client.ts
```diff
@@ -1,7 +1,9 @@
 import { Client } from "@langchain/langgraph-sdk";
 
 export function createClient(apiUrl: string) {
+  // This ensures authentication is handled through the server-side proxy
```
bracesproul: still has changes

packages/shared/src/github/verify-user.ts
bracesproul: code is still duplicated tho. you need to delete `apps/open-swe/src/security/github-auth.ts` file and import this func in `security/auth.ts` too

apps/web/src/providers/client.ts
```diff
@@ -2,6 +2,6 @@ import { Client } from "@langchain/langgraph-sdk";
 
 export function createClient(apiUrl: string) {
   return new Client({
-    apiUrl,
+    apiUrl: apiUrl,
```
bracesproul:
```suggestion
    apiUrl,
```

bracesproul (approved):

PR #311, opened 2025-06-24
add scroll: on pageload, buttonclick, tabswitch

starmorph:
#### Features
- Auto-scroll to bottom on page load/tab switching
- Scroll-to-bottom buttons for chat and actions areas
- Smart scroll tracking when stream is active
#### Implementation
- Reused `use-stick-to-bottom` library and ported scroll components from old chat page. Wrapped existing scrollable containers with minimal structural changes.
#### Result
- Users get consistent scroll behavior matching the previous chat experience with smooth auto-scrolling and manual control.

### DIFF
--- a/apps/web/src/components/v2/thread-view.tsx
+++ b/apps/web/src/components/v2/thread-view.tsx
@@ -309,53 +309,53 @@ export function ThreadView({
                     </TabsList>
                     <TabsContent value="planner">
                       <Card className="border-border bg-card px-0 py-4 dark:bg-gray-950">
-            <CardContent className="space-y-2 p-3 pt-0">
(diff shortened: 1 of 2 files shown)


apps/web/src/components/v2/thread-view.tsx
```diff
@@ -101,9 +107,23 @@ export function ThreadView({
   const [programmerSession, setProgrammerSession] =
     useState<ManagerGraphState["programmerSession"]>();
 
-  if (!stream.messages?.length) {
-    return null;
-  }
```
starmorph, line 106: I believe this was causing a bug where the page would load blank because of a race condition

PR #317, opened 2025-06-26
Feat: cancel run [manager, planner, programmer]

### LINKED ISSUE #272: Add cancel run button to manager, planner and programmer
We should add `Cancel` buttons to the manager, planner and programmer threads
To cancel a stream, you simply call `stream.stop()`

The manager submit button should transition to a loading/cancel button when its running (like how we did it in the old thread view) and add a button somewhere to cancel programmer/planner? tbd what this looks like. screenshot below for ideas

Image

starmorph:
fixes #272 
Adds a cancel run button to the terminal-input, thread-view manager send button, and top right of planner/programmer tab-- to allow for calling stream.stop on while the manager, planner or programmer is running.

For the manager - I used stream.stop()
I believe stream.stop() did not work for the planner and programmer, perhaps because they are sessions of the original manager stream? so after testing attempting to stop the main stream, planner and programmer, I found using an API request with the session runId worked to stop the stream for planner and programmer.


starmorph:
note to self to remove inline comments tmrw morning

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -303,10 +329,130 @@ export function ThreadView({
...
+                        {selectedTab === "planner" &&
+                          plannerThreadId &&
+                          plannerRunId &&
+                          plannerStream?.isLoading && (
+                            <Button
+                              onClick={async () => {
+                                try {
+                                  if (plannerThreadId && plannerRunId) {
+                                    const apiUrl =
+                                      process.env.NEXT_PUBLIC_API_URL ?? "";
```
starmorph, line 348: might be an improvement if I  add these cancel utilities to a sep file

PR #326, opened 2025-06-27
fix: managerChat autoscroll

starmorph:
1 word fix - set initial=true for scroll wrapper on manager-chat 

Adds the same scroll functionality to manager-chat that already exists for programmer and planner.
Screenshot 2025-06-27 at 1 22 50 PM


bracesproul (approved):
ty!

PR #330, opened 2025-06-27
Feat: cancel-run buttons [manager, planner, programmer]

starmorph:
### **Summary**
Refactored cancel functionality across Manager, Planner & Programmer to use native LangGraph SDK methods with better UX.

### **Key Changes**
- **SDK**: Programmer / Planner use `stream.client.runs.cancel()` 
Screenshot 2025-06-27 at 3 33 11 PM
Screenshot 2025-06-27 at 3 33 44 PM
Screenshot 2025-06-27 at 3 33 50 PM

- **Reusable Hook**: Created `useCancelStream` to eliminate duplicate code
- **Better Errors**: Smart error detection - shows appropriate notifications instead of runtime abort errors
- **Improved UX**: Proper loading states and smooth button transitions

### DIFF
--- a/apps/web/src/components/v2/manager-chat.tsx
+++ b/apps/web/src/components/v2/manager-chat.tsx
@@ -5,9 +5,12 @@ import { useState } from "react";
 import { StickToBottom } from "use-stick-to-bottom";
 import { TooltipIconButton } from "../ui/tooltip-icon-button";
 import { AnimatePresence, motion } from "framer-motion";
-import { Bot, Copy, CopyCheck, Send, User } from "lucide-react";
+import { Bot, Copy, CopyCheck, Send, User, Loader2 } from "lucide-react";
 import { Textarea } from "../ui/textarea";
 import { Button } from "../ui/button";
+import { useStream } from "@langchain/langgraph-sdk/react";
+import { ManagerGraphState } from "@open-swe/shared/open-swe/manager/types";
+import { cn } from "@/lib/utils";
 
 function MessageCopyButton({ content }: { content: string }) {
   const [copied, setCopied] = useState(false);
@@ -61,6 +64,7 @@ interface ManagerChatProps {
   chatInput: string;
   setChatInput: (input: string) => void;
   handleSendMessage: () => void;
+  stream: ReturnType<typeof useStream<ManagerGraphState>>;
 }
 
 export function ManagerChat({
@@ -68,6 +72,7 @@ export function ManagerChat({
   chatInput,
   setChatInput,
   handleSendMessage,
+  stream,
 }: ManagerChatProps) {
   return (
     <div className="border-border bg-muted/30 flex h-full w-1/3 flex-col border-r dark:bg-gray-950">
@@ -135,21 +140,36 @@ export function ManagerChat({
             onKeyDown={(e) => {
               if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) {
                 e.preventDefault();
-                handleSendMessage();
+                if (stream.isLoading) {
+                  stream.stop();
+                } else {
+                  handleSendMessage();
+                }
               }
             }}
           />
           <Button
-            onClick={handleSendMessage}
-            disabled={!chatInput.trim()}
-            size="icon"
-            variant="brand"
+            onClick={stream.isLoading ? () => stream.stop() : handleSendMessage}
+            disabled={stream.isLoading ? false : !chatInput.trim()}
+            size={stream.isLoading ? "sm" : "icon"}
+            variant={stream.isLoading ? "destructive" : "brand"}
+            className={cn(
+              "self-end transition-all duration-200 disabled:opacity-50",
+              stream.isLoading ? "h-12 px-4 py-2" : "h-10 w-10 p-0",
+            )}
           >
-            <Send className="size-4" />
+            {stream.isLoading ? (
+              <>
+                <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+                Cancel
+              </>
+            ) : (
+              <Send className="h-4 w-4" />
+            )}
           </Button>
         </div>
         <div className="text-muted-foreground mt-2 text-xs">
-          Press Cmd+Enter to send
+          Press Cmd+Enter to {stream.isLoading ? "cancel" : "send"}
         </div>
       </div>
     </div>
--- a/apps/web/src/components/v2/thread-view.tsx
+++ b/apps/web/src/components/v2/thread-view.tsx
@@ -22,7 +22,7 @@ import {
   ScrollToBottom,
 } from "../../utils/scroll-utils";
 import { ManagerChat } from "./manager-chat";
-import { toast } from "sonner";
+import { useCancelStream } from "@/hooks/useCancelStream";
 
 const PROGRAMMER_ASSISTANT_ID = process.env.NEXT_PUBLIC_PROGRAMMER_ASSISTANT_ID;
 const PLANNER_ASSISTANT_ID = process.env.NEXT_PUBLIC_PLANNER_ASSISTANT_ID;
@@ -77,33 +77,20 @@ export function ThreadView({
     }
   }, [programmerSession?.runId]);
 
-  // Cancel functions using SDK client methods
-  const cancelPlannerRun = async () => {
-    if (!plannerThreadId || !plannerRunId) return;
+  // Cancel hooks with better error handling
+  const { cancelRun: cancelPlannerRun } = useCancelStream({
+    stream: plannerStream,
+    threadId: plannerThreadId,
+    runId: plannerRunId,
+    streamName: "Planner",
+  });
 
-    try {
-      await plannerStream.client.runs.cancel(plannerThreadId, plannerRunId);
-      toast.success("Planner run cancelled successfully");
-    } catch (error) {
-      console.error("Error cancelling planner run:", error);
-      toast.error("Failed to cancel planner run");
-    }
-  };
-
-  const cancelProgrammerRun = async () => {
-    if (!programmerSession?.threadId || !programmerSession?.runId) return;
-
-    try {
-      await programmerStream.client.runs.cancel(
-        programmerSession.threadId,
-        programmerSession.runId,
-      );
-      toast.success("Programmer run cancelled successfully");
-    } catch (error) {
-      console.error("Error cancelling programmer run:", error);
-      toast.error("Failed to cancel programmer run");
-    }
-  };
+  const { cancelRun: cancelProgrammerRun } = useCancelStream({
+    stream: programmerStream,
+    threadId: programmerSession?.threadId,
+    runId: programmerSession?.runId,
+    streamName: "Programmer",
+  });
 
   const handleSendMessage = () => {
     if (chatInput.trim()) {
--- a/apps/web/src/components/v2/thread-view.tsx
+++ b/apps/web/src/components/v2/thread-view.tsx
@@ -1,10 +1,10 @@
 "use client";
 
 import { v4 as uuidv4 } from "uuid";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { Button } from "@/components/ui/button";
 import { Card, CardContent } from "@/components/ui/card";
-import { ArrowLeft, GitBranch, Terminal, Clock } from "lucide-react";
+import { ArrowLeft, GitBranch, Terminal, Clock, Loader2 } from "lucide-react";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { ThreadSwitcher } from "./thread-switcher";
 import { ThreadDisplayInfo } from "./types";
@@ -22,6 +22,7 @@ import {
   ScrollToBottom,
 } from "../../utils/scroll-utils";
 import { ManagerChat } from "./manager-chat";
+import { toast } from "sonner";
 
 const PROGRAMMER_ASSISTANT_ID = process.env.NEXT_PUBLIC_PROGRAMMER_ASSISTANT_ID;
 const PLANNER_ASSISTANT_ID = process.env.NEXT_PUBLIC_PLANNER_ASSISTANT_ID;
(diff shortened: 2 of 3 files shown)


apps/web/src/hooks/useCancelStream.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    try {
+      await stream.client.runs.cancel(threadId, runId);
+      toast.success(`${streamName} cancelled successfully`, {
+        description: "The running operation has been stopped",
+      });
+    } catch (error) {
+      const isAbortError =
+        error instanceof Error &&
```
bracesproul: Don't use `instanceof` in any JS env other than `node` as it won't work in some (e.g. nextjs api routes & middleware. not 100% sure about nextjs client)

this only applies to prod* it'll work locally, but not necessarily in prod.

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -49,6 +50,32 @@ export function ThreadView({
...
+    if (plannerRunId) {
+      plannerStream.joinStream(plannerRunId).catch(console.error);
+    }
+  }, [plannerRunId]);
+
+  useEffect(() => {
+    if (programmerSession?.runId) {
+      programmerStream.joinStream(programmerSession.runId).catch(console.error);
+    }
+  }, [programmerSession?.runId]);
```
bracesproul: we dont want to duplicate these `useStream` calls, as it'll cause two stream network requests to run at the same time (one here, and once inside the `ActionsRenderer` component.

Can you find a way to not duplicate it? And ideally we can leave the stream's inside the `ActionsRenderer` so they don't render everything on this page when the stream updates, and instead it only rerenders the `ActionsRenderer` component it's inside of

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -61,14 +65,34 @@ interface ManagerChatProps {
...
+    streamName: "Manager",
+  });
+
+  const handleCancel = () => {
+    if (threadId && runId) {
+      cancelRun();
+    } else {
+      stream.stop();
+    }
+  };
```
bracesproul: instead of doing this, just pass a `cancelRun` prop to `ManagerChat`, since the `stream` is alr managed outside of this component, we should manage canceling the run outside too

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -135,21 +159,36 @@ export function ManagerChat({
...
-            size="icon"
-            variant="brand"
+            onClick={stream.isLoading ? handleCancel : handleSendMessage}
+            disabled={stream.isLoading ? false : !chatInput.trim()}
+            size={stream.isLoading ? "sm" : "icon"}
+            variant={stream.isLoading ? "destructive" : "brand"}
+            className={cn(
+              "self-end transition-all duration-200 disabled:opacity-50",
+              stream.isLoading ? "h-12 px-4 py-2" : "h-10 w-10 p-0",
+            )}
```
bracesproul: we didnt have any `className` prop passed before, so we should only pass a custom `className` if we're showing the new loader icon & cancel text
```suggestion
            className={cn(
              stream.isLoading ? "h-12 px-4 py-2" : "",
            )}
```

(not sure if this `className` will work as intended tho, so double check locally)

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -135,21 +159,36 @@ export function ManagerChat({
...
+                Cancel
+              </>
+            ) : (
+              <Send className="h-4 w-4" />
+            )}
           </Button>
         </div>
         <div className="text-muted-foreground mt-2 text-xs">
-          Press Cmd+Enter to send
+          Press Cmd+Enter to {stream.isLoading ? "cancel" : "send"}
```
bracesproul: I dont think we should have a shortcut for canceling. if you want to cancel you should be forced to click cancel so you don't accidentally do it.

bracesproul (changes requested):
some big requests, plz re-request my review once implemented. ty!

starmorph:
one more comment to fix - the duplicate stream working on now

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -61,14 +63,21 @@ interface ManagerChatProps {
...
   messages,
   chatInput,
   setChatInput,
   handleSendMessage,
+  stream,
+  cancelRun,
 }: ManagerChatProps) {
+  const handleCancel = () => {
+    cancelRun();
+  };
```
bracesproul:
```suggestion
```

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -133,19 +142,31 @@ export function ManagerChat({
...
+            className={cn(stream.isLoading ? "h-12 px-4 py-2" : "")}
           >
-            <Send className="size-4" />
+            {stream.isLoading ? (
+              <>
+                <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+                Cancel
+              </>
+            ) : (
+              <Send className="h-4 w-4" />
```
bracesproul:
```suggestion
              <Send className="size-4" />
```

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -49,6 +49,15 @@ export function ThreadView({
...
     useState<ManagerGraphState["programmerSession"]>();
 
+  const [plannerLoading, setPlannerLoading] = useState(false);
+  const [programmerLoading, setProgrammerLoading] = useState(false);
+  const plannerCancelRef = useRef<(() => void) | null>(null);
+  const programmerCancelRef = useRef<(() => void) | null>(null);
+
+  const cancelRun = () => {
+    stream.stop();
+  };
```
bracesproul:
```suggestion
  const cancelRun = () => {
    // TODO: ideally this calls stream.client.runs.cancel(threadId, runId)
    stream.stop();
  };
```

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -105,6 +117,16 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
     }
   }, [runId]);
 
+  // Notify parent of loading state changes
+  useEffect(() => {
+    onLoadingChange?.(stream.isLoading);
+  }, [stream.isLoading, onLoadingChange]);
+
+  // Provide cancel function to parent
```
bracesproul:
```suggestion
  useEffect(() => {
    onLoadingChange?.(stream.isLoading);
  }, [stream.isLoading, onLoadingChange]);

```

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -61,14 +63,21 @@ interface ManagerChatProps {
   chatInput: string;
   setChatInput: (input: string) => void;
   handleSendMessage: () => void;
+  stream: ReturnType<typeof useStream<ManagerGraphState>>;
```
bracesproul:
```suggestion
  isLoading: boolean;
```
let's do this, b/c if we pass the stream, then we should instead refactor it to extract the messages from the stream as well, and call `stream.cancel()` here instead of cancel run. so larger refactor that we don't need to do in this PR.

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -61,14 +63,21 @@ interface ManagerChatProps {
...
+  cancelRun: () => void;
 }
 
 export function ManagerChat({
   messages,
   chatInput,
   setChatInput,
   handleSendMessage,
+  stream,
```
bracesproul:
```suggestion
  isLoading,
```

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -133,19 +142,31 @@ export function ManagerChat({
             placeholder="Type your message..."
             className="border-border bg-background text-foreground placeholder:text-muted-foreground min-h-[60px] flex-1 resize-none text-sm dark:bg-gray-900"
             onKeyDown={(e) => {
-              if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) {
+              if (
+                e.key === "Enter" &&
+                (e.metaKey || e.ctrlKey) &&
+                !stream.isLoading
```
bracesproul:
```suggestion
                !isLoading
```

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -133,19 +142,31 @@ export function ManagerChat({
...
           <Button
-            onClick={handleSendMessage}
-            disabled={!chatInput.trim()}
-            size="icon"
-            variant="brand"
+            onClick={stream.isLoading ? handleCancel : handleSendMessage}
+            disabled={stream.isLoading ? false : !chatInput.trim()}
+            size={stream.isLoading ? "sm" : "icon"}
+            variant={stream.isLoading ? "destructive" : "brand"}
+            className={cn(stream.isLoading ? "h-12 px-4 py-2" : "")}
```
bracesproul:
```suggestion
            onClick={isLoading ? cancelRun : handleSendMessage}
            disabled={isLoading ? false : !chatInput.trim()}
            size={isLoading ? "sm" : "icon"}
            variant={isLoading ? "destructive" : "brand"}
            className={cn(isLoading ? "h-12 px-4 py-2" : "")}
```

apps/web/src/components/v2/manager-chat.tsx
```diff
@@ -133,19 +142,31 @@ export function ManagerChat({
...
           >
-            <Send className="size-4" />
+            {stream.isLoading ? (
```
bracesproul:
```suggestion
            {isLoading ? (
```

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -147,10 +158,39 @@ export function ThreadView({
...
+                        <TabsTrigger value="planner">Planner</TabsTrigger>
+                        <TabsTrigger value="programmer">Programmer</TabsTrigger>
+                      </TabsList>
+
+                      <div className="flex gap-2">
+                        {selectedTab === "planner" &&
+                          plannerLoading &&
+                          plannerCancelRef.current && (
+                            <Button
+                              onClick={() => plannerCancelRef.current?.()}
```
bracesproul: i dont think we need the `?.` if you're already checking that it's defined above
```suggestion
                              onClick={() => plannerCancelRef.current()}
```

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -147,10 +158,39 @@ export function ThreadView({
...
+                            >
+                              Cancel Planner
+                            </Button>
+                          )}
+
+                        {selectedTab === "programmer" &&
+                          programmerLoading &&
+                          programmerCancelRef.current && (
+                            <Button
+                              onClick={() => programmerCancelRef.current?.()}
```
bracesproul: same
```suggestion
                              onClick={() => programmerCancelRef.current()}
```

apps/web/src/hooks/useCancelStream.tsx
```diff
@@ -0,0 +1,47 @@
+import { UseStream } from "@langchain/langgraph-sdk/react";
+import { toast } from "sonner";
+
+interface UseCancelStreamProps {
+  stream: UseStream<any>;
+  threadId?: string;
+  runId?: string;
+  streamName: string; // "Planner" | "Programmer"
```
bracesproul: just do this? always prefer strict typing when possible
```suggestion
  streamName: "Planner" | "Programmer";
```

apps/web/src/hooks/useCancelStream.tsx
```diff
@@ -0,0 +1,47 @@
+import { UseStream } from "@langchain/langgraph-sdk/react";
+import { toast } from "sonner";
+
+interface UseCancelStreamProps {
+  stream: UseStream<any>;
```
bracesproul: our state objects will always be objects, so prefer doing this for stricter typing:
```suggestion
  stream: UseStream<Record<string, unknown>>;
```

apps/web/src/hooks/useCancelStream.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    try {
+      await stream.client.runs.cancel(threadId, runId);
+      toast.success(`${streamName} cancelled successfully`, {
+        description: "The running operation has been stopped",
+      });
+    } catch (error) {
+      const errorStr = String(error);
+      const isAbortError =
+        errorStr.includes("AbortError") || errorStr.includes("abort");
```
bracesproul:
```suggestion
      const isAbortError = errorStr.toLowerCase().includes("abort");
```

apps/web/src/components/v2/cancel-stream-button.tsx
bracesproul, line 1: looked locally, and you're not actually using this component anywhere? plz fix so you're using it

apps/web/src/components/v2/thread-view.tsx
bracesproul: Nvm: pulled locally and what you had (`?.`) is actually cleaner than what's needed to make typescript happy. so i reverted to what you had

apps/web/src/components/v2/thread-view.tsx
bracesproul: Nvm: pulled locally and what you had (`?.`) is actually cleaner than what's needed to make typescript happy. so i reverted to what you had

starmorph:
Flashing loading issue details: 
When switching to the planner tab, the planner's ActionsRenderer gets rendered and its useStream hook:
Reports isLoading: true during stream initialization/connection
Triggers onLoadingChange(true) → sets plannerLoading: true
Button shows (flash appears)
Stream realizes there's no actual work → isLoading: false
Triggers onLoadingChange(false) → sets plannerLoading: false
Button disappears (flash ends)
The planner isn't actually running anything - it's just the stream connection initialization being misinterpreted as "loading".

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -24,6 +25,8 @@ interface ActionsRendererProps {
   ) => void;
   programmerSession?: ManagerGraphState["programmerSession"];
   setSelectedTab?: Dispatch<SetStateAction<"planner" | "programmer">>;
+  onLoadingChange?: (isLoading: boolean) => void;
+  onStreamReady?: (cancelFn: () => void) => void;
```
bracesproul: i dont think this needs to be optional?

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -146,10 +159,39 @@ export function ThreadView({
...
-                    </TabsList>
+                    <div className="flex items-center justify-between">
+                      <TabsList className="bg-muted/70 dark:bg-gray-800">
+                        <TabsTrigger value="planner">Planner</TabsTrigger>
+                        <TabsTrigger value="programmer">Programmer</TabsTrigger>
+                      </TabsList>
+
+                      <div className="flex gap-2">
+                        {selectedTab === "planner" &&
+                          plannerLoading &&
```
bracesproul:
```suggestion
```

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -146,10 +159,39 @@ export function ThreadView({
...
+                              stream={stream}
+                              threadId={plannerThreadId}
+                              runId={plannerRunId}
+                              streamName="Planner"
+                              isVisible={plannerLoading}
+                            />
+                          )}
+
+                        {selectedTab === "programmer" &&
+                          programmerLoading &&
```
bracesproul:
```suggestion
```

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -105,6 +117,14 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
     }
   }, [runId]);
 
+  useEffect(() => {
+    onLoadingChange?.(stream.isLoading);
+  }, [stream.isLoading, onLoadingChange]);
```
bracesproul: drop

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -105,6 +117,14 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
...
+
+  useEffect(() => {
+    onStreamReady?.(cancelRun);
+  }, [onStreamReady, runId]); // Depend on runId instead of cancelRun to avoid infinite loops
```
bracesproul: update to only pass the `cancelRun` func to this prop if graph is actually streaming

bracesproul (approved):
lets goooo!

PR #337, opened 2025-07-02
Feat: Polling /chat/threads 

### LINKED ISSUE #269: Add polling to the all threads page (`/chat/threads`)

starmorph:
Fixes #269 
fixes #269 
Built on top of PR #336 utilizing the same SWR hook for polling. 
15 second polling default, customizable per section we are polling.


package.json
bracesproul, line 1: never add deps like this to the root. always add them to the specific package they're used in

I think yarn should yell at you if you try and do this?

bracesproul (approved):
lgtm!

PR #344, opened 2025-07-02
Fix: scroll to bottom UX improvement

### LINKED ISSUE #327: Scroll to bottom button blocks actions from being taken
It looks like it's width is actually wider than the button itself. Should be fixed
https://www.loom.com/share/424026341fb14031acb1299a957113d9

starmorph:
Fixes #327 

- reduce width and improve position of scroll to bottom button to not interfere with accepting / rejecting plans.


bracesproul (approved):
lgtm. not going to pull locally but i trust you tested locally.

PR #349, opened 2025-07-03
[open-swe] feat: integrate MCPDoc Server into planner and programmer

aliyanishfaq:
Adds MCPDoc server integration to both planner and programmer graphs. This allows the planning and the programming agent to access langGraph python and javascript documentation.

### New Files
- `apps/open-swe/src/utils/mcp-client.ts` - MCP client singleton with error handling
- `apps/open-swe/src/constants.js` - MCP server configurations (MCP_SERVERS)

### Modified Files
- `apps/open-swe/src/graphs/planner/nodes/generate-message/index.ts`
  - Added MCP tools loading alongside standard tools
  - Error handling for MCP tool failures

- `apps/open-swe/src/graphs/programmer/nodes/generate-message/index.ts`
  - Added MCP tools loading alongside standard tools
  - Error handling for MCP tool failures

- `apps/open-swe/src/graphs/planner/nodes/take-action.ts`
  - Updated toolsMap to include dynamically loaded MCP tools
  - Added support for different tool return formats (string vs object)
  - Schema validation with try-catch for mixed tool types

- `apps/open-swe/src/graphs/programmer/nodes/take-action.ts`
  - Updated toolsMap to include dynamically loaded MCP tools
  - Added support for different tool return formats (string vs object)
  - Schema validation for mixed tool types


apps/open-swe/src/graphs/planner/nodes/generate-message/index.ts
```diff
@@ -47,11 +49,25 @@ export async function generateAction(
   config: GraphConfig,
 ): Promise<PlannerGraphUpdate> {
   const model = await loadModel(config, Task.ACTION_GENERATOR);
+  let mcpTools: StructuredToolInterface[] = [];
+
+  try {
```
bracesproul: Let's wrap this create MCP server & get tools w/ the try catch into a shared util since we're redefining the same code in multiple places

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -34,11 +37,19 @@ export async function takeActions(
...
+
+  let mcpTools: StructuredToolInterface[] = [];
+  try {
+    const client = mcpClient();
+    mcpTools = await client.getTools();
+  } catch (error) {
+    logger.error(`Error getting MCP tools: ${error}`);
+  }
+
+  const allTools = [shellTool, rgTool, plannerNotesTool, ...mcpTools];
```
bracesproul: Oh jeez good fix. Can't believe I was defining them each before and didn't catch this 🙃

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -82,9 +98,21 @@ export async function takeActions(
       ) {
         logger.error("Received tool input did not match expected schema", {
           toolCall,
-          expectedSchema: zodSchemaToString(tool.schema),
+          expectedSchema: (() => {
+            try {
+              return zodSchemaToString(tool.schema as ZodTypeAny);
```
bracesproul: Instead, can we try to check if the schema is a zod schema (I believe the zod package exports a util for this, otherwise lmk and I can help find the one we use in the LC repo) and parse, otherwise stringify? Also ensure we're truncating (use the truncate content util) incase the stringify call returns a massive string

bracesproul: And for all other instances in the pr

package.json
```diff
@@ -23,5 +23,8 @@
     "@langchain/langgraph-sdk": "^0.0.85",
     "@langchain/core": "^0.3.58"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "@langchain/mcp-adapters": "^0.5.2"
```
bracesproul: Deps need to be added to the specific package/app in the mono repo they're used in, not the root package.json

bracesproul (changes requested):
Really good, mostly nits/small reqs. Haven't tested locally, will do after these comments are addressed. Ty!

bracesproul:
One more thing, we should allow users to configure the MCP servers (configurable obj defined in the shared package) instead of just hard coding the LC MCP server.

packages/shared/src/open-swe/types.ts
```diff
@@ -413,6 +432,14 @@ export const GraphConfigurationMetadata: {
       type: "hidden",
     },
   },
+  mcpServers: {
+    x_open_swe_ui_config: {
+      type: "textarea",
```
bracesproul: nit
```suggestion
      type: "json",
```

packages/shared/src/open-swe/types.ts
```diff
@@ -136,6 +137,24 @@ export type CustomRules = {
...
+   */
+  args: string[];
+};
+
+export type McpServers = {
+  /**
+   * Map of server names to their configurations
+   */
+  [serverName: string]: McpServerConfig;
+};
```
bracesproul: is this config compatible with http streamable servers? If not, we want to ensure it is

apps/open-swe/src/utils/mcp-client.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Returns a shared MCP client instance
+ */
+export function mcpClient(mcpServers?: McpServers): MultiServerMCPClient {
+  const serversToUse = mcpServers || {};
+  const configHash = JSON.stringify(serversToUse);
+
+  // Recreate client if configuration changed
+  if (!mcpClientInstance || lastConfigHash !== configHash) {
+    mcpClientInstance = new MultiServerMCPClient({
+      additionalToolNamePrefix: "",
```
bracesproul, line 27: is this a required arg? If it isn't, we should drop. If it is, we should flag to the JS team since it shouldn't be

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -34,11 +37,13 @@ export async function takeActions(
   const shellTool = createShellTool(state);
   const rgTool = createRgTool(state);
   const plannerNotesTool = createPlannerNotesTool();
-  const toolsMap = {
-    [shellTool.name]: shellTool,
-    [rgTool.name]: rgTool,
-    [plannerNotesTool.name]: plannerNotesTool,
-  };
+
+  const mcpTools = await getMcpTools(_config);
```
bracesproul: `config` was only prefixed with an underscore before b/c it wasn't used. since it's used now we can drop the prefix underscore
```suggestion
  const mcpTools = await getMcpTools(config);
```

apps/open-swe/src/utils/mcp-client.ts
```diff
@@ -0,0 +1,63 @@
...
+const logger = createLogger(LogLevel.INFO, "MCP Client");
+
+// Singleton instance of the MCP client
+let mcpClientInstance: MultiServerMCPClient | null = null;
+let lastConfigHash: string | null = null;
+
+/**
+ * Returns a shared MCP client instance
+ */
+export function mcpClient(mcpServers?: McpServers): MultiServerMCPClient {
```
bracesproul: why is `mcpServers` optional here? Should be required?

apps/open-swe/src/utils/mcp-client.ts
```diff
@@ -0,0 +1,63 @@
...
+  return mcpClientInstance;
+}
+
+/**
+ * Gets MCP tools with configurable servers
+ * @param config GraphConfig containing optional MCP servers configuration
+ * @returns Array of MCP tools, empty array if error occurs
+ */
+export async function getMcpTools(
+  config?: GraphConfig,
```
bracesproul: same w/ config. should be required

apps/open-swe/src/utils/mcp-client.ts
aliyanishfaq: Not a required arg. I deliberately added because I think without it tools start with 'mcp-'. Seemed cleaner without a prefix, can, however, remove.

bracesproul: oh the adapters pkg adds that prefix for us...? im not a fan. going to flag in the channel

bracesproul: but leave the empty string prefix in* lets keep the default tool names from the servers

apps/open-swe/src/utils/mcp-client.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Gets MCP tools with configurable servers
+ * @param config GraphConfig containing optional MCP servers configuration
+ * @returns Array of MCP tools, empty array if error occurs
+ */
+export async function getMcpTools(
+  config: GraphConfig,
+): Promise<StructuredToolInterface[]> {
+  try {
+    const mergedServers: McpServers = { ...DEFAULT_MCP_SERVERS };
```
bracesproul, line 48:
```suggestion
    // TODO: Remove default MCP servers obj once UI is implemented
    const mergedServers: McpServers = { ...DEFAULT_MCP_SERVERS };
```

apps/open-swe/src/utils/mcp-client.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const mcpServersConfig = config?.configurable?.["mcpServers"];
+    if (mcpServersConfig) {
+      try {
+        const userServers: McpServers = JSON.parse(mcpServersConfig);
+        for (const serverName in userServers) {
+          const serverConfig = userServers[serverName];
+          if (!serverConfig) continue;
+          try {
+            McpServerConfigSchema.parse(serverConfig);
```
bracesproul, line 58: great, this is exactly what i was talking about!

packages/shared/src/open-swe/types.ts
```diff
@@ -136,6 +138,216 @@ export type CustomRules = {
   testingInstructions?: string;
 };
 
+export const oAuthClientProviderSchema = z.custom<OAuthClientProvider>(
```
bracesproul: since this is so much code, it would be cleaner if it was in its own file. I'll push up a fix

bracesproul (approved):
lgtm, going to push up the refactor then merge! thank you!

PR #360, opened 2025-07-07
Feat: Loading state action list

### LINKED ISSUE #263: Add loading state to planner/programmer action lists
When you switch tabs, we show an empty tab instead of a loader UI (skeleton)
We should add some sort of loading state.

If we have the run & thread session ID in the other graph (manager or planner), then we can render a loading state until the `joinStream` promise resolves.

starmorph:
Implements a loading state when loading to fetch a programmer or planner session 
Fixes #263 
Screenshot 2025-07-07 at 1 05 30 PM

### DIFF
--- a/apps/web/src/components/v2/actions-renderer.tsx
+++ b/apps/web/src/components/v2/actions-renderer.tsx
@@ -103,8 +103,9 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
   const [customNodeEvents, setCustomNodeEvents] = useState<CustomNodeEvent[]>(
     [],
   );
-  const [isJoiningStream, setIsJoiningStream] = useState(false);
-  const joinedRunId = useRef<string | null>(null);
+
+  // Track which threadId+runId combinations we've joined to avoid re-joining
+  const joinedKey = useRef<string | null>(null);
   const stream = useStream<State>({
     apiUrl: process.env.NEXT_PUBLIC_API_URL,
     assistantId: graphId,
@@ -160,26 +161,19 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
     });
   }, [stream.messages]);
 
+  // Join stream when we have a new threadId+runId combination
   useEffect(() => {
-    // Reset state when runId becomes undefined/null
-    if (!runId) {
-      joinedRunId.current = null;
-      setIsJoiningStream(false);
+    if (!runId || !threadId) {
+      joinedKey.current = null;
       return;
     }
 
-    // Only join if we haven't joined this specific runId yet
-    if (joinedRunId.current !== runId) {
-      joinedRunId.current = runId;
-      setIsJoiningStream(true);
-      // TODO: If the SDK changes go in, use this instead:
-      // stream.joinStream(runId, undefined, { streamMode: ["values", "messages", "custom"]}).catch(console.error);
-      stream
-        .joinStream(runId)
-        .catch(console.error)
-        .finally(() => setIsJoiningStream(false));
+    const currentKey = `${threadId}:${runId}`;
+    if (joinedKey.current !== currentKey) {
+      joinedKey.current = currentKey;
+      stream.joinStream(runId).catch(console.error);
     }
-  }, [runId, stream]);
+  }, [runId, threadId, stream]);
 
   // Filter out human & do not render messages
   const filteredMessages = stream.messages?.filter(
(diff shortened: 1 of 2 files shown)


apps/web/src/components/v2/thread-view-loading.tsx
```diff
@@ -8,9 +8,62 @@ interface ThreadViewLoadingProps {
   onBackToHome?: () => void;
 }
 
-const isEven = (num: number) => num % 2 === 0;
+function isEven(n: number) {
+  return n % 2 === 0;
+}
+
+export function LoadingActionsCardContent() {
```
bracesproul, line 15: shouldn't need a whole new component. instead, refactor out L81:L113 (the contents of the card)

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -167,15 +169,16 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
...
 
-  const streamJoined = useRef(false);
+  // TODO: If the SDK changes go in, use this instead:
+  // stream.joinStream(runId, undefined, { streamMode: ["values", "messages", "custom"]}).catch(console.error);
   useEffect(() => {
-    if (!streamJoined.current && runId) {
-      streamJoined.current = true;
-      // TODO: If the SDK changes go in, use this instead:
-      // stream.joinStream(runId, undefined, { streamMode: ["values", "messages", "custom"]}).catch(console.error);
+    if (runId && runId !== joinedRunId.current) {
```
bracesproul, line 187: smart fix

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -214,6 +217,14 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
     }
   }, [stream.values, graphId]);
 
+  const hasContent =
+    (filteredMessages && filteredMessages.length > 0) ||
+    customNodeEvents.length > 0;
+
+  if (runId && !hasContent) {
```
bracesproul: don't do this since technically a run can be started even if there aren't any messages. instead do what we discussed yesterday where you add a new state var for loading. set it to true before calling `joinStream` and add a `.finally(() => setStreamLoading(false))` to set it to false afterwards.

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -167,15 +169,16 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
...
+  // TODO: If the SDK changes go in, use this instead:
+  // stream.joinStream(runId, undefined, { streamMode: ["values", "messages", "custom"]}).catch(console.error);
   useEffect(() => {
-    if (!streamJoined.current && runId) {
-      streamJoined.current = true;
-      // TODO: If the SDK changes go in, use this instead:
-      // stream.joinStream(runId, undefined, { streamMode: ["values", "messages", "custom"]}).catch(console.error);
+    if (runId && runId !== joinedRunId.current) {
+      joinedRunId.current = runId;
       stream.joinStream(runId).catch(console.error);
```
bracesproul: here
```suggestion
      setStreamLoading(true)
      stream.joinStream(runId).catch(console.error).finally(() => setStreamLoading(false));
```

bracesproul (dismissed):

starmorph:
logic details: streamLoading doesn't resolve until the agent is done running so when I implemented the review comments above @bracesproul the loading state would show until the planner was done running. 

I did a hybrid of the CR + original implementation with the setLoading state variable and finally, but also using the content to stop the loader at the appropriate time before the streamLoading resolves (when agent finishes). I believe this gets the desired loading result

apps/web/src/components/v2/thread-view-loading.tsx
```diff
@@ -58,7 +58,26 @@ export function LoadingActionsSection() {
...
+  return (
+    <Card className="border-border bg-card dark:bg-gray-950">
+      <CardHeader className="p-3">
+        <div className="flex items-center gap-2">
+          <div className="bg-muted h-4 w-32 animate-pulse rounded dark:bg-gray-700"></div>
+          <div className="ml-auto flex items-center gap-1">
+            <Loader2 className="text-muted-foreground h-3 w-3 animate-spin" />
+            <span className="text-muted-foreground text-xs">
+              Loading actions...
+            </span>
```
bracesproul: Loading actions is duplicated now. Will push up a fix, but remember to test all ways the component can be rendered if you're updating it. In this case it was the initial loading state when you load/refresh the page

PR #370, opened 2025-07-09
Feat: Thread Status SWR Polling 

### LINKED ISSUE #275: Fix thread status

starmorph:
# Thread Status SWR Polling System
Screenshot 2025-07-10 at 3 13 36 PM

Fixes #275 
[Audit Review of This PR from open-swe](https://github.com/langchain-ai/chat-langchain/pull/451/files)
## 🎯 **Overview**
This PR completely refactors the thread status polling system, replacing manual useEffect-based polling with a modern SWR-first architecture. The implementation eliminates polling conflicts, reduces API load, and provides superior caching with automatic request deduplication and intelligent revalidation strategies.

### **Performance Optimization**
- **Request Deduplication:** SWR prevents duplicate API calls when multiple components request the same thread status, reducing server load
- **Intelligent Caching:** Stale-while-revalidate pattern provides instant UI updates from cache while fetching fresh data in background
- **Reduced Memory Footprint:** Eliminated manual interval management and cleanup logic, preventing memory leaks from abandoned timers
- **Background Revalidation:** Automatic data freshness without blocking UI interactions or causing loading states

## 🎨 **User Experience Impact**
- **Perceived Performance:** faster status updates due to SWR cache hits providing instant feedback
- **Reduced Loading States:** Stale-while-revalidate eliminates unnecessary loading spinners during background updates
- **Improved Responsiveness:** Real-time status updates without polling delays or race conditions

### **State Management Optimization**
- **Separation of Concerns:** Zustand store for UI state only (active thread, polling preferences) while SWR handles all data caching
- **Reactive Updates:** Components automatically re-render when SWR cache updates, eliminating need for manual state propagation
- **Normalized Data Flow:** Single source of truth through SWR cache instead of multiple competing state sources

### **Developer Experience**
- **Declarative API:** Replaced imperative polling management with simple `useThreadStatus(threadId)` hook calls
- **Standardized Patterns:** Consistent SWR configuration across all thread-related data fetching
- **Simplified Mental Model:** Clear data flow from API → SWR Cache → UI Components without complex state synchronization
- **Enhanced Debugging:** SWR DevTools support and predictable cache behavior instead of scattered polling logic

### **Reliability & Error Handling**
- **Eliminated Race Conditions:** Single polling source prevents timing conflicts between multiple useEffect intervals
- **Exponential Backoff:** Built-in retry logic with configurable intervals (3 retries, 5s intervals) instead of manual error handling
- **Network Resilience:** Automatic revalidation on network reconnection and focus events
- **Graceful Degradation:** Proper error boundaries with fallback states instead of application crashes

#### **New Files Created (1)**
- **`lib/swr-config.ts`** - Centralized SWR configuration constants with standardized intervals, retry logic, and error handling

#### **Files Deleted (4)**
- **`hooks/useThreadPolling.ts`** - Deprecated custom polling hook with manual interval management
- **`lib/polling/thread-poller.ts`** - Class-based polling implementation with memory leak potential
- **`hooks/useThreads.tsx`** - Redundant thread fetching logic superseded by useThreadsSWR
- **`components/error-boundary.tsx`** - Custom error boundary replaced by native SWR error handling

#### **Files Significantly Refactored (4)**
- **`hooks/useThreadsSWR.ts`** - Simplified from 94→55 lines, removed manual Zustand synchronization
- **`hooks/useThreadStatus.ts`** - Rebuilt as direct SWR consumer instead of Zustand cache reader
- **`stores/thread-store.ts`** - Reduced from 190+→30 lines, eliminated data caching in favor of UI state only
- **`lib/schemas/thread-status.ts`** - Removed 65 lines of unused schema definitions and dead code

#### **Files Updated (5)**
- **`app/(v2)/chat/[thread_id]/page.tsx`** - Migrated from deprecated useThreads to useThreadsSWR
- **`components/v2/thread-card.tsx`** - Implemented proper SWR error handling patterns
- **`components/v2/thread-view.tsx`** - Added SWR error state management with graceful fallbacks
- **`providers/Thread.tsx`** - Removed manual polling logic and simplified provider responsibilities

This refactor establishes a production-ready foundation for thread status polling that leverages industry-standard patterns for optimal performance, reliability, and maintainability.

### DIFF
--- a/apps/web/src/services/thread-status.service.ts
+++ b/apps/web/src/services/thread-status.service.ts
@@ -343,7 +343,7 @@ async function performFullStatusCheck(
     // No run fetch needed for planners - thread status is sufficient
 
     setCachedSessionData(sessionCache, plannerCacheKey, {
-      plannerData: { thread: plannerThread, run: null },
+      plannerData: { thread: plannerThread },
     });
   }
 
@@ -399,7 +399,7 @@ async function performFullStatusCheck(
     // No run fetch needed - we only check task completion from thread data
 
     setCachedSessionData(sessionCache, programmerCacheKey, {
-      programmerData: { thread: programmerThread, run: null },
+      programmerData: { thread: programmerThread },
     });
   }
 


starmorph:
Audit from Open-SWE of this PR https://github.com/langchain-ai/chat-langchain/pull/451/files

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -110,7 +110,10 @@ async function getInstallationNameFromReq(
     if (isNewRunRequest(req.url, req.method)) {
       return await getInstallationName(installationId);
     }
-    return "";
+
+    // For all other requests (like GET /threads/{id}), also get installation name
+    // This ensures the GitHub installation name header is always present
+    return await getInstallationName(installationId);
```
Comment (starmorph) [inline apps/web/src/app/api/[..._path]/route.ts]: this was added to avoid an error, but will be replaced + improved my GH org PR

apps/web/src/app/(v2)/chat/[thread_id]/page.tsx
```diff
@@ -28,13 +29,29 @@ export default function ThreadPage({
...
 
-  const { threads, threadsLoading } = useThreads<GraphState>(MANAGER_GRAPH_ID);
+  const { threads, isLoading: threadsLoading } = useThreadsSWR<GraphState>({
+    assistantId: MANAGER_GRAPH_ID,
+  });
+
   // Find the thread by ID
   const thread = threads.find((t) => t.thread_id === thread_id);
 
+  // Call hooks before any early returns
```
Comment (bracesproul) [inline apps/web/src/app/(v2)/chat/[thread_id]/page.tsx]:
```suggestion
```

apps/web/src/app/(v2)/chat/threads/page.tsx
```diff
@@ -22,38 +22,43 @@ function AllThreadsPageContent() {
...
-    refreshInterval: 15000, // Poll every 15 seconds
   });
   const [searchQuery, setSearchQuery] = useState("");
   const [statusFilter, setStatusFilter] = useState<FilterStatus>("all");
 
-  // Convert Thread objects to ThreadDisplayInfo for UI
-  const displayThreads: ThreadDisplayInfo[] = threads.map(threadToDisplayInfo);
+  if (!threads) {
+    return <div>No threads</div>;
+  }
```
bracesproul: we already have a no threads state that's rendered in the main JSX returned below. Ideally we can reuse that state and not return early. If for whatever reason that's impossible with these changes, could we refactor it out into a shared `NoThreads` component and return that instead?

Def dont want to just render the text `No threads` if there aren't any threads found.

apps/web/src/app/(v2)/chat/threads/page.tsx
```diff
@@ -22,38 +22,43 @@ function AllThreadsPageContent() {
...
-    running: displayThreads.filter((t) => t.status === "running").length,
-    completed: displayThreads.filter((t) => t.status === "completed").length,
-    failed: displayThreads.filter((t) => t.status === "failed").length,
-    pending: displayThreads.filter((t) => t.status === "pending").length,
+    all: filteredThreads.length,
+    running: filteredThreads.length,
+    completed: filteredThreads.length,
+    failed: filteredThreads.length,
+    pending: filteredThreads.length,
   };
```
bracesproul: this def wont work? we need to figure out how to actually render the proper counts

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -119,7 +119,6 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
         setCustomNodeEvents((prev) => [...prev, event]);
       }
     },
-    fetchStateHistory: false,
```
bracesproul: why'd you drop this? plz add back

apps/web/src/components/v2/terminal-input.tsx
```diff
@@ -57,7 +57,6 @@ export function TerminalInput({
     apiUrl,
     assistantId,
     reconnectOnMount: true,
-    fetchStateHistory: false,
```
bracesproul: same here. please be careful when opening PRs to not do stuff like this

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -4,24 +4,45 @@ import {
...
 import { useRouter } from "next/navigation";
 import { Badge } from "../ui/badge";
 import { Button } from "../ui/button";
 import { Skeleton } from "../ui/skeleton";
-
-export function ThreadCard({ thread }: { thread: ThreadDisplayInfo }) {
+import { useThreadStatus } from "@/hooks/useThreadStatus";
+import { ThreadMetadata } from "./types";
+import { ThreadDisplayStatus } from "@/lib/schemas/thread-status";
+function ThreadCardContent({ thread }: { thread: ThreadMetadata }) {
```
bracesproul: remember to look out for code styling issues like this.
```suggestion
import { ThreadDisplayStatus } from "@/lib/schemas/thread-status";

function ThreadCardContent({ thread }: { thread: ThreadMetadata }) {
```

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -4,24 +4,45 @@ import {
...
   const router = useRouter();
 
-  const getStatusColor = (status: ThreadDisplayInfo["status"]) => {
+  // Re-enable real-time status now that cookies are confirmed present
```
bracesproul: what does this mean/why is it referencing cookies? i think we can drop?

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -4,24 +4,45 @@ import {
...
+  const getStatusColor = (status: ThreadDisplayStatus) => {
     switch (status) {
       case "running":
         return "dark:bg-blue-950 bg-blue-100 dark:text-blue-400 text-blue-700";
       case "completed":
         return "dark:bg-green-950 bg-green-100 dark:text-green-400 text-green-700";
+      case "error":
+        return "dark:bg-red-950 bg-red-100 dark:text-red-400 text-red-700";
+      case "idle":
+        return "dark:bg-gray-800 bg-gray-200 dark:text-gray-400 text-gray-700";
```
bracesproul: this is the same styles as the `default` case. just drop so we don't reimplement the same styling twice
```suggestion
```

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -31,12 +52,18 @@ export function ThreadCard({ thread }: { thread: ThreadDisplayInfo }) {
...
+  const getStatusIcon = (status: ThreadDisplayStatus) => {
     switch (status) {
       case "running":
         return <Loader2 className="h-4 w-4 animate-spin" />;
       case "completed":
         return <CheckCircle className="h-4 w-4" />;
+      case "error":
+        return <AlertCircle className="h-4 w-4" />;
+      case "idle":
+        return <AlertCircle className="h-4 w-4" />;
```
bracesproul: why is idle an alert? i think we can do better than this!

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -82,11 +109,11 @@ export function ThreadCard({ thread }: { thread: ThreadDisplayInfo }) {
           </div>
           <Badge
             variant="secondary"
-            className={`${getStatusColor(thread.status)} text-xs`}
+            className={`${getStatusColor(displayStatus)} text-xs`}
```
bracesproul: oversight from last review, but this needs to be wrapped in cn
```suggestion
            className={cn("text-xs", getStatusColor(displayStatus))}
```

apps/web/src/components/v2/thread-card.tsx
```diff
@@ -138,6 +165,10 @@ export function ThreadCard({ thread }: { thread: ThreadDisplayInfo }) {
   );
 }
 
+export function ThreadCard({ thread }: { thread: ThreadMetadata }) {
+  return <ThreadCardContent thread={thread} />;
+}
```
bracesproul: This seems unnecessary. lets just delete this/rename `ThreadCardContent` to `ThreadCard`

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -50,6 +52,13 @@ export function ThreadView({
   const [programmerSession, setProgrammerSession] =
     useState<ManagerGraphState["programmerSession"]>();
 
+  // Re-enable real-time status now that cookies are confirmed present
```
bracesproul: same as above, not sure what this means/dont think we need it
```suggestion
```

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -100,11 +109,17 @@ export function ThreadView({
           <div className="flex min-w-0 flex-1 items-center gap-2">
             <div
               className={`size-2 flex-shrink-0 rounded-full ${
```
bracesproul: another className missing `cn`. It's very easy to miss these in a review so plz make sure you're looking out for them in dev and fixing anytime you run into them

apps/web/src/components/v2/thread-view.tsx
```diff
@@ -260,3 +275,19 @@ export function ThreadView({
     </div>
   );
 }
+
+export function ThreadView({
```
bracesproul: same as `ThreadContent` component. rename `ThreadViewContent` to `ThreadView` and delete this wrapper

apps/web/src/components/v2/types.ts
```diff
@@ -2,11 +2,12 @@ import { getThreadTitle } from "@/lib/thread";
 import { Thread } from "@langchain/langgraph-sdk";
 import { getActivePlanItems } from "@open-swe/shared/open-swe/tasks";
 import { ManagerGraphState } from "@open-swe/shared/open-swe/manager/types";
+import { ThreadDisplayStatus } from "@/lib/schemas/thread-status";
 
-export interface ThreadDisplayInfo {
+// Thread metadata without status (status comes from useThreadStatus hook)
```
bracesproul: reminder: only add comments when someone who's a core contributor to the repo would struggle to understand some code without the comment
```suggestion
```

apps/web/src/components/v2/types.ts
```diff
@@ -22,35 +23,19 @@ export interface ThreadDisplayInfo {
   };
 }
 
-// Utility functions to convert between Thread and ThreadDisplayInfo
-export function threadToDisplayInfo(
+// Complete thread display info (metadata + status)
```
bracesproul:
```suggestion
```

apps/web/src/components/v2/types.ts
```diff
@@ -74,18 +59,41 @@ export function threadToDisplayInfo(
...
+    fallbackStatus = "error";
+  } else if (thread.status === "interrupted") {
+    fallbackStatus = "paused";
+  }
+
+  return {
+    ...metadata,
+    status: fallbackStatus,
+  };
+}
```
bracesproul: there usually won't be any case where our frontend would need a backwards compat function + this ins't used anywhere. drop
```suggestion
```

apps/web/src/hooks/useThreadDisplayInfo.ts
```diff
@@ -0,0 +1,37 @@
+import { Thread } from "@langchain/langgraph-sdk";
+import { ManagerGraphState } from "@open-swe/shared/open-swe/manager/types";
+import { ThreadDisplayInfo, threadToMetadata } from "@/components/v2/types";
+import { useThreadStatus } from "./useThreadStatus";
+import { useMemo } from "react";
+
+/**
+ * Hook that combines thread metadata with real-time status
+ * Replaces the synchronous threadToDisplayInfo function
```
bracesproul:
```suggestion
```

apps/web/src/hooks/useThreadsSWR.ts
bracesproul, line 1: what's the difference between this and `useThreadStatus`? looks to me like `useThreadStatus` is always called in components which are children of components this hook is called in, so can't we just get the status from this func and pass it as props instead of having two different hooks?

apps/web/src/lib/schemas/thread-status.ts
```diff
@@ -0,0 +1,33 @@
...
+  "running",
+  "completed",
+  "failed",
+  "pending",
+  "idle",
+  "paused",
+  "error",
+]);
+
+export type ThreadDisplayStatus = z.infer<typeof ThreadDisplayStatus>;
```
bracesproul: a few things:
1. try to avoid naming variables and types the same thing. TS won't yell at you for it, but it makes the DX a little more confusing.
2. don't think this needs to be a zod schema
3. why can't we just reuse the `ThreadStatus` type from the LangGraph SDK?

apps/web/src/lib/schemas/thread-status.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Thread status response schema for SWR typing
+ * Used by useThreadStatus hook and fetchThreadStatus service
+ */
+export const ThreadPollingResponseSchema = z.object({
```
bracesproul: same thing w/ naming, and not sure why this needs to be a zod object

apps/web/src/lib/schemas/thread-status.ts
```diff
@@ -0,0 +1,33 @@
...
+  graph: z.enum(["manager", "planner", "programmer"]),
```
bracesproul:
```suggestion
  graph: z.enum([MANAGER_GRAPH_ID, PLANNER_GRAPH_ID, PROGRAMMER_GRAPH_ID]),
```

apps/web/src/lib/schemas/thread-status.ts
```diff
@@ -0,0 +1,33 @@
...
+  runId: z.string(),
+  threadId: z.string(),
+  status: ThreadDisplayStatus,
+  taskPlan: z.any().optional(), // TaskPlan type from shared package
```
bracesproul: plz use the actual type, not any

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,297 @@
...
+  runId: string;
+  threadId: string;
+  status: ThreadDisplayStatus;
+  taskPlan?: any;
+}
+
+/**
+ * Service class for fetching thread status information
+ */
+export class ThreadStatusService {
```
bracesproul: this isn't used anywhere, why do we have it?

bracesproul (dismissed):

apps/web/src/hooks/useThreadsSWR.ts
starmorph: I originally had it as you just described. By passing the status as props there was logic for the status on both the server and client side which created 2 sources of truth. So by having the two hooks (explained below), we can maintain one source of truth for calculated status only from the server computed status. The UseThreadSWR doesn't distinguish between manager, planner, programmer, it checks for thread changes and then gets the single source of truth status from useThreadStatus when metadata needs revalidation.

### Before 
Server State: Authoritative status computed by fetchThreadStatus (complex logic across manager/planner/programmer)
UI State: Status cached in parent and passed as props (could become stale)
Result: Cache invalidation complexity, stale data, and manual state management

### Current Solution - Two Focused Hooks:
`useThreadsSWR`: Basic thread metadata (ID, title, repository, etc.)
- Simple, infrequent polling for thread list changes
- No complex status computation

`useThreadStatus`: Real-time computed status
- Handles complex business logic (manager > planner > programmer priority)
- Makes multiple API calls per thread to analyze session states
- Only runs when components actually need status (performance optimization)

### Benefits:
✅ Single Source of Truth: Status always comes directly from server via SWR
✅ Performance: Status computed only for visible/needed threads
✅ No Manual State Management: SWR handles caching, retries, staleness automatically
✅ Separation of Concerns: Thread data vs complex status computation are isolated
The fetchThreadStatus service is too complex (multiple API calls, session analysis, task completion logic) to cache in UI state without creating maintenance and synchronization problems.

apps/web/src/components/v2/types.ts
```diff
@@ -2,11 +2,11 @@ import { getThreadTitle } from "@/lib/thread";
...
 import { ManagerGraphState } from "@open-swe/shared/open-swe/manager/types";
+import { ThreadUIStatus } from "@/lib/schemas/thread-status";
 
-export interface ThreadDisplayInfo {
+export interface ThreadMetadata {
   id: string;
   title: string;
-  status: "running" | "completed" | "failed" | "pending";
   lastActivity: string;
   taskCount: number;
```
bracesproul, line 8: what does `lastActivity` mean here? And could we add a new field: `taskPlan?: TaskPlan` which contains the current task plan (active plans + all prev) from the programmer state, if defined?

apps/web/src/app/(v2)/chat/page.tsx
```diff
@@ -12,23 +12,22 @@ import { MANAGER_GRAPH_ID } from "@open-swe/shared/constants";
...
   return (
     <div className="bg-background h-screen">
       <Suspense>
         <Toaster />
         <GitHubAppProvider>
           <DefaultView
-            threads={displayThreads}
+            threads={threadMetadata}
             threadsLoading={threadsLoading}
+            originalThreads={threads}
```
bracesproul: tbh, i dont really like defining it twice. is there anything we can do to avoid this? Maybe pass the base threads through, and use a util to convert to the `ThreadMetadata` where we need it? this will end up making the DX more complex

apps/web/src/app/(v2)/chat/threads/page.tsx
```diff
@@ -1,60 +1,83 @@
...
-    pending: filteredThreads.filter((t) => t.status === "pending"),
-  };
-
-  const statusCounts = {
-    all: displayThreads.length,
-    running: displayThreads.filter((t) => t.status === "running").length,
-    completed: displayThreads.filter((t) => t.status === "completed").length,
-    failed: displayThreads.filter((t) => t.status === "failed").length,
-    pending: displayThreads.filter((t) => t.status === "pending").length,
+  const realGroupedThreads = {
```
bracesproul: why the var name change? imo `groupedThreads` makes more sense than `realGroupedThreads`

apps/web/src/components/v2/types.ts
```diff
@@ -22,37 +22,18 @@ export interface ThreadDisplayInfo {
   };
 }
 
-// Utility functions to convert between Thread and ThreadDisplayInfo
-export function threadToDisplayInfo(
+export interface ThreadDisplayInfo extends ThreadMetadata {
```
bracesproul: can we just add `status` to `ThreadMetadata` and drop this type?

apps/web/src/components/v2/types.ts
bracesproul: this will make implementing the task list UI much easier*

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+    timestamp: Date.now(),
+    type,
+  });
+}
+
+export async function fetchThreadStatus(
+  threadId: string,
+  lastPollingState: ThreadStatusData | null = null,
+  managerThreadData?: Thread<ManagerGraphState> | null,
+  sessionCache?: Map<string, any>,
```
bracesproul: this var is never used. not good.

looking to see if claude can implement for me

apps/web/src/hooks/useThreadsStatus.ts
```diff
@@ -0,0 +1,189 @@
...
+  running: number;
+  completed: number;
+  failed: number;
+  pending: number;
+  idle: number;
+  paused: number;
+  error: number;
+}
+
+interface GroupedThreads {
```
bracesproul: i pushed a commit renaming this from `GroupedThreads` to `GroupedThreadIds`. want to make sure it's clear from naming what a struct contains

apps/web/src/hooks/useThreadsStatus.ts
```diff
@@ -0,0 +1,189 @@
...
+  groupedThreads: GroupedThreadIds;
+  isLoading: boolean;
+  hasErrors: boolean;
+}
+
+const sessionDataCache = new Map<
+  string,
+  {
+    plannerData?: { thread: any; run: any };
+    programmerData?: { thread: any; run: any };
```
bracesproul: we should never use `any`. Can you add types for this please?

apps/web/src/lib/schemas/thread-status.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface ThreadStatusData {
+  graph:
+    | typeof MANAGER_GRAPH_ID
+    | typeof PLANNER_GRAPH_ID
+    | typeof PROGRAMMER_GRAPH_ID;
+  runId: string;
+  threadId: string;
+  status: ThreadUIStatus;
+  taskPlan?: TaskPlan; // Task plan data when available from programmer sessions
```
bracesproul, line 44: i had to push up a commit changing this from `any` to `TaskPlan`. plz make sure you're always adding proper typing

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+/**
+ * Determines if all tasks in a task plan are completed
+ */
+function areAllPlanItemsCompleted(taskPlan: TaskPlan): boolean {
+  if (!taskPlan?.tasks || !Array.isArray(taskPlan.tasks)) {
+    return false;
+  }
+  const activePlanItems = getActivePlanItems(taskPlan);
+  return activePlanItems.every((planItem) => planItem.completed);
+}
```
bracesproul, line 31: i pushed a commit fixing this func. before it was checking all tasks were completed, which will not always be the case if it creates a new task before the first is complete. instead, we should be checking for the active task plan and confirming that all of the plan items are complete

bracesproul: i also had to add proper typing here. before it was typed as `any`

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    return programmer;
+  }
+}
+
+const CACHE_TTL = 30 * 1000;
+
+function getCachedSessionData(
+  sessionCache: Map<string, any> | undefined,
```
bracesproul: fix typing to remove `any`

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+  sessionKey: string,
+): any | null {
```
bracesproul: same here & everywhere else in your PR you've defined an `any`

apps/web/src/services/thread-status.service.ts
bracesproul: pushed up a commit claude generated which uses this var. remember to catch things like this before opening the PR

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+      graph,
+      runId,
+      threadId: errorThreadId,
+      status: "error",
+    };
+  }
+}
+
+async function checkLastKnownGraph(
+  client: ReturnType<typeof createClient>,
```
bracesproul: import from langgraph sdk
```suggestion
  client: Client,
```

apps/web/src/services/thread-status.service.ts
```diff
@@ -0,0 +1,425 @@
...
+
+      return null;
+    }
+  }
+
+  return null;
+}
+
+async function performFullStatusCheck(
+  client: ReturnType<typeof createClient>,
```
bracesproul:
```suggestion
  client: Client,
```

package.json
```diff
@@ -23,5 +23,8 @@
     "@langchain/langgraph-sdk": "^0.0.95",
     "@langchain/core": "^0.3.58"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "swr": "^2.3.4"
```
bracesproul: dont add global deps. always should be package/app scoped

bracesproul (changes requested):
a handful of comments, mainly code style. looks to be implemented properly though!

apps/web/src/components/v2/types.ts
bracesproul: following up: looks like we fetch this in the service worker. you should figure out how to expose this up to the hook which fetches the data and returns this type

starmorph: lastActivity is the little text at the bottom of a thread card that says "16 min ago" or last time the thread was updated. And yes will add taskPlan and look into integrating type into hook now

apps/web/src/services/thread-status.service.ts
starmorph: thank you

apps/web/src/app/(v2)/chat/[thread_id]/page.tsx
```diff
@@ -28,13 +29,30 @@ export default function ThreadPage({
     assistantId: MANAGER_GRAPH_ID,
     threadId: thread_id,
     reconnectOnMount: true,
-    fetchStateHistory: false,
```
Comment (bracesproul) [inline apps/web/src/app/(v2)/chat/[thread_id]/page.tsx:31]: Why did you delete this?

apps/web/src/app/(v2)/chat/[thread_id]/page.tsx
```diff
@@ -43,16 +61,14 @@ export default function ThreadPage({
     return <ThreadViewLoading onBackToHome={handleBackToHome} />;
   }
 
-  // Convert all threads to display format
-  const displayThreads: ThreadDisplayInfo[] = threads.map(threadToDisplayInfo);
-  const currentDisplayThread = threadToDisplayInfo(thread);
+  // Threads metadata is already computed by the hook
```
Comment (bracesproul) [inline apps/web/src/app/(v2)/chat/[thread_id]/page.tsx]: unnecessary
```suggestion
```

apps/web/src/hooks/useThreadsSWR.ts
```diff
@@ -1,30 +1,38 @@
...
-) {
+/**
+ * Hook for fetching manager threads.
+ * This hook is specifically designed for ManagerGraphState threads,
+ * which are the top-level threads in the Open SWE system.
+ *
+ * For UI display, use `threadsToMetadata(threads)` utility to convert
+ * raw threads to ThreadMetadata objects.
+ */
+export function useThreadsSWR(options: UseThreadsSWROptions = {}) {
```
bracesproul: why isn't this generic anymore? you can configure the graph ID passed, so we prob shouldn't be hard coding the `ManagerGraphState` as the state type for it?

apps/web/src/hooks/useThreadsStatus.ts
```diff
@@ -0,0 +1,192 @@
...
+import { THREAD_STATUS_SWR_CONFIG } from "@/lib/swr-config";
+import { useMemo, useRef } from "react";
+import { Thread } from "@langchain/langgraph-sdk";
+import { ManagerGraphState } from "@open-swe/shared/open-swe/manager/types";
+import { PlannerGraphState } from "@open-swe/shared/open-swe/planner/types";
+import { GraphState } from "@open-swe/shared/open-swe/types";
+
+export interface SessionCacheData {
+  plannerData?: { thread: Thread<PlannerGraphState>; run: null };
+  programmerData?: { thread: Thread<GraphState>; run: null };
```
bracesproul: if run is always null just drop it?
```suggestion
  plannerData?: { thread: Thread<PlannerGraphState> };
  programmerData?: { thread: Thread<GraphState> };
```

PR #394, opened 2025-07-11
hack: Subgraph streaming messages support

bracesproul:
pretty annoying hack. ideally we can remove this when proper support for streaming subgraph messages is implemented

### DIFF
--- a/apps/web/src/components/v2/actions-renderer.tsx
+++ b/apps/web/src/components/v2/actions-renderer.tsx
@@ -298,7 +298,6 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
 
   useEffect(() => {
     debouncedSetMessages(stream.messages);
-    // Return cleanup function to cancel any pending debounced calls when unmounting
     return () => {
       debouncedSetMessages.cancel();
     };


apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -242,6 +290,19 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
...
   }, [stream.values, graphId]);
 
+  const debouncedSetMessages = useRef(
+    debounce((messages: Message[]) => {
+      setMergedMessages((prev) => addMessagesToState(prev, messages));
+    }, 100),
+  ).current;
+
+  useEffect(() => {
+    debouncedSetMessages(stream.messages);
```
bracesproul, line 300: added `debounce` since this was giving a max update depth exceeded error when the graph would first load (but never after that)

ugh

apps/web/src/components/v2/actions-renderer.tsx
```diff
@@ -119,6 +155,18 @@ export function ActionsRenderer<State extends PlannerGraphState | GraphState>({
         setCustomNodeEvents((prev) => [...prev, event]);
       }
     },
+    onLangChainEvent: (data) => {
```
bracesproul, line 158: this hack is only required for the reviewer subgraph, so we only need to update events when we call those nodes

PR #414, opened 2025-07-15
Feat:  MCP UI component

starmorph:
Screenshot 2025-07-15 at 2 10 10 PM

### DIFF
--- a/apps/web/src/components/gen-ui/action-step.tsx
+++ b/apps/web/src/components/gen-ui/action-step.tsx
@@ -12,6 +12,7 @@ import {
   XCircle,
   Loader2,
   Globe,
+  Zap,
   FileCode,
   CloudDownload,
 } from "lucide-react";
@@ -101,6 +102,13 @@ type SearchActionProps = BaseActionProps &
     errorCode?: number;
   };
 
+type McpActionProps = BaseActionProps & {
+  actionType: "mcp";
+  toolName: string;
+  args: Record<string, any>;
+  output?: string;
+};
+
 export type ActionItemProps =
   | (BaseActionProps & { status: "loading" })
   | ShellActionProps
@@ -108,6 +116,7 @@ export type ActionItemProps =
   | InstallDependenciesActionProps
   | PlannerNotesActionProps
   | GetURLContentActionProps
+  | McpActionProps
   | SearchActionProps;
 
 export type ActionStepProps = {
@@ -213,7 +222,11 @@ function ActionItem(props: ActionItemProps) {
     }
 
     if (props.status === "generating") {
-      return ACTION_GENERATING_TEXT_MAP[props.actionType];
+      if (props.actionType === "mcp") {
+        return `Running ${props.toolName}...`;
+      } else {
+        return ACTION_GENERATING_TEXT_MAP[props.actionType];
+      }
     }
 
     if (props.status === "done") {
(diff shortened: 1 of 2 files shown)


apps/web/src/components/gen-ui/action-step.tsx
```diff
@@ -213,7 +222,11 @@ function ActionItem(props: ActionItemProps) {
     }
 
     if (props.status === "generating") {
-      return ACTION_GENERATING_TEXT_MAP[props.actionType];
+      if (props.actionType === "mcp") {
+        return `Running ${props.toolName}...`;
+      } else {
+        return ACTION_GENERATING_TEXT_MAP[props.actionType];
```
bracesproul: just add a new entry to the `ACTION_GENERATING_TEXT_MAP` for `mcp`
```suggestion
      return ACTION_GENERATING_TEXT_MAP[props.actionType];
```

apps/web/src/components/gen-ui/action-step.tsx
```diff
@@ -418,6 +446,31 @@ function ActionItem(props: ActionItemProps) {
...
+            <div className="text-muted-foreground mt-1 flex flex-wrap gap-x-3 gap-y-1 text-xs">
+              {keyArgs.map(([key, value]) => {
+                const displayValue =
+                  typeof value === "string" && value.length > 50
+                    ? value.slice(0, 50) + "..."
+                    : String(value);
+                return (
+                  <span key={key}>
+                    <span className="font-medium">{key}:</span> {displayValue}
+                  </span>
```
bracesproul: what does this look like if there's 10+ args? I'd assume it'll overflow a look a little off. could you test locally and if it looks off implement a ui fix?

bracesproul: Maybe we just dont include key + truncated value in the header? if we're already showing it below not sure if we need to duplicate it here

apps/web/src/components/gen-ui/action-step.tsx
```diff
@@ -451,6 +504,55 @@ function ActionItem(props: ActionItemProps) {
           )}
         </div>
       );
+    } else if (props.actionType === "mcp") {
+      const hasArgs = props.args && Object.keys(props.args).length > 0;
+      const hasOutput = props.output;
```
bracesproul: good to make it an actual boolean
```suggestion
      const hasOutput = !!props.output;
```

apps/web/src/components/gen-ui/action-step.tsx
```diff
@@ -451,6 +504,55 @@ function ActionItem(props: ActionItemProps) {
...
+
+      return (
```
bracesproul, line 466: should we just be returning `null` if there's no args & no output rather than this empty div?

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -104,6 +104,20 @@ type ConversationHistorySummaryToolArgs = z.infer<
...
+  const knownToolNames = [
+    shellTool.name,
+    applyPatchTool.name,
+    installDependenciesTool.name,
+    plannerNotesTool.name,
+    getURLContentTool.name,
+    openPrTool.name,
+    diagnoseErrorTool.name,
+  ];
+  return !knownToolNames.includes(toolName);
```
bracesproul: slightly better check. what you have probably won't give us issues, but technically it could if the `toolName` exists as a substring of one of our main tools.
```suggestion
  return !knownToolNames.some((t) => t === toolName);
```

apps/web/src/components/thread/messages/ai.tsx
```diff
@@ -248,6 +262,18 @@ export function mapToolMessageToActionStepProps(
       reasoningText,
     };
   }
+  // Handle MCP tools
+  if (toolCall && isMcpTool(toolCall.name)) {
```
bracesproul: make this an `else if` to match above

bracesproul (dismissed):
high level lgtm, couple nits & questions

bracesproul (approved):
ty!

PR #433, opened 2025-07-16
feat: Added LLM fallback

aliyanishfaq:
# Feat: Added LLM fallback

Refactored the model loading system to use a new `ModelManager` class that adds fallback capabilities
### `apps/open-swe/src/utils/load-model.ts`
- **Refactored**: Replaced direct `initChatModel` calls with new `ModelManager` system

### `apps/open-swe/src/utils/model-manager.ts`
- **Circuit Breaker Pattern**: Circuit breaker with configurable failure threshold (default: 3 failures)
- **Fallback**: `anthropic → google-genai → openai` with configurable order
### `apps/open-swe/src/utils/runtime-fallback.ts`
- Extended `ConfigurableModel` class to have fallback functionality during runtime.

### DIFF
--- a/apps/open-swe/src/utils/runtime-fallback.ts
+++ b/apps/open-swe/src/utils/runtime-fallback.ts
@@ -53,7 +53,7 @@ export class FallbackRunnable<
     messages: BaseMessage[],
     options?: Record<string, any>,
   ): Promise<ChatResult> {
-    const result = await this.invoke(messages as any, options);
+    const result = await this.invoke(messages, options);
     const generation: ChatGeneration = {
       message: result,
       text: result?.content ? getMessageContentString(result.content) : "",
@@ -131,10 +131,13 @@ export class FallbackRunnable<
       this.config,
       this.task,
       this.modelManager,
-    );
+    ) as unknown as ConfigurableModel<RunInput, CallOptions>;
   }
 
-  withConfig(config?: RunnableConfig): any {
+  // @ts-expect-error - types are hard man :/
+  withConfig(
+    config?: RunnableConfig,
+  ): ConfigurableModel<RunInput, CallOptions> {
     const configuredPrimary =
       this.primaryRunnable.withConfig?.(config) ?? this.primaryRunnable;
     return new FallbackRunnable(
@@ -142,7 +145,7 @@ export class FallbackRunnable<
       this.config,
       this.task,
       this.modelManager,
-    );
+    ) as unknown as ConfigurableModel<RunInput, CallOptions>;
   }
 
   private getPrimaryModel(): ConfigurableModel {
--- a/apps/open-swe/src/graphs/planner/nodes/rewrite-plan.ts
+++ b/apps/open-swe/src/graphs/planner/nodes/rewrite-plan.ts
@@ -4,7 +4,6 @@
 import { GraphConfig, PlanItem } from "@open-swe/shared/open-swe/types";
 import { z } from "zod";
 import { tool } from "@langchain/core/tools";
-import { ConfigurableModel } from "langchain/chat_models/universal";
 import { traceable } from "langsmith/traceable";
 import {
   PlannerGraphState,
@@ -19,6 +18,7 @@ import {
   supportsParallelToolCallsParam,
   Task,
 } from "../../../utils/load-model.js";
+import { FallbackRunnable } from "../../../utils/runtime-fallback.js";
 
 const systemPromptIdentifyChanges = `You are operating as an agentic coding assistant built by LangChain. You've previously been given a task to generate a plan of action for, to address the user's initial request.
 
@@ -97,7 +97,7 @@ const formatSysPromptRewritePlan = (
 
 async function identifyTasksToModifyFunc(
   state: PlannerGraphState,
-  model: ConfigurableModel,
+  model: FallbackRunnable,
   supportsParallelToolCallsParam: boolean,
 ): Promise<PlanItem[]> {
   if (!state.planChangeRequest) {
@@ -188,7 +188,7 @@ const identifyTasksToModify = traceable(identifyTasksToModifyFunc, {
 async function updatePlanTasksFunc(
   state: PlannerGraphState,
   tasksToModify: PlanItem[],
-  model: ConfigurableModel,
+  model: FallbackRunnable,
   supportsParallelToolCallsParam: boolean,
 ): Promise<string[]> {
   if (!state.planChangeRequest) {


apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+  temperature?: number;
+  maxTokens?: number;
+  thinkingModel?: boolean;
+  thinkingBudgetTokens?: number;
+}
+
+export enum CircuitState {
+  CLOSED = "CLOSED", // Normal operation
+  OPEN = "OPEN", // Failing, use fallback
+}
```
bracesproul: in js when adding comments about a type like this, always prefer JSDoc comments so you can read them in your IDE without having to actually navigate to the file (hovering over the type/arg in your IDE renders the JSDoc)
```suggestion
export enum CircuitState {
  /**
  * Normal operation
  */
  CLOSED = "CLOSED",
  /**
  * Failing, use fallback
  */
  OPEN = "OPEN",
}
```
please update the rest in this file to follow this convention

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+};
+
+export class ModelManager {
+  private config: ModelManagerConfig;
+  private circuitBreakers: Map<string, CircuitBreakerState> = new Map();
+
+  constructor(config: Partial<ModelManagerConfig> = {}) {
+    this.config = { ...DEFAULT_MODEL_MANAGER_CONFIG, ...config };
+
+    logger.info("ModelManager initialized", {
```
bracesproul: nit, since you alr set the logger name to `ModelManager`, this will duplicate it in the log.
```suggestion
    logger.info("Initialized", {
```

same w/ other logs in this file

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+    });
+  }
+
+  /**
+   * Load a single model (no fallback during loading)
+   */
+  async loadModel(graphConfig: GraphConfig, task: Task) {
+    const baseConfig = this.getBaseConfigForTask(graphConfig, task);
+
+    if (baseConfig.modelName) {
```
bracesproul: `baseConfig.modelName` is a required arg, shouldn't need the if statement here

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+        const model = await this.initializeModel(
+          {
+            ...baseConfig,
+            temperature: baseConfig.temperature,
+            maxTokens: baseConfig.maxTokens,
+            thinkingModel: baseConfig.thinkingModel,
+            thinkingBudgetTokens: baseConfig.thinkingBudgetTokens,
+          },
+          graphConfig,
+        );
```
bracesproul: nit, you shouldn't need to manually define these args since you're spreading the object already
```suggestion
        const model = await this.initializeModel(baseConfig, graphConfig);
```

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+          modelName: baseConfig.modelName,
+          error: error instanceof Error ? error.message : String(error),
+        });
+        throw error;
+      }
+    }
+
+    const errorMsg = `No model available for task ${task}`;
+    logger.error(errorMsg, { task });
+    throw new Error(errorMsg);
```
bracesproul: two nits:
- no need to define task twice in the error message. either include in string or in the data field.
- if we're throwing an error which won't be caught, and will cause the entire system to error like we're doing here, we shouldn't add a log & throw error since that'll just duplicate the logs. only throwing an error is fine here

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+  /**
+   * Initialize the model instance
+   */
+  // should be set to return configurable chat model
+  public async initializeModel(
```
bracesproul: remember to drop comments after they've been resolved
```suggestion
  /**
   * Initialize the model instance
   */
  public async initializeModel(
```

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+          providerToApiKey(provider, apiKeys),
+          secretsEncryptionKey,
+        );
+        if (!apiKey) {
+          throw new Error("No API key found for provider: " + provider);
+        }
+      }
+    }
+
+    const modelOptions: any = {
```
bracesproul: ideally this isn't typed as `any`. I would switch this to one of two:
1. better than `any`, but still not great would be typing it as `Record<string, any>` so at least we're ensuring it'll always be an object
2. extracting the type from `initChatModel` so it's typed exactly how `initChatModel` expects it:
```typescript
type InitChatModelArgs = Parameters<typeof initChatModel>[1];

...

const modelOptions: InitChatModelArgs = { ... }
```

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+            thinking: { budget_tokens: thinkingBudgetTokens, type: "enabled" },
+            maxTokens: thinkingMaxTokens,
+          }
+        : { maxTokens: finalMaxTokens }),
+    };
+
+    logger.debug("Initializing model", {
+      provider,
+      modelName,
+      options: modelOptions,
```
bracesproul: dont log the whole `modelOptions` object since that could contain a users API key. either ensure it's filtered out from the log or dont log the `modelOptions` obj entirely

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+    config: GraphConfig,
+    task: Task,
+    selectedModel: BaseChatModel,
+  ) {
+    const configs: ModelLoadConfig[] = [];
+    const baseConfig = this.getBaseConfigForTask(config, task);
+
+    // Get selected model config
+    const modelInstance = selectedModel as any;
+    const defaultConfig = modelInstance._defaultConfig;
```
bracesproul: `_defaultConfig` is a `ConfigurableModel` field, so we should be typing this as `ConfigurableModel`, not `any`/`BaseChatModel`

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+    let selectedModelConfig: ModelLoadConfig | null = null;
+
+    if (defaultConfig) {
+      const provider = defaultConfig.modelProvider as Provider;
+      const modelName = defaultConfig.model;
```
bracesproul, line 194: the typical langchain convention is `modelName`, not `model`. Did this work as intended? I suspect it needs to be changed to `modelName`

bracesproul: ah `initChatModel` appears to use `model` instead.

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+
+      if (provider && modelName) {
+        selectedModelConfig = {
+          provider,
+          modelName,
+          temperature: defaultConfig.temperature ?? baseConfig.temperature,
+          maxTokens: defaultConfig.maxTokens ?? baseConfig.maxTokens,
+          thinkingModel:
+            baseConfig.thinkingModel && this.supportsThinking(provider),
+          thinkingBudgetTokens: THINKING_BUDGET_TOKENS,
```
bracesproul: i believe we should only be passing these args if it is a thinking model? Passing extra args can sometimes cause the LLM API to throw an error (e.g. instead of just silently ignoring them, the Azure LLM API will throw if you include any additional args which aren't supported)

bracesproul: apply same change below

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+        [Task.SUMMARIZER]: "gemini-2.5-pro",
+      },
+      openai: {
+        [Task.PROGRAMMER]: "gpt-4o",
+        [Task.ROUTER]: "gpt-4o-mini",
+        [Task.SUMMARIZER]: "gpt-4o",
+      },
+    };
+
+    const modelName = defaultModels[provider]?.[task];
```
bracesproul: nit
```suggestion
    const modelName = defaultModels[provider][task];
```

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+      thinkingModel: provider === "openai" && modelName.startsWith("o"),
+      thinkingBudgetTokens: THINKING_BUDGET_TOKENS,
+    };
+  }
+
+  /**
+   * Circuit breaker methods
+   */
+  public isCircuitClosed(modelKey: string): boolean {
+    const state = this.getCircuitState(modelKey);
```
bracesproul, line 325: we should have a better name than `state` here, since this obj already has a field named `state`. Maybe something like `circuitState`?

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+    const circuitState = this.getCircuitState(modelKey);
+
+    circuitState.state = CircuitState.CLOSED;
+    circuitState.failureCount = 0;
+    delete circuitState.openedAt;
+
+    logger.debug(
+      `${modelKey}: Circuit breaker reset after successful request`,
+      {
+        modelKey,
```
bracesproul: same nit as above, dont need to duplicate `modelKey`

bracesproul: & apply change to rest of logs

apps/open-swe/src/utils/model-manager.ts
```diff
@@ -0,0 +1,458 @@
...
+            now + this.config.circuitBreakerTimeoutMs,
+          ).toISOString(),
+        },
+      );
+    }
+  }
+
+  private supportsThinking(provider: Provider): boolean {
+    return provider === "anthropic" || provider === "openai";
+  }
```
bracesproul: not sure how useful this method is? Google supports thinking too and you already have the `baseConfig.thinkingModel` boolean. I might drop this tbh

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+      disableStreaming: false,
+    });
+    this.primaryRunnable = primaryRunnable;
+    this.config = config;
+    this.task = task;
+    this.modelManager = modelManager;
+  }
+
+  lc_serializable = false;
+  lc_namespace = ["open-swe", "fallback"];
```
bracesproul: these arent required, we should drop
```suggestion
```

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  async _generate(messages: BaseMessage[], options?: any): Promise<ChatResult> {
```
bracesproul: Prefer unknown object over any. Please do the same w/ all other instances of this below
```suggestion
  async _generate(messages: BaseMessage[], options?: Record<string, any>): Promise<ChatResult> {
```

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+    const result = await this.invoke(messages as any, options);
+    const generation: ChatGeneration = {
+      message: result,
+      text:
+        typeof result?.content === "string"
+          ? result.content
+          : JSON.stringify(result?.content) || "",
```
bracesproul: the `getMessageContentString` util will properly handle converting the message content to a str
```suggestion
      text: result?.content ? getMessageContentString(result.content) : ""
```

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+    };
+    return {
+      generations: [generation],
+      llmOutput: {},
+    };
+  }
+
+  async invoke(input: any, options?: Partial<RunnableConfig>): Promise<any> {
```
bracesproul: avoid `any` whenever possible. `BaseLanguageModelInput` will work here. I'm also going to push up some more type changes to this func to give it better typing

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+      }
+    }
+
+    throw new Error(
+      `All fallback models exhausted for task ${this.task}. Last error: ${lastError?.message}`,
+    );
+  }
+
+  bindTools(
+    tools: any[],
```
bracesproul: same here. I'm pushing up a fix
```suggestion
    tools: BindToolsInput[],
```

apps/open-swe/src/utils/runtime-fallback.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (current && typeof current._llmType !== "undefined") {
+      return current;
+    }
+
+    throw new Error(
+      "Could not extract primary model from runnable - no _llmType found",
+    );
+  }
+
+  private extractBoundTools(): ExtractedTools | null {
```
bracesproul, line 169: this kinda worries me, if `_queuedMethodOperations` exists on `current`, that indicates it's an instance of `ConfigurableModel`. However at the end, you're calling `current.bound`, which makes it an instance of `RunnableBinding`

Has this method been tested?

bracesproul (changes requested):
- if we fallback to a provider that provider should persist across all requests until the fallback expires. e.g. if a google model fails and we fallback to openai, the next req which uses the same `Task` should use the provider we just fell back to (openai in this case), instead of trying on google again.

apps/open-swe/src/utils/model-manager.ts
aliyanishfaq: fixed, was added since was using properties from some other function. Don't need it now

apps/open-swe/src/utils/model-manager.ts
aliyanishfaq: removed from files

apps/open-swe/src/utils/runtime-fallback.ts
aliyanishfaq: It can either be `RunnableBinding` or `_queuedMethodOperations`. The method has been tested.

apps/open-swe/src/utils/model-manager.ts
aliyanishfaq: yup, it works for me

PR #452, opened 2025-07-21
fix: temporarily disable requests

bracesproul:


apps/open-swe/src/security/auth.ts
```diff
@@ -1,24 +1,24 @@
...
-  API_KEY_REQUIRED_MESSAGE,
-  GITHUB_INSTALLATION_NAME,
-  GITHUB_INSTALLATION_TOKEN_COOKIE,
-  GITHUB_TOKEN_COOKIE,
-  GITHUB_USER_ID_HEADER,
-  GITHUB_USER_LOGIN_HEADER,
-} from "@open-swe/shared/constants";
-import { decryptSecret } from "@open-swe/shared/crypto";
-import { verifyGitHubWebhookOrThrow } from "./github.js";
+// import {
```
eyurtsev, line 2: Could we delete the comented out code?

bracesproul: fixed in main

PR #469, opened 2025-07-22
feat: add initial updates to create cli 

Palashio:
This PR is the first set of updates to add the CLI tool to `Open SWE`. 

- Creates the CLI interface using Ink
- The interface starts with a login question, and if anything other than yes it'll exist. in the case where you do specify yes, it'll start up the authentication server for github oauth. right now it's hardcoded to the client secret that i received for the open swe dev cli. 
- Adds the Github auth server. This is a quick express JS server that's spun up just for the purposes of authentication. 
- The authentication token is stored in a github token json file with mode `0o600`, which means that only the user can access/edit the file. This will be used in the future for persistence. Checking whether the access token is still valid has not been implemented yet (will be in followup PR).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8205,13 +8205,13 @@ __metadata:
   linkType: hard
 
 "axios@npm:^1.6.1, axios@npm:^1.6.8, axios@npm:^1.8.3":
-  version: 1.10.0
-  resolution: "axios@npm:1.10.0"
+  version: 1.11.0
+  resolution: "axios@npm:1.11.0"
   dependencies:
     follow-redirects: ^1.15.6
-    form-data: ^4.0.0
+    form-data: ^4.0.4
     proxy-from-env: ^1.1.0
-  checksum: b5fd840d499469bf968e44b8ac96f4b363c6aa4c791a50834c086a7cffbc2d77fe24f27af1aba46c3e1f4840aaf991461fc27537990596b93dea0f4df3245a86
+  checksum: 0a33dc600b588bfd3111b198d5985527ed89f722817455d7cdb66c1d055e5f8859cc2bebb7320888957fc8458ebe77d5f83af02af9cd260217c91c4e92b6dfb6
   languageName: node
   linkType: hard
 
@@ -11502,7 +11502,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"form-data@npm:^4.0.0":
+"form-data@npm:^4.0.0, form-data@npm:^4.0.4":
   version: 4.0.4
   resolution: "form-data@npm:4.0.4"
   dependencies:


apps/cli/package-lock.json
bracesproul, line 1: drop in favor of yarn

apps/cli/package.json
bracesproul, line 1: we should use the same base package.json as the other apps/packages. There isn't a standard (yet), so for now I'd start with copying what's in the `packages/shared/package.json` file, and updating the deps to remove whats not used

Do this so that we have standard scripts, linters, tests, etc

apps/cli/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "cli",
```
bracesproul: naming convention in JS monorepos should be `@<overall_name>/<app_name>`

in this case it should be:
```suggestion
  "name": "@open-swe/cli",
```

apps/cli/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
bracesproul: mit license
```suggestion
  "license": "MIT",
```

apps/cli/src/TerminalInterface.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Box, Text } from "ink";
+
+interface TerminalInterfaceProps {
```
bracesproul, line 4: should be 2 space indents. once you bring in the other formatters/linters it'll auto-fix for you

apps/cli/src/TerminalInterface.tsx
```diff
@@ -0,0 +1,31 @@
...
+	submitted: string | null;
+	setSubmitted: (value: string) => void;
```
bracesproul: this naming is a little weird to me? shouldn't it be `input` or `message`?

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
+import dotenv from 'dotenv';
+dotenv.config();
```
bracesproul: nit:
```suggestion
import 'dotenv/config';
```
does the same thing

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+import express from 'express';
+import fetch from 'node-fetch';
```
bracesproul: why `node-fetch` instead of just `fetch`?

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+const CLIENT_ID = process.env.GITHUB_CLIENT_ID || '';
+const CLIENT_SECRET = process.env.GITHUB_CLIENT_SECRET || '';
+const PORT = process.env.PORT || 3456;
+const CALLBACK_URL = process.env.GITHUB_CALLBACK_URL || `http://localhost:${PORT}/api/auth/github/callback`;
+
+const TOKEN_PATH = path.join(os.homedir(), '.open-swe-cli', 'github_token.json');
+
+let accessToken: string | null = null;
+
+function saveToken(tokenData: any) {
```
bracesproul: avoid `any` whenever possible
this should prob be `Record<string, any>`, but ideally it has a proper type

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+  fs.writeFileSync(TOKEN_PATH, JSON.stringify(tokenData, null, 2), { mode: 0o600 });
+}
+
+function loadToken() {
+  if (fs.existsSync(TOKEN_PATH)) {
+    return JSON.parse(fs.readFileSync(TOKEN_PATH, 'utf8'));
+  }
+  return null;
+}
```
bracesproul: might want to wrap this (and all the other `fs` calls above) in a `try/catch` incase something goes astray with the files, so it doesnt crash the server

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const app = express();
+
+// 1. Start OAuth flow
+app.get('/api/auth/github/login', (_req: Request, res: Response) => {
+  const state = Math.random().toString(36).substring(2);
+  const githubAuthUrl = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&redirect_uri=${encodeURIComponent(CALLBACK_URL)}&state=${state}`;
```
bracesproul: prefer structured URL composition 
```suggestion
  const baseGithubAuthUrl = "https://github.com/login/oauth/authorize";
  const queryParams = new URLSearchParams({
    client_id: CLIENT_ID,
    redirect_uri: CALLBACK_URL,
    state,
  });
  const githubAuthUrl = `${baseGithubAuthUrl}?${queryParams.toString()}`;
  
```

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+  res.redirect(githubAuthUrl);
```
bracesproul: does this need to be returned, or no?

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+    headers: { Accept: 'application/json', 'Content-Type': 'application/json' },
+    body: JSON.stringify({
+      client_id: CLIENT_ID,
+      client_secret: CLIENT_SECRET,
+      code,
+      redirect_uri: CALLBACK_URL,
+    }),
+  });
+  const tokenData = await tokenRes.json();
+  if (tokenData.error) {
```
bracesproul, line 90: `.ok` will be true if the fetch res is between 200-299. good to include (although your `.error` check might be doing the same thing?
```suggestion
  const tokenData = await tokenRes.json();
  if (tokenData.error || !tokenRes.ok) {
```

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+    return res.status(400).send('Error exchanging code for token: ' + tokenData.error);
+  }
+  accessToken = tokenData.access_token;
+  // Store the token in a config file
+  try {
+    saveToken(tokenData);
+  } catch (err) {
+    console.error('Failed to store token in config file:', err);
+  }
+  res.send('Authentication successful! You can close this window.');
```
bracesproul: same as above on whether or not it needs to be returned

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,84 @@
...
+export function startAuthServer() {
+  app.listen(PORT, () => {});
+}
+
+export function getAccessToken() {
+  // Try to get from memory first, then from config file
+  if (accessToken) return accessToken;
+  const tokenData = loadToken();
+  return tokenData ? tokenData.access_token : null;
+}
```
bracesproul: should this assign the `accessToken` a value after calling `loadToken` the first time?

apps/cli/src/index.tsx
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env node
```
bracesproul, line 1: will this work even thought his is a `tsx` file? (not sure myself)

apps/cli/src/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+	return (
+		<Box>
+			<Text color="cyan">&gt; {value}</Text>
+		</Box>
+	);
+};
+
+const App: React.FC = () => {
+	const [authPrompt, setAuthPrompt] = useState<null | boolean>(null);
```
bracesproul: why not just `boolean`?

apps/cli/tsconfig.json
bracesproul, line 1: prefer using an existing `tsconfig.json`, then modifying it if you need

bracesproul (changes requested):
overall looks good, mainly code style nits.

for now, copy the tsconfig/package.json/lint/formatter files from `packages/shared` and make modifications if needed. ill work on putting together a standards doc for next time

apps/cli/src/TerminalInterface.tsx
Palashio: changing to message

apps/cli/src/auth-server.ts
Palashio: `node-fetch` ensures backwards compatibility (older node versions) but don't think we need, will change to fetch.

apps/cli/src/auth-server.ts
Palashio: creating a `GithubTokenResponse` type.

apps/cli/src/auth-server.ts
Palashio: good call

apps/cli/src/auth-server.ts
Palashio: so right now the cli and the server are just communicating through the filesystem so i don't think it needs to bereturned. we don't want to spin down the server after (because we'll need it for app installation) and so i thought this was fine.

apps/cli/src/auth-server.ts
Palashio: i added a polling mechanism on the frontend, so that it'll check for the token to exist in the filesystem. basically because the token is recieved in /callback we need a way to communicate that the token has been created.

apps/cli/src/auth-server.ts
Palashio: it does not the frontend just polls the config file for the access token so once auth is complete it'll log you into the cli

apps/cli/eslint.config.js
```diff
@@ -0,0 +1,41 @@
...
+    plugins: {
+      "@typescript-eslint": tseslint,
+      "react-hooks": reactHooks,
+      "react-refresh": reactRefresh,
+    },
+    rules: {
+      ...reactHooks.configs.recommended.rules,
+      "@typescript-eslint/no-explicit-any": 0,
+      "@typescript-eslint/no-unused-vars": [
+        "warn",
```
bracesproul: lets make this error, good to have (i need to update this in the rest of the apps)
```suggestion
        "error",
```

apps/cli/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@open-swe/cli",
+  "version": "1.0.0",
+  "license": "MIT",
+  "private": true,
+  "type": "module",
+  "main": "index.js",
+  "scripts": {
```
bracesproul, line 8: how do you start the CLI? Would be good to have a `dev` script for staring in dev mode, and adding the `bin` object with the script to start once we've released on NPM. see the [langchainjs scripts](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-scripts/package.json) package for an example of this

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,111 @@
+import "dotenv/config";
```
bracesproul, line 1: ideally we dont need to ship w/ the dotenv import like this. small nit and dont need to worry about it rn, but before we ship to prod we should investigate a way to load env vars (prob still via dotenv), but through some cli setup script

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+const app = express();
+
+// 1. Start OAuth flow
+app.get("/api/auth/github/login", (_req: Request, res: Response) => {
+  const state = Math.random().toString(36).substring(2);
+  const baseGithubAuthUrl = "https://github.com/login/oauth/authorize";
+  const url = new URL(baseGithubAuthUrl);
+  url.searchParams.set("client_id", CLIENT_ID);
```
bracesproul, line 62: nice. much cleaner

apps/cli/src/auth-server.ts
```diff
@@ -0,0 +1,111 @@
...
+  url.searchParams.set("redirect_uri", CALLBACK_URL);
+  url.searchParams.set("state", state);
+  res.redirect(url.toString());
```
bracesproul: good practice to return this so that we dont accidentally add code which executes after the response has been sent
```suggestion
  return res.redirect(url.toString());
```
and update the same w/ the one below

apps/cli/src/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+      process.exit(0);
+    }
+  }, [exit]);
+
+  // Start auth server and open browser when user says yes
+  useEffect(() => {
+    if (authPrompt === true && !authStarted) {
+      setAuthStarted(true);
+      startAuthServer();
+      open("http://localhost:3000/api/auth/github/login");
```
bracesproul: could we make this an env var? Once this is in prod it'll need to call a server which isnt running on the users local machine

bracesproul: same w/ the other github auth api URLs

bracesproul (commented):
lgtm overall, a couple quick comments. I also pushed a commit adding some more files, only one you need to act on is the `.env.example`. should add the env var names which are required to run this in dev (I think that's just GH vars for now)

apps/cli/src/auth-server.ts
Palashio: i'll create a task to create this setup script as a part of a cli implementation.

apps/cli/package.json
Palashio: i'll also create a task here

apps/cli/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@open-swe/cli",
+  "version": "1.0.0",
```
bracesproul: final nit lol. dont want to forger this then cut a release where the version is `1.0.0` (ive done this more times than id like to admit 🙃)
```suggestion
  "version": "0.0.0",
```

bracesproul (approved):
ty!

PR #473, opened 2025-07-22
ci: Better linter rules

bracesproul:


bracesproul:
Closing for now. will likely want to come back to this eventually

PR #494, opened 2025-07-23
feat: Thread title proposed plan

### LINKED ISSUE #296: Thread title should come from the plan title generated in proposed plan

starmorph:
fixes #296 
Once available, Retrieves title from first task title which is assigned the same value as the Proposed Plan title and more accessible from the UI. Falls back to previous thread title logic (before proposed plan exists)

Screenshot 2025-07-23 at 12 22 04 PM
Screenshot 2025-07-23 at 3 14 53 PM

Code Review Manual CI/CD (pre script creation)
- no comments
- no logging
- no unused vars/funcs
- cn (not relevant)
- self reviewed 
- expanded issue spec (pre-plan)

`getThreadTitle`: Handles thread → title extraction
`computeThreadTitle`: Handles taskPlan vs fallback logic


apps/web/src/components/tasks/index.tsx
```diff
@@ -328,7 +328,9 @@ export function TasksSidebar({
                         <>
                           <div className="mb-1 flex items-start justify-between gap-2">
                             <p className="text-sm leading-relaxed text-gray-900 dark:text-gray-100">
-                              <BasicMarkdownText>{item.plan}</BasicMarkdownText>
+                              <InlineMarkdownText>
+                                {item.plan}
+                              </InlineMarkdownText>
```
starmorph: Why: Fixed hydration error where `<blockquote>` was rendered inside `<p>` tags
Cause: New task titles can contain markdown with block elements
Impact: Without this, you'd get HTML validation errors

apps/web/src/hooks/useThreadMetadata.ts
```diff
@@ -18,21 +18,24 @@ export function useThreadMetadata(thread: Thread<ManagerGraphState>): {
...
-      title: getThreadTitle(thread),
+      title: computeThreadTitle(realTimeTaskPlan, getThreadTitle(thread)),
       lastActivity: calculateLastActivity(thread.updated_at),
-      taskCount: values?.taskPlan?.tasks.length ?? 0,
+      taskCount: realTimeTaskPlan?.tasks.length ?? 0,
       repository: values?.targetRepository
         ? `${values.targetRepository.owner}/${values.targetRepository.repo}`
         : "",
       branch: values?.targetRepository?.branch || "main",
-      taskPlan: values?.taskPlan,
```
starmorph, line 35: replaced manager taskPlan vs realtime multi graph task plan

apps/web/src/components/tasks/index.tsx
bracesproul: lets just drop the `p` tag, and pass the `className` to `InlineMarkdownText` instead of wrapping `InlineMarkdownText` inside a `p` tag

bracesproul: and could we revert back to `BasicMarkdownText`? If the plan has bullet points or code blocks id like to continue rendering them. same w/ the change below

apps/web/src/hooks/useThreadMetadata.ts
```diff
@@ -18,21 +18,24 @@ export function useThreadMetadata(thread: Thread<ManagerGraphState>): {
     status,
     isLoading: isStatusLoading,
     error: statusError,
-  } = useThreadStatus(thread.thread_id);
+    taskPlan: realTimeTaskPlan,
+  } = useThreadStatus(thread.thread_id, {
+    useTaskPlanConfig: true,
```
bracesproul: this is going to start refreshing much quicker for the threads we render on the homepage? Could we avoid this and just get the task plan how we were getting it before?

apps/web/src/lib/thread.ts
```diff
@@ -1,8 +1,21 @@
...
-import { GraphState } from "@open-swe/shared/open-swe/types";
+import { GraphState, TaskPlan } from "@open-swe/shared/open-swe/types";
 import { getActivePlanItems } from "@open-swe/shared/open-swe/tasks";
 
+export function computeThreadTitle(
+  taskPlan: TaskPlan | undefined,
+  fallbackTitle: string,
+): string {
+  if (taskPlan?.tasks && taskPlan.tasks.length > 0) {
+    const firstTaskTitle = taskPlan.tasks[0]?.title;
```
bracesproul, line 11: just need to wrap in a try/catch since `getActiveTask` will error if no tasks exist
```suggestion
    const firstTaskTitle = getActiveTask(taskPlan).title;
```

bracesproul (changes requested):
pretty much lgtm, couple questions/feedback

apps/web/src/hooks/useThreadMetadata.ts
starmorph: 1. yes, fixed. removed taskplanconfig: true to stay at 15 sec poll freq, def ideal.
2. I think we still want to use realtimeTaskplan just without the faster polling bc the previous taskPlan in this call was manager only (didnt require the new realtime taskplan prev)

apps/web/src/lib/thread.ts
starmorph: I tried this however I believe we want the `const firstTaskTitle = taskPlan.tasks[0]?.title;` because getActivetask does not always grab the plan title (plan title is first task only, rather than currently active task)

bracesproul: ah this sounds like a bug in how we set task revisions then. the `title` field is a required field on all `Task` objects, so in theory it should always exist.

separate issue tho, can ignore it for now and use your implementation

bracesproul (approved):
lgtm, ty!

PR #498, opened 2025-07-23
feat: update repository selector

Palashio:
This PR adds the repository fetching & search functionality. It allows users to search over their full list of repositories and select the one that they'd like to modify.

DO NOT MERGE THIS BEFORE THE OTHER CLI PR.

here's a screenshot of what this looks like:


apps/cli/src/TerminalInterface.tsx
```diff
@@ -5,12 +5,14 @@ interface TerminalInterfaceProps {
   message: string | null;
   setMessage: () => void;
   CustomInput: React.FC<{ onSubmit: () => void }>;
+  repoName: string;
```
bracesproul, line 8: would be nice to show `owner/repo`

apps/cli/src/index.tsx
```diff
@@ -33,13 +33,106 @@ const CustomInput: React.FC<{ onSubmit: () => void }> = ({ onSubmit }) => {
   );
 };
 
+const RepoSearchSelect: React.FC<{
+  repos: any[];
```
bracesproul, line 37: can we add types for this? (or drop, ci says never used)

apps/cli/src/index.tsx
```diff
@@ -33,13 +33,106 @@ const CustomInput: React.FC<{ onSubmit: () => void }> = ({ onSubmit }) => {
...
+}
+
 const App: React.FC = () => {
   const [authPrompt, setAuthPrompt] = useState<null | boolean>(null);
   const [authInput, setAuthInput] = useState("");
   const [message, setMessage] = useState<string | null>(null);
   const [exit, setExit] = useState(false);
   const [authStarted, setAuthStarted] = useState(false);
   const [isLoggedIn, setIsLoggedIn] = useState(false);
+  const [repos, setRepos] = useState<any[]>([]);
```
bracesproul, line 133: same here, need type

the shared package has the `@octokit/types` pkg installed. you could add a new file at like `src/github/types.ts` and export the repo type, then import it here

bracesproul: same with all `any`s below

PR #500, opened 2025-07-23
Feat: Document TOC Generation and Search Tool for Large Documentation

aliyanishfaq:
Added documentation processing capabilities including table of contents generation and targeted document search. When documents exceed size limits, the system generates a structured TOC and provides a search tool for detailed exploration.

### `packages/shared/src/open-swe/tools.ts`
- **Added**: `createSearchDocumentForToolFields()` function for document search tool schema
- **Schema**: Accepts `url` (document URL) and `query` (natural language search query)

### `apps/open-swe/src/tools/index.ts`
- **Added**: Export for `search-document-for.js` module

### `apps/open-swe/src/constants.ts`
- **Added**: `DOCUMENT_SEARCH_PROMPT` with comprehensive anti-hallucination instructions

### `apps/open-swe/src/utils/mcp-output/index.ts`
- **Added**: `handleMcpDocumentationOutput()` function for processing large document outputs
- **TOC Generation**: Uses Gemini to create comprehensive table of contents when content exceeds 40k characters
- **Integration**: Suggests `search_document_for` tool usage with specific URL and query format
- **Fallback**: Graceful degradation to simple truncation if TOC generation fails

### `apps/web/src/components/thread/messages/ai.tsx`
- **Added**: Import and type definitions for `createSearchDocumentForToolFields`
- **Added**: Tool recognition and action mapping for `search_document_for`
- **Integration**: Handles both generating and completed states for the new tool

### `apps/web/src/components/gen-ui/action-step.tsx`
- **Added**: Import and tool schema for `searchDocumentForTool`
- **Added**: `SearchDocumentForActionProps` type definition
- **UI Components**: 
  - FileText icon with "Search Document" tooltip
  - Header display showing search query and document URL
  - Output rendering for extracted document content
- **Status**: "Searching document..." while processing, "Document search completed/failed" when done

### `apps/open-swe/src/utils/mcp-output/prompt.ts`
- **TOC Prompt**: Detailed instructions for maximum breadth coverage of document content

### `apps/open-swe/src/tools/search-document-for.ts`
- Fetches document content via FireCrawl and processes with Gemini 

## Integration Flow

1. **Large Document Detection**: When MCP outputs exceed 40k characters, triggers TOC generation
2. **TOC Generation**: Gemini creates comprehensive table of contents using `Task.TOC_GENERATION` 
3. **Search Suggestion**: TOC output includes instructions to use `search_document_for("${url}", "your natural language query")`
4. **Document Search**: Tool fetches document → Gemini extracts relevant content → Returns structured results
5. **UI Display**: Clean interface showing query + URL in header, expandable output with extracted information

This allows exploration of large documentation through a two-step process: overview via TOC, then targeted deep-dive via natural language search queries.

### DIFF
--- a/apps/open-swe/src/constants.ts
+++ b/apps/open-swe/src/constants.ts
@@ -27,3 +27,63 @@ export const LANGGRAPH_USER_PERMISSIONS = [
 export enum RequestSource {
   GITHUB_ISSUE_WEBHOOK = "github_issue_webhook",
 }
+
+export const DOCUMENT_SEARCH_PROMPT = `<identity>
+You are a terminal-based agentic coding assistant built by LangChain. You wrap LLM models to enable natural language interaction with local codebases. You are precise, safe, and helpful.
+</identity>
+
+<role>
+Document Search Agent - Information Extraction Phase
+</role>
+
+<primary_objective>
+Extract ALL information from the provided document that relates to the natural language query. Preserve code snippets, URLs, file paths, and references exactly as they appear in the source document.
+</primary_objective>
+
+<instructions>
+    <core_behavior>
+        - **Extract Only What Exists**: Only extract information that is explicitly present in the document. NEVER add, infer, assume, or generate any information not directly found in the source material.
+        - **Comprehensive Coverage**: Scan the entire document for any content related to the query, including direct mentions and relevant examples or context.
+        - **Exact Preservation**: Copy all code snippets, file paths, URLs, and technical content exactly as written. Maintain original formatting, indentation, and structure.
+        - **No Hallucination**: Do not create, modify, or infer any information. If something is not in the document, do not include it.
+        - **Context Inclusion**: When extracting text, include enough surrounding context to make the information meaningful.
+    </core_behavior>
+
+    <output_format>
+        Your response must use this exact structure:
+
+        <extracted_document_info>
+            <relevant_information>
+            [All prose, explanations, and descriptions from the document that relate to the query. Preserve original wording and include sufficient context.]
+            </relevant_information>
+
+            <code_snippets>
+            [All code blocks and technical examples related to the query. Use markdown code blocks with language tags. Preserve exact formatting.]
+            </code_snippets>
+
+            <links_and_paths>
+            [All URLs, file paths, import statements, and references found. Format as:
+            - URLs: "Display Text: [URL]" or "[URL]"
+            - Paths: "Path: [path/to/file]"
+            - Imports: "Import: [statement]"
+            - Packages: "Package: [name]"]
+            </links_and_paths>
+        </extracted_document_info>
+    </output_format>
+
+    <critical_rules>
+        - Only extract content that actually exists in the provided document
+        - Never add explanations, interpretations, or additional context not present in the source
+        - If no relevant information is found, leave sections empty but still include them
+        - Preserve all technical details exactly as written
+    </critical_rules>
+</instructions>
+
+<natural_language_query>
+{NATURAL_LANGUAGE_QUERY}
+</natural_language_query>
+
+<document_page_content>
+{DOCUMENT_PAGE_CONTENT}
+</document_page_content>
+`;
--- a/apps/open-swe/src/graphs/planner/nodes/generate-message/index.ts
+++ b/apps/open-swe/src/graphs/planner/nodes/generate-message/index.ts
@@ -6,6 +6,7 @@ import {
 import {
   createGetURLContentTool,
   createShellTool,
+  createSearchDocumentForTool,
 } from "../../../../tools/index.js";
 import {
   PlannerGraphState,
@@ -80,6 +81,7 @@ export async function generateAction(
     createShellTool(state),
     createPlannerNotesTool(),
     createGetURLContentTool(),
+    createSearchDocumentForTool(config),
     ...mcpTools,
   ];
   logger.info(
--- a/apps/open-swe/src/graphs/planner/nodes/take-action.ts
+++ b/apps/open-swe/src/graphs/planner/nodes/take-action.ts
@@ -7,6 +7,7 @@ import {
 import {
   createGetURLContentTool,
   createShellTool,
+  createSearchDocumentForTool,
 } from "../../../tools/index.js";
 import { GraphConfig } from "@open-swe/shared/open-swe/types";
 import {
@@ -51,6 +52,7 @@ export async function takeActions(
   const searchTool = createSearchTool(state);
   const plannerNotesTool = createPlannerNotesTool();
   const getURLContentTool = createGetURLContentTool();
+  const searchDocumentForTool = createSearchDocumentForTool(config);
   const mcpTools = await getMcpTools(config);
 
   const mcpToolNames = mcpTools.map((t) => t.name);
@@ -60,6 +62,7 @@ export async function takeActions(
     searchTool,
     plannerNotesTool,
     getURLContentTool,
+    searchDocumentForTool,
     ...mcpTools,
   ];
   const toolsMap = Object.fromEntries(
--- a/apps/open-swe/src/graphs/planner/nodes/take-action.ts
+++ b/apps/open-swe/src/graphs/planner/nodes/take-action.ts
@@ -18,7 +18,7 @@ import {
   safeSchemaToString,
   safeBadArgsError,
 } from "../../../utils/zod-to-string.js";
-import { truncateOutput } from "../../../utils/truncate-outputs.js";
+
 import { createSearchTool } from "../../../tools/search.js";
 import {
   getChangedFilesStatus,
@@ -32,6 +32,7 @@ import { shouldDiagnoseError } from "../../../utils/tool-message-error.js";
 import { Command } from "@langchain/langgraph";
 import { filterHiddenMessages } from "../../../utils/message/filter-hidden.js";
 import { DO_NOT_RENDER_ID_PREFIX } from "@open-swe/shared/constants";
+import { processToolCallContent } from "../../../utils/tool-output-processing.js";
 
 const logger = createLogger(LogLevel.INFO, "TakeAction");
 
@@ -51,6 +52,8 @@ export async function takeActions(
   const plannerNotesTool = createPlannerNotesTool();
   const getURLContentTool = createGetURLContentTool();
   const mcpTools = await getMcpTools(config);
+
+  const mcpToolNames = mcpTools.map((t) => t.name);
 
   const allTools = [
     shellTool,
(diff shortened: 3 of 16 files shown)


packages/shared/src/open-swe/tools.ts
```diff
@@ -391,6 +391,28 @@ export function createGetURLContentToolFields() {
...
+      .string()
+      .describe(
+        "The natural language query to search for within the document content. Be specific about what information you're looking for.",
+      ),
+  });
+
+  return {
+    name: "search_document_for",
+    description:
+      "Search for specific information within a previously fetched document using natural language queries. This tool is particularly useful when working with large documents that have been summarized with a table of contents.",
```
bracesproul: i might add something like "this tool should only be called after attempting to read documentation which was returned as a summarized table of contents"

def improve on that wording, but make it very clear it should first try to read the doc, and only if a summary is returned it should try to call this tool

packages/shared/src/open-swe/tools.ts
```diff
@@ -391,6 +391,28 @@ export function createGetURLContentToolFields() {
...
+  const searchDocumentForSchema = z.object({
+    url: z
+      .string()
+      .describe(
+        "The URL of the document to search within. This should be a URL that was previously fetched and processed.",
+      ),
+    query: z
+      .string()
+      .describe(
+        "The natural language query to search for within the document content. Be specific about what information you're looking for.",
```
bracesproul: would add something here making it clear the `query` will be passed to an LLM which will use it to extract content from the URL

packages/shared/src/open-swe/types.ts
```diff
@@ -369,6 +369,25 @@ export const GraphConfigurationMetadata: {
       description: "Controls randomness (0 = deterministic, 2 = creative)",
     },
   },
+  tocGenerationModelName: {
```
bracesproul: lets just switch the `summarizerModelName` to default to gemini 2.5 pro, and use that instead of adding a new config field

apps/open-swe/src/constants.ts
```diff
@@ -27,3 +27,63 @@ export const LANGGRAPH_USER_PERMISSIONS = [
 export enum RequestSource {
   GITHUB_ISSUE_WEBHOOK = "github_issue_webhook",
 }
+
+export const DOCUMENT_SEARCH_PROMPT = `<identity>
```
bracesproul: instead of this, just create a directory inside `tools` named something like `tools/search-document-for`

then, inside you can have
- `index.ts` for the tool code
- `prompt.ts` for this prompt

apps/open-swe/src/constants.ts
```diff
@@ -27,3 +27,63 @@ export const LANGGRAPH_USER_PERMISSIONS = [
...
+You are a terminal-based agentic coding assistant built by LangChain. You wrap LLM models to enable natural language interaction with local codebases. You are precise, safe, and helpful.
```
bracesproul: this should be updated. it doesnt need to know it's a terminal based coding assistant, or at least say something more targeted to this specific task

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -142,22 +148,19 @@ export async function takeActions(
...
-          })
-        : truncateOutput(result);
+    const content = await processToolCallContent(toolCall, result, config, {
+      mcpToolNames,
+      urlContentToolName: getURLContentTool.name,
+    });
 
     const toolMessage = new ToolMessage({
-      id: uuidv4(),
+      id: `${DO_NOT_RENDER_ID_PREFIX}${uuidv4()}`,
```
bracesproul: why this? I dont think we want to do this

apps/open-swe/src/utils/tool-output-processing.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+): Promise<string> {
+  const { mcpToolNames, urlContentToolName } = options;
+
+  if (mcpToolNames.includes(toolCall.name)) {
+    const url = toolCall.args?.url || toolCall.args?.uri || toolCall.args?.path;
+    return await handleMcpDocumentationOutput(result, config, {
+      url: typeof url === "string" ? url : undefined,
+    });
+  } else if (toolCall.name === urlContentToolName) {
```
bracesproul: i think it makes sense to do the same document summarization with URLs from `urlContentTool` too

in this case, this func should prob just accept a `toolNames: string[]` arg instead of `mcpToolNames` and `urlContentToolName`

apps/open-swe/src/tools/search-document-for.ts
```diff
@@ -0,0 +1,100 @@
...
+      try {
+        parsedUrl = new URL(url);
+      } catch (e) {
+        const errorString = e instanceof Error ? e.message : String(e);
+        logger.error("Failed to parse URL", { url, error: errorString });
+        return {
+          result: `Failed to parse URL: ${url}\nError:\n${errorString}\nPlease ensure the URL provided is properly formatted.`,
+          status: "error",
+        };
+      }
```
bracesproul: lets refactor this into a `formatUrl` or `parseUrl` util. (update the `url-content.ts` file too after refactor

apps/open-swe/src/constants.ts
bracesproul: unless this is used in multiple places? if so, it would be a good time to create a new top level `shared-prompts/` dir, and add a file to it w/ this shared prompt.

bracesproul (changes requested):
mostly lgtm. my one major piece of feedback is around the firecrawl reqs. with this implementation, if the agent wants to ask a question it needs to first try to get the contents via the get url search, or mcp doc tool (this could be 1 firecrawl credit if it uses the get url tool), then to ask a question, we fire off another firecrawl req.

instead, could we persist the full page contents from firecrawl/mcp doc search, and the URL inside the `additional_kwargs` object on `ToolMessage`. This way, if it wants to ask a question about a doc, we can first look through the messages list to see if we've already fetched the contents of this page, and use the string from `additional_kwargs`. otherwise we can fallback to another firecrawl req

packages/shared/src/open-swe/tools.ts
aliyanishfaq: addressed

packages/shared/src/open-swe/types.ts
aliyanishfaq: removed

apps/open-swe/src/utils/tool-output-processing.ts
aliyanishfaq: added

apps/open-swe/src/graphs/planner/nodes/generate-message/index.ts
```diff
@@ -79,7 +80,8 @@ export async function generateAction(
     createSearchTool(state),
     createShellTool(state),
     createPlannerNotesTool(),
-    createGetURLContentTool(),
+    createGetURLContentTool(state),
+    createSearchDocumentForTool(config, state),
```
bracesproul: prefer `state`, `config` just to match the convention of langgraph nodes/runnables

apps/open-swe/src/tools/search-documents-for/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const logger = createLogger(LogLevel.INFO, "SearchDocumentForTool");
+
+type SearchDocumentForInput = z.infer<
+  ReturnType<typeof createSearchDocumentForToolFields>["schema"]
+>;
+
+export function createSearchDocumentForTool(
+  config: GraphConfig,
+  state: Pick<GraphState, "documentCache">,
```
bracesproul: update here

apps/open-swe/src/tools/search-documents-for/prompt.ts
```diff
@@ -0,0 +1,59 @@
+export const DOCUMENT_SEARCH_PROMPT = `<identity>
+You are a specialized document information extraction agent. Your sole purpose is to find and extract relevant information from web documents and documentation based on natural language queries. You are precise, thorough, and never add information not present in the source.
```
bracesproul, line 2: nice, looks much better

apps/open-swe/src/utils/tool-output-processing.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Handles search_document_for, MCP tools, and regular tools with different truncation strategies.
+ */
+export async function processToolCallContent(
+  toolCall: ToolCall,
+  result: string,
+  config: GraphConfig,
+  options: {
+    higherContextLimitToolNames: string[];
+    state: Pick<GraphState, "documentCache">;
+  },
```
bracesproul: nit
```suggestion
  options: {
    higherContextLimitToolNames: string[];
    state: Pick<GraphState, "documentCache">;
    config: GraphConfig,
  },
```

apps/open-swe/src/utils/tool-output-processing.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  if (toolCall.name === "search_document_for") {
+    return truncateOutput(result, {
+      numStartCharacters: 20000,
+      numEndCharacters: 20000,
+    });
+  } else if (higherContextLimitToolNames.includes(toolCall.name)) {
+    const url = toolCall.args?.url || toolCall.args?.uri || toolCall.args?.path;
+    if (url) {
+      state.documentCache[url] = result;
```
bracesproul: does this work? I don't really like this pattern. Could you update this to return a new object which you return to state so we don't have to modify the object like this?

Prob means adding a custom reducer on the state field to handle merging updates with the existing state obj, removing duplicates, etc

packages/shared/src/open-swe/planner/types.ts
```diff
@@ -36,6 +36,16 @@ export const PlannerGraphStateObj = MessagesZodState.extend({
       fn: (_state, update) => update,
     },
   }),
+  /**
+   * Cache of fetched document content keyed by URLs.
+   */
+  documentCache: withLangGraph(z.custom<Record<string, string>>(), {
+    reducer: {
+      schema: z.custom<Record<string, string>>(),
+      fn: (state, update) => ({ ...state, ...update }),
```
bracesproul, line 45: ah you alr have a custom reducer nice.

packages/shared/src/open-swe/types.ts
bracesproul: still present here, plz drop from both places in this file

bracesproul (changes requested):
main concern is how we're updating the doc cache state. after that is fixed lgtm

bracesproul (approved):
ty!!

PR #502, opened 2025-07-23
feat: filter threads by selected Github org [draft]

### LINKED ISSUE #355: Filter threads by installation ID
We should update the threads fetcher to filter by the selected GH installation ID instead of all threads on the user

starmorph:
fixes #355 
Only show threads of the currently selected github org while on homepage or /chat/threads page 
Screenshot 2025-07-23 at 4 33 22 PM
Screenshot 2025-07-23 at 4 33 27 PM
Screenshot 2025-07-23 at 4 33 36 PM
Screenshot 2025-07-23 at 4 33 43 PM

### DIFF
--- a/apps/web/src/app/(v2)/chat/threads/page.tsx
+++ b/apps/web/src/app/(v2)/chat/threads/page.tsx
@@ -11,7 +11,7 @@ import { useThreadsSWR } from "@/hooks/useThreadsSWR";
 import { ThreadCard, ThreadCardLoading } from "@/components/v2/thread-card";
 import { ThemeToggle } from "@/components/theme-toggle";
 import { InstallationSelector } from "@/components/github/installation-selector";
-import { GitHubAppProvider } from "@/providers/GitHubApp";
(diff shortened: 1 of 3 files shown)


apps/web/src/app/(v2)/chat/threads/page.tsx
```diff
@@ -82,7 +84,7 @@ function AllThreadsPageContent() {
   };
 
   return (
-    <GitHubAppProvider>
+    <>
```
starmorph: `<>` to prevent a 100 line indentation cosmetic git diff

bracesproul: ty for this. i removed it after reviewing

bracesproul (changes requested):
this looks good, but it looks like theres a regression. if i enter the threads page for a thread under the langchain org, then in a new tab switch to my personal org and go back to the thread view page and refresh, it fails to load the content. this should not be the case

could you resolve this then re-request a review? ty

bracesproul (approved):
ty!

PR #503, opened 2025-07-23
feat: add updates to cli application

Palashio:
This PR makes significant progress to the CLI application. Only a couple of remaining outstanding tasks for CLI:

- Fix multiple interrupt messages 
- Store messages in state and display on the log side


apps/cli/src/auth-server.ts
```diff
@@ -1,53 +1,44 @@
-import "dotenv/config";
+import dotenv from "dotenv";
+dotenv.config();
```
bracesproul: nit, cleaner before
```suggestion
import "dotenv/config";
```

apps/cli/src/auth-server.ts
```diff
@@ -1,53 +1,44 @@
...
-  scope?: string;
-  error?: string;
-  error_description?: string;
-  error_uri?: string;
-}
+let accessToken: string | null = null;
+let serverStarted = false;
 
-function saveToken(tokenData: GitHubTokenResponse) {
+function saveToken(tokenData: any) {
```
bracesproul, line 28: could we avoid `any` here

bracesproul: any reason we cant use the type that was here before?

apps/cli/src/auth-server.ts
```diff
@@ -57,52 +48,73 @@ const app = express();
 // 1. Start OAuth flow
 app.get("/api/auth/github/login", (_req: Request, res: Response) => {
   const state = Math.random().toString(36).substring(2);
-  const baseGithubAuthUrl = "https://github.com/login/oauth/authorize";
-  const url = new URL(baseGithubAuthUrl);
-  url.searchParams.set("client_id", CLIENT_ID);
-  url.searchParams.set("redirect_uri", CALLBACK_URL);
-  url.searchParams.set("state", state);
-  return res.redirect(url.toString());
+  const githubAuthUrl = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&redirect_uri=${encodeURIComponent(CALLBACK_URL)}&state=${state}`;
```
bracesproul, line 50: same here, looks like changes were reverted. ideally we keep the old pattern of using `new URL`

apps/cli/src/auth-server.ts
```diff
@@ -57,52 +48,73 @@ const app = express();
...
+  res.redirect(githubAuthUrl);
```
bracesproul, line 51: and returning any `res.<something>`
could you update here & in the rest of this file

apps/cli/src/auth-server.ts
bracesproul, line 1: i think this file will be removed when you implement the gh auth-less flow. if so, ignore all my comments

apps/cli/src/utils.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Type guard to check if a value is a valid AgentInboxInterrupt schema.
+ * @param value The value to check.
+ * @returns True if the value matches the schema, false otherwise.
+ */
+export function isAgentInboxInterruptSchema(value: unknown): boolean {
```
bracesproul: could we instead add this to the `shared` package, and delete where it's defined in the `web` app in favor of using the shared export

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+import {
+  startAuthServer,
+  getAccessToken,
+  getInstallationAccessToken,
+  getInstallationId,
+} from "./auth-server.js";
 import open from "open";
-import TerminalInterface from "./TerminalInterface.js";
+import { v4 as uuidv4 } from "uuid";
+import { encryptSecret } from "../../../packages/shared/dist/crypto.js";
```
bracesproul: any import from shared needs to be via the package, and not the path:

```suggestion
import { encryptSecret } from "@open-swe/shared/crypto";
```

bracesproul: plz fix for other imports from shared too

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+      <Text>
+        {text}
+        {dots}
+      </Text>
+    </Box>
+  );
+};
+
+// eslint-disable-next-line no-unused-vars
+const CustomInput: React.FC<{ onSubmit: (value: string) => void }> = ({
```
bracesproul, line 47: if vars are unused, could we just delete them?

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+        </Box>
+      )}
+      <Box marginTop={1}>
+        <Text dimColor>Use ↑/↓ to navigate, Enter to select</Text>
+      </Box>
     </Box>
   );
 };
 
+function isAgentInboxInterruptSchema(value: unknown): boolean {
```
bracesproul: duplicated

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+  const [loadingRepos, setLoadingRepos] = useState(false);
+  const [loadingLogs, setLoadingLogs] = useState(false);
+
+  const sendInterruptMessage = useCallback(
+    async (message: string) => {
+      if (!client || !threadId || !selectedRepo) {
+        return;
+      }
+
+      setLogs((prev) => [...prev, `📤 Interrupt: "${message}"`]);
```
bracesproul: nit
```suggestion
      setLogs((prev) => [...prev, `📤 Interrupt Response: "${message}"`]);
```

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+            branch: selectedRepo.default_branch || "main",
+          },
+        };
+        await client.runs.create(threadId, MANAGER_GRAPH_ID, {
+          input: interruptInput,
+          config: { recursion_limit: 400 },
+          ifNotExists: "create",
+          streamResumable: true,
+          multitaskStrategy: "enqueue",
+          streamMode: ["values", "messages"],
```
bracesproul: the `streamMode` arr should prob be exported as a constant from `shared` so that we can reuse here, and everywhere else in the app to ensure consistent stream modes
```suggestion
          streamMode: ["values", "updates", "messages"],
```

apps/cli/src/index.tsx
```diff
@@ -1,45 +1,253 @@
...
+          targetRepository: {
+            owner,
+            repo: repoName,
+            branch: selectedRepo.default_branch || "main",
+          },
+        };
+        await client.runs.create(threadId, MANAGER_GRAPH_ID, {
+          input: interruptInput,
+          config: { recursion_limit: 400 },
+          ifNotExists: "create",
```
bracesproul, line 221: this applies to the `threadId`. if it doesn't exist, it'll create a new thread which uses that ID. Not sure that's necessary here

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+          setIsLoggedIn(true);
+          setPollingForToken(false);
+        }
+      }, 1000);
+      return () => clearInterval(interval);
+    }
+  }, [pollingForToken, isLoggedIn]);
+
+  // Custom input for planner feedback (must be inside App)
+  const PlannerFeedbackInput: React.FC = () => {
```
bracesproul: why does this need to be inside `App`?

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
-          CustomInput={CustomInput}
-        />
-        <Box flexGrow={1} />
-        <Box width="100%" justifyContent="flex-end">
-          <Text color="gray" dimColor>
-            logged in
-          </Text>
-        </Box>
+      <Box>
+        <Text>Plan feedback (approve/deny): {input}</Text>
```
bracesproul: could we make this a select instead?

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+    );
+  };
+
+  // Streaming logic: when prompt is set, stream logs
+  useEffect(() => {
+    if (prompt && selectedRepo && streamingPhase === "streaming") {
+      setLogs([]);
+      setPlannerFeedback(null);
+      setLoadingLogs(true);
+      (async () => {
```
bracesproul: could we abstract this func into a separate file to make it easier to debug?

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
-          </Text>
-        </Box>
+      <Box>
+        <Text>Plan feedback (approve/deny): {input}</Text>
+      </Box>
+    );
+  };
+
+  // Streaming logic: when prompt is set, stream logs
+  useEffect(() => {
```
bracesproul: not the biggest fan of putting this in a `useEffect`

could we move it to be more event driven? e.g. a user submits, that calls a func which starts streaming...

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+            },
+          });
+          setClient(newClient);
+          const thread = await newClient.threads.create();
+          const threadId = thread.thread_id;
+          setThreadId(threadId);
+          const run = await newClient.runs.create(threadId, MANAGER_GRAPH_ID, {
+            input: runInput,
+            config: { recursion_limit: 400 },
+            ifNotExists: "create",
```
bracesproul: if doing `ifNotExists: create`, just create a new UUID and send that
```suggestion
          const threadId = uuidv4();
          setThreadId(threadId);
          const run = await newClient.runs.create(threadId, MANAGER_GRAPH_ID, {
            input: runInput,
            config: { recursion_limit: 400 },
            ifNotExists: "create",
```

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+            if (
+              !plannerStreamed &&
+              chunk.data &&
+              chunk.data.plannerSession &&
+              typeof chunk.data.plannerSession.threadId === "string" &&
+              typeof chunk.data.plannerSession.runId === "string"
+            ) {
+              plannerStreamed = true;
+              setPlannerThreadId(chunk.data.plannerSession.threadId);
+              for await (const subChunk of newClient.runs.joinStream(
```
bracesproul: dont really like to have this nested inside a use effect, inside another `for await ...`
could you refactor out so it gets called independently from its parent `for await ...`

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+                // Check for programmer session
+                if (
+                  !programmerStreamed &&
+                  subChunk.data?.programmerSession?.threadId &&
+                  typeof subChunk.data.programmerSession.threadId ===
+                    "string" &&
+                  typeof subChunk.data.programmerSession.runId === "string"
+                ) {
+                  programmerStreamed = true;
+                  for await (const programmerChunk of newClient.runs.joinStream(
```
bracesproul: same w/ this

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+  useEffect(() => {
+    if (
+      streamingPhase === "awaitingFeedback" &&
+      plannerFeedback &&
+      plannerThreadId
+    ) {
+      // Immediately switch to streaming mode to hide the feedback prompt
+      setStreamingPhase("streaming");
+
+      const submitFeedback = async () => {
```
bracesproul: could this also get refactored out of this file

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+            installationAccessToken,
+            encryptionKey,
+          );
+          const [owner] = selectedRepo?.full_name.split("/") || [];
+
+          const installationId = getInstallationId();
+          const client = new Client({
+            apiUrl: LANGGRAPH_URL,
+            defaultHeaders: {
+              "x-github-access-token": encryptedUserToken,
```
bracesproul: use var names instead of strings

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+            autoAcceptPlan: false,
+          };
+          const installationId = getInstallationId();
+          const newClient = new Client({
+            apiUrl: LANGGRAPH_URL,
+            defaultHeaders: {
+              GITHUB_TOKEN_COOKIE: encryptedUserToken,
+              GITHUB_INSTALLATION_TOKEN_COOKIE: encryptedInstallationToken,
+              GITHUB_INSTALLATION_NAME: owner,
+              GITHUB_INSTALLATION_ID: installationId,
```
bracesproul: needs brackets or it'll just get interpreted as a string instead of the value of the var
```suggestion
              [GITHUB_TOKEN_COOKIE]: encryptedUserToken,
              [GITHUB_INSTALLATION_TOKEN_COOKIE]: encryptedInstallationToken,
              [GITHUB_INSTALLATION_NAME]: owner,
              [GITHUB_INSTALLATION_ID]: installationId,
```

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+
+          // Create a new stream with the feedback
+          const stream = await client.runs.stream(
+            plannerThreadId,
+            PLANNER_GRAPH_ID,
+            {
+              command: {
+                resume: [
+                  {
+                    type: plannerFeedback === "approve" ? "accept" : "ignore",
```
bracesproul: why only `accept` or `ignore` here? are we not allowing `edit` rn? (thats fine if so, `edit`, can be a p1)

apps/cli/src/index.tsx
```diff
@@ -95,38 +336,538 @@ const App: React.FC = () => {
...
+                  },
+                ],
+              },
+              streamMode: ["updates", "messages"],
+            },
+          );
+
+          let programmerStreamed = false;
+          // Process the stream response
+          for await (const chunk of stream) {
```
bracesproul: same w/ this. plz refactor out of this file

apps/cli/src/index.tsx
Palashio: i was getting some weird type errors here because onSubmit is passed down. tried wrangling with it for ~20 min and figured it probably wasn't worth it for now. if this code remains the same way after local dev is complete i can make a note to come back and remove these.

apps/cli/src/index.tsx
Palashio: oh yes, i think this slipped through the cracks during implementation, i'l add that as a p1.

apps/cli/src/index.tsx
Palashio: this is the main streaming loop so i feel like it should stay in the file?

apps/cli/src/index.tsx
Palashio: doesn't remove the comment; i just kept it in the file because it's a smaller component so thought it'd be easier.

apps/cli/src/index.tsx
Palashio: i don't think the select type exists as a default in ink so i kept it as a text: https://github.com/vadimdemedes/ink

apps/cli/src/index.tsx
Palashio: i removed if not exists: create!

apps/cli/src/index.tsx
Palashio: i removed the `ifNotExists`: "create" functionality.

apps/cli/src/index.tsx
bracesproul: you're doing some sort of select for showing the GH repos. could we use that same component here?

apps/cli/src/index.tsx
bracesproul: bump, why does this need to be in a useEffect?

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,274 @@
...
+  type: string;
+  args?: string;
+  name?: string;
+  index?: number;
+  id?: string;
+}
+/**
+ * Format a tool result based on its type and content
+ */
+function formatToolResult(message: MessageData): string {
```
bracesproul: instead of this type you can convert the message object returned from the stream into an actual langchain message via `coerceMessageLikeToMessage`, then update this func to accept `ToolMessage`

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,274 @@
...
+  const content =
+    typeof message.content === "string"
+      ? message.content
+      : message.content?.[0]?.text || message.content?.[0]?.input || "";
```
bracesproul: after converting to a LC message you can do this:
```suggestion
  const content = getMessageContentString(message.content);
```
import `getMessageContentString` from shared/messages

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,274 @@
...
+              argsString.length > maxLength
+                ? argsString.slice(0, maxLength) + "... [trunc]"
+                : argsString;
+            const toolName = tool.name || "unknown";
+            logs.push(`[TOOL CALL] ${toolName}: ${truncatedArgs}`);
+          });
+        }
+
+        // Handle technical notes
+        if (msg.additional_kwargs?.notes) {
```
bracesproul: instead of this, extract the technical notes from the tool call

`createWriteTechnicalNotesToolFields()` func to get the name, then once you've confirmed the message contains a tool call w/ that name, you can get the notes from the tool call args: `tool_calls[0].args.notes`

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,274 @@
...
+          logs.push(
+            "[TECHNICAL NOTES]",
+            ...msg.additional_kwargs.notes.map((note: string) => `  • ${note}`),
+          );
+        }
+
+        // Handle regular AI messages
+        const text =
```
bracesproul: use the same `getMessageContentString` here from above. and replace anywhere else we're getting the message contents as a string

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,274 @@
...
+    nestedDataObj &&
+    typeof nestedDataObj === "object" &&
+    "messages" in nestedDataObj
+  ) {
+    const messages = Array.isArray(nestedDataObj.messages)
+      ? nestedDataObj.messages
+      : [nestedDataObj.messages];
+    for (const msg of messages) {
+      // Handle tool results
+      if (msg.type === "tool") {
```
bracesproul: once you convert the messages to langchain messages, you should use the type narrowing utils from langchain core. e.g.:

```suggestion
      if (isToolMessage(msg)) {
```

apps/cli/src/streaming.ts
```diff
@@ -0,0 +1,239 @@
...
+  getAccessToken,
+  getInstallationAccessToken,
+  getInstallationId,
+} from "./auth-server.js";
+import { formatDisplayLog } from "./logger.js";
+import { isAgentInboxInterruptSchema } from "@open-swe/shared/agent-inbox-interrupt";
+
+const LANGGRAPH_URL = process.env.LANGGRAPH_URL || "http://localhost:2024";
+
+type StreamMode = "values" | "updates" | "messages";
```
bracesproul: can you import this type from `@langchain/langgraph-sdk` or `@langchain/langgraph` instead?

apps/web/src/components/v2/terminal-input.tsx
```diff
@@ -104,7 +104,7 @@ export function TerminalInput({
             },
             ifNotExists: "create",
             streamResumable: true,
-            streamMode: ["values", "messages-tuple", "custom"],
+            streamMode: ["values", "updates", "messages-tuple", "custom"],
```
bracesproul: prefer importing from shared

packages/shared/src/constants.ts
```diff
@@ -42,3 +42,5 @@ export const DEFAULT_MCP_SERVERS = {
 
 export const API_KEY_REQUIRED_MESSAGE =
   "Unknown users must provide API keys to use the Open SWE demo application";
+
+export const OPEN_SWE_STREAM_MODE = ["values", "updates", "messages"];
```
bracesproul:
```suggestion
export const OPEN_SWE_STREAM_MODE = ["values", "updates", "messages-tuple", "messages", "custom"];
```

apps/web/src/components/v2/terminal-input.tsx
bracesproul: & update other files where we pass `streamMode` throughout the repo

bracesproul (changes requested):

apps/cli/src/logger.ts
Palashio: this makes sense, thanks for pointing me in the right direction with the function!

apps/cli/src/logger.ts
```diff
@@ -0,0 +1,293 @@
...
+  ) {
+    const messages = Array.isArray(nestedDataObj.messages)
+      ? nestedDataObj.messages
+      : [nestedDataObj.messages];
+    for (const msg of messages) {
+      try {
+        const message = coerceMessageLikeToMessage(msg);
+
+        // Handle tool messages
+        if (message instanceof ToolMessage) {
```
bracesproul: prefer using these over `instanceof`. LangChain Core exports one for each message type
```suggestion
        if (isToolMessage(message)) {
```

PR #573, opened 2025-07-28
feat: add file validation to prevent committing unwanted files

aliyanishfaq:
## **Add file validation to prevent committing unwanted files**

Modified `apps/open-swe/src/utils/github/git.ts` to add file filtering before git operations to exclude:
- Dependencies (`node_modules`, `langgraph_api`) 
- Build artifacts (`dist`, `build`, `.next`)
- Environment files (`.env*`)
- Log files (`*.log`)
- OS files (`.DS_Store`, `Thumbs.db`)

**Changes:**
- `getValidFilesToCommit()` - filters files using `git status --porcelain`
- `shouldExcludeFile()` - pattern matching for exclusions


apps/open-swe/src/utils/github/git.ts
```diff
@@ -20,6 +20,109 @@ import { addTaskPlanToIssue } from "./issue-task.js";
...
+
+  return validFiles;
+}
+
+/**
+ * Escapes regex metacharacters in a string to treat them as literal characters.
+ */
+function escapeRegExp(string: string): string {
+  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
```
bracesproul: refactor to other util file, doesnt belong here

apps/open-swe/src/utils/github/git.ts
```diff
@@ -20,6 +20,109 @@ import { addTaskPlanToIssue } from "./issue-task.js";
...
+
+/**
+ * Checks if a file should be excluded from commits based on patterns.
+ */
+function shouldExcludeFile(
```
bracesproul: the output of `git status --porcelain` won't just include the file path. e.g. when a file is deleted, it outputs this:
```
 D apps/web/package.json
```
or when a new file is created, it outputs:
```
?? apps/web/test.txt
```

will this func work even with those prefixes?

bracesproul: would be nice to have a test file for this function*

make random changes in a repo (deletes, file location moves, edits, etc etc), run the git status command manually, capture the outputs and use them in the tests

apps/open-swe/src/utils/github/git.ts
```diff
@@ -20,6 +20,109 @@ import { addTaskPlanToIssue } from "./issue-task.js";
 
 const logger = createLogger(LogLevel.INFO, "GitHub-Git");
 
+// Default patterns for files/directories that should not be committed
+const DEFAULT_EXCLUDED_PATTERNS = [
```
bracesproul: can we also move this to a `constants` file? maybe something like `apps/open-swe/src/utils/github/constants.ts`?

bracesproul (changes requested):
some questions, very close

apps/open-swe/src/__tests__/git-file-validation.test.ts
```diff
@@ -0,0 +1,392 @@
+import { describe, it, expect } from "@jest/globals";
+
+function shouldExcludeFile(
```
bracesproul: import instead of redefining

bracesproul: same w/ rest of non-test funcs/vars in this file

apps/open-swe/src/__tests__/git-file-validation.test.ts
```diff
@@ -0,0 +1,346 @@
+import { describe, it, expect } from "@jest/globals";
+import { shouldExcludeFile } from "../utils/github/git.js";
+import { DEFAULT_EXCLUDED_PATTERNS } from "../utils/github/constants.js";
+
+function parseGitStatus(gitStatusOutput: string): string[] {
```
bracesproul: if youre defining this anyways, lets just add to the `git.ts` file, use it there, and import it here

bracesproul (approved):

apps/open-swe/src/__tests__/git-file-validation.test.ts
aliyanishfaq: done

PR #602, opened 2025-07-30
feat: add planner and manager changes

Palashio:
This PR makes the first set of changes to allow for local development in the CLI. It modifies the tools to accept a local mode header, and then skips a bunch of the github tests in the case that we're in local mode. 

A follow up PR with the programmer changes will follow this.


apps/cli/src/index.tsx
```diff
@@ -12,17 +12,49 @@ import {
...
-
 const GITHUB_LOGIN_URL =
   process.env.GITHUB_LOGIN_URL || "http://localhost:3000/api/auth/github/login";
 
 startAuthServer();
 
+const isLocalMode = process.argv.includes("--local");
+
+// Show usage help
+if (process.argv.includes("--help") || process.argv.includes("-h")) {
```
bracesproul: not a huge deal, but id prefer using something like `commander` for this

bracesproul: also, eventually we'll want to only allow local mode via the cli. is there still a need to support both atm?

apps/cli/src/index.tsx
```diff
@@ -390,8 +455,8 @@ const App: React.FC = () => {
       plannerFeedback &&
       plannerThreadId
     ) {
-      // Immediately switch to streaming mode to hide the feedback prompt
-      setStreamingPhase("streaming");
+      // Don't immediately switch to streaming mode - let the feedback submission complete naturally
+      // setStreamingPhase("streaming"); // REMOVED: This was causing the input to disappear
```
bracesproul: if commented out, should we j remove?

apps/cli/src/index.tsx
```diff
@@ -400,11 +465,22 @@ const App: React.FC = () => {
           selectedRepo,
           setLogs,
           setPlannerFeedback: () => setPlannerFeedback(null),
+          setStreamingPhase,
         });
       })();
     }
   }, [streamingPhase, plannerFeedback, selectedRepo, plannerThreadId]);
 
+  // Add debug logging for streaming phase changes
```
bracesproul: drop these two

apps/cli/src/streaming.ts
```diff
@@ -165,49 +166,83 @@ export class StreamingService {
...
+          messages: [
+            {
+              id: uuidv4(),
+              type: "human",
+              content: [{ type: "text", text: prompt }],
+            },
+          ],
+          targetRepository: {
+            owner: "local",
+            repo: "local",
```
bracesproul, line 190: if you pass in the name of the root directory here, we'll be able to use it inside the tools when we tell the LLM what the base path is when it needs to generate paths

apps/cli/src/streaming.ts
```diff
@@ -165,49 +166,83 @@ export class StreamingService {
...
-
-      const installationId = getInstallationId();
-      const newClient = new Client({
-        apiUrl: LANGGRAPH_URL,
-        defaultHeaders: {
+          autoAcceptPlan: false,
+        };
+
+        headers = {
+          "x-local-mode": "true",
```
bracesproul: could you make this key an exported var from shared, so we can use it here & in open swe to ensure we're always using the correct key name

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -38,11 +38,15 @@ import {
   PLANNER_GRAPH_ID,
 } from "@open-swe/shared/constants";
 import { createLogger, LogLevel } from "../../../../utils/logger.js";
-import { PlannerGraphState } from "@open-swe/shared/open-swe/planner/types";
 import { createClassificationPromptAndToolSchema } from "./utils.js";
 import { RequestSource } from "../../../../constants.js";
 import { StreamMode } from "@langchain/langgraph-sdk";
 
+// Add local mode utility function
```
bracesproul: nit
```suggestion
```

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -38,11 +38,15 @@ import {
...
+function isLocalMode(config: GraphConfig): boolean {
+  return (config.configurable as any)?.["x-local-mode"] === "true";
```
bracesproul: if you add the `x-local-mode` key to the graph config type then you won't need to cast as any

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -60,21 +64,26 @@ export async function classifyMessage(
     throw new Error("No human message found.");
   }
 
-  const langGraphClient = createLangGraphClient({
-    defaultHeaders: getDefaultHeaders(config),
-  });
+  // In local mode, skip LangGraph client creation since we don't need external API calls
+  let plannerThread: any = undefined;
+  let programmerThread: any = undefined;
+  let langGraphClient: any = undefined;
```
bracesproul: can you add types for this? if it's the entire thread you can import the `Thread` type from langgraph/langgraph sdk, and pass the state type as a generic

apps/open-swe/src/graphs/planner/nodes/proposed-plan.ts
```diff
@@ -85,9 +86,16 @@ async function startProgrammerRun(input: {
...
+  let langGraphClient;
+  if (isLocalMode(config)) {
+    langGraphClient = createLangGraphClient({
+      defaultHeaders: { "x-local-mode": "true" },
+    });
+  } else {
+    langGraphClient = createLangGraphClient({
+      defaultHeaders: getDefaultHeaders(config),
+    });
+  }
```
bracesproul: can make this more concise:
```suggestion
  const langGraphClient = createLangGraphClient({
    defaultHeaders: isLocalMode(config)
      ? { "x-local-mode": "true" }
      : getDefaultHeaders(config),
  });
```

apps/open-swe/src/graphs/planner/nodes/proposed-plan.ts
```diff
@@ -112,7 +120,10 @@ async function startProgrammerRun(input: {
       input: runInput,
       config: {
         recursion_limit: 400,
-        configurable: getCustomConfigurableFields(config),
+        configurable: {
+          ...getCustomConfigurableFields(config),
+          ...(isLocalMode(config) && { "x-local-mode": "true" }),
```
bracesproul: not necessary. if headers are supplied (and contain the `x-` prefix then they'll automatically get included in the configurable object
```suggestion
```

apps/open-swe/src/graphs/planner/nodes/proposed-plan.ts
```diff
@@ -206,22 +230,24 @@ export async function interruptProposedPlan(
     });
   }
 
-  await addProposedPlanToIssue(
-    {
+  if (!isLocalMode(config) && state.githubIssueId) {
```
bracesproul: nit
```suggestion
  if (!isLocalMode(config)) {
```

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -207,7 +213,7 @@ export async function takeActions(
         changedFiles,
       },
     );
-    await stashAndClearChanges(repoPath, sandbox);
+    await stashAndClearChanges(repoPath, sandbox, config);
```
bracesproul: not sure we want to do this if its local mode? I'm imagining a user spins it up on a branch they alr have changes on, then the planner would end up stashing & clearing those changes

apps/open-swe/src/graphs/shared/initialize-sandbox.ts
```diff
@@ -344,3 +360,114 @@ export async function initializeSandbox(
     branchName: newBranchName,
   };
 }
+
+/**
+ * Local mode version of initializeSandbox
+ * Skips sandbox creation and repository cloning, works directly with local filesystem
+ */
+async function initializeSandboxLocal(
```
bracesproul, line 368: no need to emit any custom events, or return messages w/ the custom events. thats only necessary for rendering stuff in the web ui (unless you're using this in the terminal to render events?)

apps/open-swe/src/security/auth.ts
```diff
@@ -112,6 +112,20 @@ export const auth = new Auth()
       };
     }
 
+    // Check for local mode first
+    const localModeHeader = request.headers.get("x-local-mode");
```
bracesproul: we should prob figure out a better way to manage this, since we dont want people adding this header to their req and being able to access threads.

I'd add an additional check here which looks at an env var, if it's set then allow this, otherwise don't. this way we can auto-set it for users when running in the terminal, but won't have it set on our deployed instance

apps/open-swe/src/tools/builtin-tools/text-editor.ts
bracesproul, line 1: might be a little cleaner to instead refactor the `handle<xyz>Command` func to execute both? Pass through `isLocal`, don't pass sandbox. then inside the func either execute the local cmd or get the sandbox and execute the cmd

apps/open-swe/src/tools/builtin-tools/view.ts
```diff
@@ -1,33 +1,77 @@
...
           throw new Error(`Unknown command: ${command}`);
         }
 
-        const result = await handleViewCommand(
-          sandbox,
-          path,
-          workDir,
-          view_range as [number, number] | undefined,
-        );
+        let workDir;
```
bracesproul: always prefer typing these vars:
```suggestion
        let workDir: string;
```

bracesproul: same w/ occurrences below/elsewhere in this PR

apps/open-swe/src/tools/builtin-tools/view.ts
bracesproul, line 1: if you make the abstraction i mentioned above, it should cleanup some of the duplicate code in this file too

apps/open-swe/src/tools/url-content.ts
```diff
@@ -2,13 +2,14 @@ import { tool } from "@langchain/core/tools";
...
 import { FireCrawlLoader } from "@langchain/community/document_loaders/web/firecrawl";
-import { GraphState } from "@open-swe/shared/open-swe/types";
+import { GraphState, GraphConfig } from "@open-swe/shared/open-swe/types";
 import { parseUrl } from "../utils/url-parser.js";
 
 const logger = createLogger(LogLevel.INFO, "GetURLContentTool");
 
 export function createGetURLContentTool(
   state: Pick<GraphState, "documentCache">,
+  _config: GraphConfig,
```
bracesproul: drop if unused

apps/open-swe/src/utils/github/git.ts
```diff
@@ -139,15 +173,34 @@ export async function getChangedFilesStatus(
...
+  config?: GraphConfig,
 ): Promise<ExecuteResponse | false> {
   try {
-    const gitStashOutput = await sandbox.process.executeCommand(
-      "git add -A && git stash && git reset --hard",
-      absoluteRepoDir,
-      undefined,
-      TIMEOUT_SEC,
-    );
+    let gitStashOutput: any; // Use any to handle type incompatibility
```
bracesproul: no any plz

apps/open-swe/src/utils/custom-rules.ts
```diff
@@ -117,8 +121,13 @@ export function parseCustomRulesFromString(
 export async function getCustomRules(
   sandbox: Sandbox,
   rootDir: string,
+  config?: GraphConfig,
```
bracesproul, line 127: why is this optional? we should be able to always pass it in, regardless of whether or not it's running locally.

apps/open-swe/src/utils/custom-rules.ts
```diff
@@ -155,6 +164,48 @@ export async function getCustomRules(
   return undefined;
 }
 
+/**Add a comment on lines R168 to R201Add diff commentMarkdown input: edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
```
bracesproul: whats this? can we drop?

apps/open-swe/src/utils/local-mode.ts
```diff
@@ -0,0 +1,29 @@
+import { GraphConfig } from "@open-swe/shared/open-swe/types";
+import path from "path";
+
+/**
+ * Checks if the current execution context is in local mode
+ * (working on local files instead of sandbox/Daytona)
+ */
+export function isLocalMode(config: GraphConfig): boolean {
+  return (config.configurable as any)?.["x-local-mode"] === "true";
```
bracesproul: add `"x-local-mode"` to the config object type and you wont need to cast as any

apps/open-swe/src/utils/local-mode.ts
```diff
@@ -0,0 +1,29 @@
...
+    process.env.OPEN_SWE_PROJECT_PATH ||
+    path.join(process.env.HOME || "", "Desktop", "test")
+  );
+}
+
+/**
+ * Checks if we're in local mode based on environment variables
+ * (useful for contexts where GraphConfig is not available)
+ */
+export function isLocalModeFromEnv(): boolean {
```
bracesproul: wondering if this would be better than dealing with config? We can ensure it's always set if running locally, and won't need to worry about always passing the config value through when we execute diff funcs

apps/open-swe/src/utils/local-shell-executor.ts
```diff
@@ -0,0 +1,211 @@
...
+    logger.info("LocalShellExecutor created", { workingDirectory });
+  }
+
+  async executeCommand(
+    command: string,
+    workdir?: string,
+    env?: Record<string, string>,
+    timeout: number = 30,
+    localMode: boolean = false,
+  ): Promise<ExecuteResponse> {
```
bracesproul: after 3 positional args, i like to move into named args to make the code easier to read. in this case i'd make everything after `command` a named arg since they're all optional:

```suggestion
  async executeCommand(
    command: string,
    args?: {
      workdir?: string,
      env?: Record<string, string>,
      timeout: number = 30,
      localMode: boolean = false,
    }
  ): Promise<ExecuteResponse> {
```

apps/open-swe/src/utils/local-shell-executor.ts
```diff
@@ -0,0 +1,211 @@
...
+          result: spawnError.message,
+          artifacts: {
+            stdout: "",
+            stderr: spawnError.message,
+          },
+        };
+      }
+    }
+
+    // Non-local mode: try exec first, fallback to spawn
```
bracesproul: is this ever used?

apps/open-swe/src/utils/read-write.ts
```diff
@@ -173,7 +201,116 @@ ${delimiter}`;
...
+          : join(workingDirectory, filePath);
+        await fs.writeFile(fullPath, "", "utf-8");
+        return {
+          success: true,
+          output: "",
+        };
+      } catch (createError: any) {
+        return {
+          success: false,
+          output: `FAILED TO CREATE FILE '${filePath}'. Error: ${createError.message}`,
```
bracesproul: add something that indicates the code automatically tried to create after getting an `ENOENT` err when reading:
```suggestion
          output: `FAILED TO AUTOMATICALLY CREATE FILE '${filePath}' AFTER READING FILE ERRORED WITH CODE: ${error.code}. Error: ${createError.message}`,
```

apps/open-swe/src/utils/sandbox.ts
```diff
@@ -96,6 +97,18 @@ export async function getSandboxWithErrorHandling(
...
+  if (isLocalMode(config)) {
+    const mockSandbox = {
+      id: sandboxSessionId || "local-mock-sandbox",
+      state: "started",
+    } as Sandbox;
+
+    return {
+      sandbox: mockSandbox,
+      codebaseTree: null,
+      dependenciesInstalled: false,
```
bracesproul: make this `null` instead?
```suggestion
      dependenciesInstalled: null,
```

apps/cli/src/index.tsx
Palashio: will add commander

apps/open-swe/src/utils/wrap-script.ts
bracesproul, line 1: have you verified this works the same against the deployed sandbox?

packages/shared/src/open-swe/tools.ts
```diff
@@ -546,8 +546,22 @@ export function createTextEditorToolFields(targetRepository: TargetRepository) {
...
+  config?: GraphConfig,
+) {
+  // Local mode utility function (duplicated here since shared package doesn't have access to local-mode)
+  function isLocalMode(config?: GraphConfig): boolean {
+    return (config?.configurable as any)?.["x-local-mode"] === "true";
+  }
+
+  function getLocalWorkingDirectory(): string {
+    return process.cwd();
+  }
```
bracesproul: move them to the shared pkg so you dont need to duplicate like this

packages/shared/src/git.ts
```diff
@@ -1,9 +1,24 @@
 import { SANDBOX_ROOT_DIR } from "./constants.js";
-import { TargetRepository } from "./open-swe/types.js";
+import { TargetRepository, GraphConfig } from "./open-swe/types.js";
+
+// Local mode utility function (duplicated here since shared package doesn't have access to local-mode)
+function isLocalMode(config?: GraphConfig): boolean {
```
bracesproul: oh wait they're alr here. delete the above code and import here instead

bracesproul: defined 4x we should consolidate

Image

Palashio: this...was an egregious mistake. fixing lol

apps/open-swe/src/utils/local-shell-executor.ts
Palashio: no it shouldn't be, removing.

apps/open-swe/src/utils/wrap-script.ts
Palashio: i ended up reverting this to old since i stopped using it

apps/cli/src/index.tsx
```diff
@@ -12,15 +13,30 @@ import {
...
 const GITHUB_LOGIN_URL =
   process.env.GITHUB_LOGIN_URL || "http://localhost:3000/api/auth/github/login";
 
+// Set up Commander.js
+const program = new Command();
+
+program
+  .name("open-swe")
+  .description("Open SWE CLI")
+  .version("0.0.0")
```
bracesproul: version should be a constant (keep in the `cli` app) so its easier to find & update

apps/cli/src/index.tsx
```diff
@@ -235,6 +251,18 @@ const App: React.FC = () => {
 
   // On mount, check for existing token
   useEffect(() => {
+    if (isLocalMode) {
+      setIsLoggedIn(true);
+      setLoadingRepos(false); // Ensure no loading state in local mode
+      // Set up local mode defaults
+      setSelectedRepo({
+        full_name: process.env.OPEN_SWE_LOCAL_PROJECT_PATH || "",
```
bracesproul, line 260: this value is something users need to hard code rn right? I dont see anywhere where we're setting this in the code

if so, we should add a new gh issue to update it to this logic: (and lmk if you think this isn't the correct flow):
1. have logic to find the nearest `.git` directory
2. present two options to the user when they startup the cli:
    a. set the root to the current directory they started the cli in
    b. set the root to the `.git` directory we found

apps/cli/src/utils.ts
```diff
@@ -79,6 +100,13 @@ export async function submitFeedback({
         ],
       },
       streamMode: OPEN_SWE_STREAM_MODE as StreamMode[],
+      ...(isLocalMode && {
+        config: {
+          configurable: {
+            [LOCAL_MODE_HEADER]: "true",
+          },
+        },
+      }),
```
bracesproul: since you're already passing via headers, this is redundant
```suggestion
```

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -60,21 +62,25 @@ export async function classifyMessage(
     throw new Error("No human message found.");
   }
 
-  const langGraphClient = createLangGraphClient({
-    defaultHeaders: getDefaultHeaders(config),
-  });
+  let plannerThread: Thread<PlannerGraphState> | undefined;
+  let programmerThread: Thread<GraphState> | undefined;
+  let langGraphClient: Client | undefined;
```
bracesproul: `langGraphClient` is only assigned once. just assign here using a `const` instead of `let`

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -60,21 +62,25 @@ export async function classifyMessage(
     throw new Error("No human message found.");
   }
 
-  const langGraphClient = createLangGraphClient({
-    defaultHeaders: getDefaultHeaders(config),
-  });
+  let plannerThread: Thread<PlannerGraphState> | undefined;
+  let programmerThread: Thread<GraphState> | undefined;
```
bracesproul, line 66: only assigning these when you're not in local means followups locally will have slightly incorrect prompts
Since these will always be undefined for local, the prompt will always say `not_started`, even if both have already ran once & finished (in this case they should both be `idle`)

ok to leave like this for now, just create a gh issue to track so we can fix before launch

apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -157,178 +163,202 @@ export async function classifyMessage(
     });
   }
 
-  const { githubAccessToken } = getGitHubTokensFromConfig(config);
-  let githubIssueId = state.githubIssueId;
+  // Skip GitHub token requirements in local mode
+  if (!isLocalMode(config)) {
```
bracesproul: instead of nesting all this, could we do an `if (isLocalMode(config)) { ... }` first, and early return? Then the non-local code won't be nested inside an `if` statement, and will be a little easier to read

apps/open-swe/src/graphs/manager/nodes/start-planner.ts
```diff
@@ -22,20 +24,35 @@ const logger = createLogger(LogLevel.INFO, "StartPlanner");
...
-  });
-
   const plannerThreadId = state.plannerSession?.threadId ?? uuidv4();
   const followupMessage = getRecentUserRequest(state.messages, {
     returnFullMessage: true,
+    config,
   });
+
   try {
+    let langGraphClient;
```
bracesproul: can make this more concise, and assign the headers dynamically inline

apps/open-swe/src/graphs/manager/nodes/start-planner.ts
```diff
@@ -46,14 +63,20 @@ export async function startPlanner(
...
       {
         input: runInput,
         config: {
           recursion_limit: 400,
-          configurable: getCustomConfigurableFields(config),
+          configurable: {
+            ...getCustomConfigurableFields(config),
+            ...(isLocalMode(config) && {
+              [LOCAL_MODE_HEADER]: "true",
+            }),
```
bracesproul: same as above, no need to supply if already passing through headers, and the header is prefixed with `x-`. all those headers will be automatically included in configurable fields

apps/open-swe/src/graphs/planner/nodes/take-action.ts
```diff
@@ -198,7 +202,9 @@ export async function takeActions(
       { documentCache: {} } as { documentCache: Record<string, string> },
     );
 
-  const repoPath = getRepoAbsolutePath(state.targetRepository);
+  const repoPath = isLocalMode(config)
```
bracesproul: need to nest all of this inside a falsy `isLocalMode` check. since we dont care if the codebase has changes during the planning step for local mode, we shouldn't:
1. bother checking
2. update the tool messages to have a note about making changes

apps/open-swe/src/graphs/shared/initialize-sandbox.ts
bracesproul: not sure you saw this comment?

apps/open-swe/src/graphs/shared/initialize-sandbox.ts
```diff
@@ -344,3 +360,114 @@ export async function initializeSandbox(
...
+    data: {
+      status: "pending",
+      sandboxSessionId: null,
+      branch: branchName,
+      repo: repoName,
+    },
+  };
+  emitStepEvent(baseGenerateCodebaseTreeAction, "pending");
+
+  let codebaseTree = undefined;
```
bracesproul, line 449: remember to always add types

apps/open-swe/src/security/auth.ts
```diff
@@ -48,6 +49,20 @@ export const auth = new Auth()
       };
     }
 
+    // Check for local mode first
+    const localModeHeader = request.headers.get(LOCAL_MODE_HEADER);
```
bracesproul, line 53: I think you missed my comment here from before. need to only allow this if the node env is dev/an env var specifies its running in CLI mode so that people cant pass this header in prod and get through

apps/open-swe/src/tools/builtin-tools/text-editor.ts
```diff
@@ -32,58 +36,189 @@ export function createTextEditorTool(
...
         } = input;
 
+        let workDir: string;
+        if (isLocalMode(config)) {
+          // In local mode, use the local working directory
+          workDir = getLocalWorkingDirectory();
+        } else {
+          // In sandbox mode, use the sandbox path
+          workDir = getRepoAbsolutePath(state.targetRepository);
+        }
```
bracesproul: can make this a little more concise if its all inline

apps/open-swe/src/tools/builtin-tools/text-editor.ts
```diff
@@ -32,58 +36,189 @@ export function createTextEditorTool(
...
+        if (isLocalMode(config)) {
+          // Local mode: use LocalShellExecutor for file operations
+          const executor = getLocalShellExecutor(getLocalWorkingDirectory());
+
+          // Convert sandbox path to local path
+          let localPath = path;
+          if (path.startsWith("/home/daytona/project/")) {
+            // Remove the sandbox prefix to get the relative path
+            localPath = path.replace("/home/daytona/project/", "");
+          }
```
bracesproul, line 55: hmm ideally this never happens. if it does, i'd assume its due to two issues:
1. we're passing through the wrong default path to the tools
2. we're passing through the wrong CWD in the prompt, and the LLM generates this instead of the actual path

Don't need to fix now, but should create a GH issue to try and track down the cause so we can remove this code

apps/open-swe/src/tools/builtin-tools/text-editor.ts
```diff
@@ -32,58 +36,189 @@ export function createTextEditorTool(
...
+          const filePath = `${workDir}/${localPath}`;
```
bracesproul: always prefer `path.join` (ik im not always the best at remembering to do this, so feel free to remind me if you see me doing this somewherer/rewrite legacy code that does this)

apps/open-swe/src/tools/builtin-tools/text-editor.ts
bracesproul, line 1: i dont really like how we have two ways to execute local vs sandbox commands. for the sake of speed ill let this get merged, but im going to create a gh issue and see if open swe can refactor it for us: https://github.com/langchain-ai/open-swe/issues/634

apps/open-swe/src/tools/install-dependencies.ts
```diff
@@ -16,25 +21,40 @@ const DEFAULT_ENV = {
...
           workdir,
         });
-        const response = await sandbox.process.executeCommand(
-          command,
-          workdir,
-          DEFAULT_ENV,
-          TIMEOUT_SEC * 2.5, // add a 2.5 min timeout
-        );
+
+        let response;
```
bracesproul: types

apps/open-swe/src/utils/local-shell-executor.ts
```diff
@@ -0,0 +1,157 @@
+import { spawn } from "child_process";
+import { createLogger, LogLevel } from "./logger.js";
+
+const logger = createLogger(LogLevel.INFO, "LocalShellExecutor");
+
+export interface ExecuteResponse {
```
bracesproul: i renamed to `LocalExecuteResponse` to avoid naming conflicts with the type daytona exports

apps/open-swe/src/utils/shell-executor.ts
bracesproul, line 1: looks like you alr built the abstraction but didnt implement it everywhere? can be done in another PR if you confirm this is the case

packages/shared/src/open-swe/local-mode.ts
```diff
@@ -0,0 +1,29 @@
...
+    process.env.OPEN_SWE_PROJECT_PATH ||
+    path.join(process.env.HOME || "", "Desktop", "test")
+  );
+}
+
+/**
+ * Checks if we're in local mode based on environment variables
+ * (useful for contexts where GraphConfig is not available)
+ */
+export function isLocalModeFromEnv(): boolean {
```
bracesproul, line 27: this is never used. in a diff pr we should switch everything to use this instead of config, and drop the is local header. issue: https://github.com/langchain-ai/open-swe/issues/635

packages/shared/src/git.ts
```diff
@@ -1,9 +1,19 @@
 import { SANDBOX_ROOT_DIR } from "./constants.js";
-import { TargetRepository } from "./open-swe/types.js";
+import { TargetRepository, GraphConfig } from "./open-swe/types.js";
+import {
+  isLocalMode,
+  getLocalWorkingDirectory,
+} from "./open-swe/local-mode.js";
 
 export function getRepoAbsolutePath(
```
bracesproul, line 8: you're already doing a lot of checks in the code to check if its local mode, and using `getLocalWorkingDirectory`, else falling back to this method. since you alr have a check here, we should remove those conditionals and just use this method everywhere: https://github.com/langchain-ai/open-swe/issues/636

packages/shared/package.json
```diff
@@ -57,6 +57,11 @@
       "import": "./dist/index.js",
       "require": "./dist/index.js"
     },
+    "./open-swe/*": {
+      "types": "./dist/open-swe/*.d.ts",
+      "import": "./dist/open-swe/*.js",
+      "require": "./dist/open-swe/*.js"
+    },
```
bracesproul: shouldn't be necessary since we're exporting everything (see the `./*` below)

apps/cli/src/index.tsx
Palashio: yep, this is something that needs to be hardcoded. i created this issue #637 for it.

apps/open-swe/src/tools/builtin-tools/text-editor.ts
Palashio: Created #31 for it. I'll tackle these set of issues next with evals, but just want to get this bulky PR out.

apps/open-swe/src/tools/builtin-tools/text-editor.ts
Palashio: i created this! i just didn't replace it everywhere for this PR. it'll be changed everywhere in the following PR.

apps/open-swe/src/utils/shell-executor.ts
Palashio: done in next PR.

packages/shared/src/open-swe/local-mode.ts
Palashio: awesome

PR #643, opened 2025-08-01
fix: authentication bugs and example

Palashio:
fixes an auth bug, adds to env example


apps/open-swe/src/graphs/manager/nodes/classify-message/index.ts
```diff
@@ -64,11 +65,16 @@ export async function classifyMessage(
 
   let plannerThread: Thread<PlannerGraphState> | undefined;
   let programmerThread: Thread<GraphState> | undefined;
-  const langGraphClient = createLangGraphClient({
-    defaultHeaders: getDefaultHeaders(config),
-  });
 
+  // Only create LangGraph client if not in local mode
```
bracesproul: ah i see the issue. no default headers to get if its running in local mode

could you just move this below then so its value is assigned inside the `!isLocalMode` check on line 77? no need for two if statements

bracesproul (approved):
lgtm, just one nit comment

PR #664, opened 2025-08-04
feat: bg dev server tool & custom envs

aliyanishfaq:


apps/open-swe/src/graphs/programmer/nodes/generate-message/index.ts
bracesproul, line 1: why expose in planner?

apps/open-swe/src/utils/llms/model-manager.ts
```diff
@@ -73,15 +73,15 @@ const THINKING_BUDGET_TOKENS = 5000;
...
   switch (providerName) {
     case "openai":
-      return apiKeys.openaiApiKey;
+      return apiKeys.openai.api_key;
     case "anthropic":
-      return apiKeys.anthropicApiKey;
+      return apiKeys.anthropic.api_key;
     case "google-genai":
-      return apiKeys.googleApiKey;
+      return apiKeys["google-genai"].api_key;
```
bracesproul: prefer avoiding key names which we need to wrap in strings like this

bracesproul: also prefer camel case over snake case for TS projects (referencing `api_key`)

apps/open-swe/src/utils/sandbox.ts
```diff
@@ -65,9 +111,35 @@ export async function deleteSandbox(
...
+    // Get user environment variables
+    const userEnvVars = config ? getUserEnvironmentVariables(config) : {};
+    const envFingerprint = createEnvFingerprint(userEnvVars);
+
+    const sandboxParams = {
+      ...DEFAULT_SANDBOX_CREATE_PARAMS,
+      labels: {
+        ...DEFAULT_SANDBOX_CREATE_PARAMS.labels,
+
+        OPENSWE_ENV_FINGERPRINT: envFingerprint,
```
bracesproul: why all caps & snake case? would be easy to mistake this for a variable that contains the key name instead of a key name itself

apps/open-swe/src/utils/sandbox.ts
```diff
@@ -115,10 +188,23 @@ export async function getSandboxWithErrorHandling(
...
 
-    // Check sandbox state
+    // Check if environment variables have changed
+    const currentUserEnvs = getUserEnvironmentVariables(config);
+    const currentEnvFingerprint = createEnvFingerprint(currentUserEnvs);
+    const sandboxEnvFingerprint =
+      sandbox.labels?.["OPENSWE_ENV_FINGERPRINT"] || null;
+
+    if (sandboxEnvFingerprint !== currentEnvFingerprint) {
+      logger.info("Environment variables changed, forcing sandbox recreation", {
```
bracesproul, line 199: how could this happen?

wouldn't this only be possible if they changed before invoking a graph, and thus we'd always be setting whatever was passed in the config?

or is this for when users change env vars & send followups?

apps/open-swe/src/utils/user-environment.ts
```diff
@@ -0,0 +1,43 @@
...
+    throw new Error("SECRETS_ENCRYPTION_KEY environment variable is required");
+  }
+
+  const apiKeys = config.configurable?.apiKeys;
+  if (!apiKeys) {
+    return userEnvs;
+  }
+
+  for (const envVarConfig of Object.values(apiKeys)) {
+    if (isEnvVarConfig(envVarConfig) && envVarConfig.allowed_in_dev) {
```
bracesproul: same here. prefer camel case

apps/web/src/lib/parse-env.ts
bracesproul, line 1: is this what langsmith does?

packages/shared/src/open-swe/types.ts
```diff
@@ -617,9 +623,21 @@ export const GraphConfiguration = z.object({
   /**
    * User defined API keys to use
    */
-  apiKeys: withLangGraph(z.record(z.string(), z.string()).optional(), {
-    metadata: GraphConfigurationMetadata.apiKeys,
-  }),
+  apiKeys: withLangGraph(
+    z
+      .record(
```
bracesproul: use z.custom for this so it can reuse the `EnvVarConfig` type

packages/shared/src/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const TIMEOUT_SEC = 60; // 1 minute
 export const SANDBOX_ROOT_DIR = "/home/daytona";
 export const DAYTONA_IMAGE_NAME = "daytonaio/langchain-open-swe:0.1.0";
-export const DAYTONA_SNAPSHOT_NAME = "open-swe-vcpu2-mem4-disk5";
+export const DAYTONA_SNAPSHOT_NAME = "langchain-open-swe-tmux";
```
bracesproul, line 4: is this just an extension of old our snapshot + tmux?

bracesproul (changes requested):

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,39 @@ export function createWriteDefaultTsConfigToolFields(
...
+  targetRepository: TargetRepository,
+) {
+  const repoRoot = getRepoAbsolutePath(targetRepository);
+  const monitorDevServerToolSchema = z.object({
+    command: z
+      .array(z.string())
+      .describe(
+        "The command to start the development server. Examples: ['npm', 'run', 'dev'], ['yarn', 'dev'], ['python', 'app.py'], ['node', 'server.js']. Ensure the command is properly formatted with arguments in the correct order.",
+      ),
+    request: z
```
bracesproul: 1. better name
2. make an object with: `command: string[]` and `workdir?: string` args

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,39 @@ export function createWriteDefaultTsConfigToolFields(
     schema: writeDefaultTsConfigToolSchema,
   };
 }
+export function createMonitorDevServerToolFields(
+  targetRepository: TargetRepository,
+) {
+  const repoRoot = getRepoAbsolutePath(targetRepository);
+  const monitorDevServerToolSchema = z.object({
```
bracesproul: two objects
1. for starting the dev sever
2. running a shell command against it

both accept cmd, workdir
server startup accepts wait time

apps/open-swe/src/graphs/programmer/nodes/generate-message/index.ts
aliyanishfaq: resolved

apps/web/src/lib/parse-env.ts
aliyanishfaq: yes, directly taken from LangGraph code

packages/shared/src/constants.ts
aliyanishfaq: Yes

apps/open-swe/src/utils/llms/model-manager.ts
aliyanishfaq: addressed

apps/open-swe/src/utils/sandbox.ts
aliyanishfaq: addressed

apps/open-swe/src/utils/user-environment.ts
aliyanishfaq: addressed

packages/shared/src/open-swe/tools.ts
aliyanishfaq: renamed to devServer

packages/shared/src/open-swe/tools.ts
aliyanishfaq: addressed

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,49 @@ export function createWriteDefaultTsConfigToolFields(
     schema: writeDefaultTsConfigToolSchema,
   };
 }
+export function createDevServerToolFields(
+  targetRepository: TargetRepository,
+) {
+  const repoRoot = getRepoAbsolutePath(targetRepository);
+  const devServerToolSchema =  z.object({
+    serverConfig: z.object({
```
bracesproul: rename to:
```suggestion
    serverStartupCommand: z.object({
```
so it accurately matches what it contains.

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,49 @@ export function createWriteDefaultTsConfigToolFields(
...
+          "The command to start the development server. Examples: ['langgraph', 'dev'], ['npm', 'run', 'dev'], ['yarn', 'dev'], ['python', 'app.py'], ['node', 'server.js']. Ensure the command is properly formatted with arguments in the correct order.",
+        ),
+      workdir: z
+        .string()
+        .default(repoRoot)
+        .describe(
+          `The working directory where the server command should be executed. Defaults to the root of the repository (${repoRoot}).`,
+        ),
+    }),
+    requestConfig: z.object({
```
bracesproul: same here
```suggestion
    requestCommand: z.object({
```

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,49 @@ export function createWriteDefaultTsConfigToolFields(
...
+export function createDevServerToolFields(
+  targetRepository: TargetRepository,
+) {
+  const repoRoot = getRepoAbsolutePath(targetRepository);
+  const devServerToolSchema =  z.object({
+    serverConfig: z.object({
+      command: z
+        .array(z.string())
+        .describe(
+          "The command to start the development server. Examples: ['langgraph', 'dev'], ['npm', 'run', 'dev'], ['yarn', 'dev'], ['python', 'app.py'], ['node', 'server.js']. Ensure the command is properly formatted with arguments in the correct order.",
```
bracesproul, line 672: i wouldnt include examples here. the prompt should have examples, and prompt it to not just look at the examples, but to search around the codebase to find the actual command. ill push up changes

packages/shared/src/open-swe/tools.ts
```diff
@@ -662,3 +662,49 @@ export function createWriteDefaultTsConfigToolFields(
...
+        .default(repoRoot)
+        .describe(
+          `The working directory where the server command should be executed. Defaults to the root of the repository (${repoRoot}).`,
+        ),
+    }),
+    requestConfig: z.object({
+      testCommand: z
+        .array(z.string())
+        .describe(
+          "The HTTP request command to test the server. Should be a complete command that will be executed after the server starts. Examples: ['curl', '-s', 'http://localhost:3000'], ['curl', '-X', 'POST', '-H', '\"Content-Type: application/json\"', '-d', '{\"test\":true}', 'http://localhost:3000/api/test'], ['curl', '-I', 'http://localhost:8080/health'].",
```
bracesproul: same with this

apps/web/src/app/api/[..._path]/route.ts
```diff
@@ -34,11 +35,14 @@ export const { GET, POST, PUT, PATCH, DELETE, OPTIONS, runtime } =
...
-            encryptedApiKeys[key] = encryptSecret(value, encryptionKey);
+        for (const [envVarId, envVarConfig] of Object.entries(apiKeys)) {
+          if (isEnvVarConfig(envVarConfig)) {
+            encryptedApiKeys[envVarId] = {
+              ...envVarConfig,
+              apiKey: encryptSecret(envVarConfig.apiKey, encryptionKey),
+            };
           } else {
-            encryptedApiKeys[key] = value;
+            encryptedApiKeys[envVarId] = envVarConfig;
```
Comment (bracesproul) [inline apps/web/src/app/api/[..._path]/route.ts:45]: I think should we be checking for the old schema here (`Record<string, string>`) and converting it to the new schema.

E.g. if the key is `anthropicApiKey` and the value is a string, we can convert it to the `EnvVarConfig` schema, then forward it to the server instead of forwarding `{ anthropicApiKey: "api_key_string" }`

apps/open-swe/src/utils/user-environment.ts
```diff
@@ -0,0 +1,43 @@
...
+          envVarConfig.apiKey,
+          secretsEncryptionKey,
+        );
+        if (decryptedKey && envVarConfig.name) {
+          userEnvs[envVarConfig.name] = decryptedKey; // Use the actual env var name
+        }
+      } catch (error) {
+        throw new Error(
+          `Failed to decrypt key for env var ${envVarConfig.name}: ${error}`,
+        );
```
bracesproul, line 37:
```suggestion
        const errorString = error instanceof Error ? error.message : String(error);
        throw new Error(
          `Failed to decrypt key for env var ${envVarConfig.name}: ${errorString}`,
        );
```

Dockerfile
bracesproul, line 1: we should give this a diff name
`LangGraph.Dockerfile`

aliyanishfaq: addressed

bracesproul:

PR #679, opened 2025-08-06
fix: bug with checking out the branch during followups

bracesproul:
- sometimes it throws a fatal err after followup sent where it cant read a property on the messages
- token counts gets reset. need to persist in gh issue after task is completed, then read when initing state
  - same for planner programmer. need to set in gh issue when planner finishes, not pass to programmer, and have programmer read when initiing state


Metjasit999:
เป็นความเมตาที่ไดัรับ เป็นความช่วยเหลือที่ไม่คาดคิด

PR #696, opened 2025-08-06
chore: added docs secrets page

aliyanishfaq:
Added secrets page in documentation


apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+- **Purpose**: Personal API keys and custom development variables
+- **Examples**: `ANTHROPIC_API_KEY`, `MY_DATABASE_URL`, `STRIPE_TEST_KEY`
+- **Access**: User-controlled via settings page in UI
+- **Security**: AES-256-GCM encrypted + explicit user consent required to pass them to the development server
+
+### When API Keys Are Exposed to Sandbox Environments
+
+Your API keys are **never automatically exposed** to sandbox environments. Exposure requires explicit user consent.
+
+You may wonder, why does the sandbox environment even need API keys? We've designed Open-SWE to be your junior dev. Part of development is testing code in real-time. **Open-SWE has the ability to spin up development servers that can monitor your LangGraph agents, Next.js apps, Flask servers, and more.** If you think this option can be useful, please review what it means to pass your keys to the sandbox.
```
bracesproul:
```suggestion
You may wonder, why does the sandbox environment even need API keys? We've designed Open-SWE to be your junior dev. Part of development is testing code in real-time. **Open-SWE has the ability to spin up development servers that can run your LangGraph agents, Next.js apps, Flask servers, and more, all within the sandbox its running in.** If you think this option can be useful, please review what it means to pass your keys to the sandbox.
```

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+<Note>
+  Consider using separate development API keys with limited permissions instead of your production keys when enabling sandbox access. This reduces the impact of potential exposure while maintaining functionality.
+</Note>
+
+**Default Behavior (Secure):**
+- API keys are only used for LLM model initialization & setting up Daytona using your own key, if you passed it
+- Keys are NOT available as environment variables in sandboxes or accessible to LLMs
+- Each key has `allowed_in_dev: false` by default
```
bracesproul:
```suggestion
- Each key has `allowedInDev: false` by default
```

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+**When You Enable "Include in Dev Server":**
+- You manually toggle the switch for each API key
+- Key becomes available as an environment variable in sandbox
+- You maintain control over which keys are exposed
+
+## Sandbox Security & Isolation
+
+### Container Isolation
+
+Each user session runs in a completely isolated Daytona container with multiple security boundaries. These containers modify the code and don't exchange information across sessions.
```
bracesproul:
```suggestion
Each user session runs in a completely isolated Daytona container with multiple security boundaries. These containers can modify the code in your repo on a new branch and don't exchange information across sessions.
```

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+  - Sandboxed processes cannot access host system or code
+
+  **Temporal Isolation:**
+  - Automatic deletion after 15 minutes of inactivity
+  - No persistent storage across sessions
+  - Fresh environment for each new session
+
+  **Network Isolation:**
+  - Containers cannot communicate with each other
+  - Controlled outbound internet access only
```
bracesproul: sandboxes can make arbitrary network requests tho. not sure this is exactly true?

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+</Accordion>
+
+### Credential Lifespan
+
+Your credentials have a very limited lifespan in sandbox environments:
+
+**Auto-Deletion:**
+- Sandboxes automatically delete after 15 minutes of inactivity
+- All environment variables and credentials are destroyed
```
bracesproul: they persist (encrypted) in the thread on the langgraph server indefinitely though

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+- No persistent storage of sensitive data
+
+**Change-Based Recreation:**
+- Sandbox recreation when you update environment variables
+- Old credentials immediately invalidated
+- Fresh environment with updated credentials
+
+**Session Boundaries:**
+- New session = new sandbox = fresh credentials
+- No credential persistence across user sessions
```
bracesproul: if they dont remove from the config in the browser then it'll be set in every new session?

apps/docs/usage/secrets.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+### Development Environment Exposure Risks
+
+You should only use this feature if you find it useful to monitor development servers. We recommend not using your main keys and having access control for keys that you pass to sandbox environments.
+
+While the Open-SWE UI and agent exchange these using encryption, **LLM may leave environment variables exposed in the code.**
+
+**When you delete environment variables:**
+- Key deleted from frontend localStorage
+- Encrypted version removed from database
```
bracesproul: we don't actually do this tho. once it's set on a thread, we never delete it (encrypted version)

apps/docs/usage/secrets.mdx
aliyanishfaq: Yes, but the context was they are not on local network and can't communicate with another sandbox session

apps/docs/usage/secrets.mdx
aliyanishfaq: I see, will remove this part then

n-sviridenko:
@bracesproul was checking the code, but haven't found that logic. Is it something you guys are planning to work on or maybe it's already there elsewhere?

PR #697, opened 2025-08-06
Palash/swe evals

Palashio:


apps/open-swe/package.json
```diff
@@ -39,6 +39,7 @@
     "@octokit/rest": "^22.0.0",
     "@octokit/webhooks": "^14.0.2",
     "@open-swe/shared": "*",
+    "commander": "^14.0.0",
```
bracesproul: dev deps

bracesproul: not used in prod, so dont need in hard deps

apps/open-swe/package.json
```diff
@@ -71,5 +73,13 @@
     "typescript-eslint": "^8.22.0",
     "vitest": "^3.2.3"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "description": "The core LangGraph agent application that powers Open SWE's autonomous code understanding, planning, and execution capabilities.",
+  "main": "eslint.config.js",
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
bracesproul:
```suggestion
  "license": "MIT",
```

apps/open-swe/langbench/static/langgraph_prs.json
bracesproul, line 1: this should get uploaded to a langsmith dataset, and then we can just reference it by its ID, and pull it in via the langsmith SDK

also, how do we know which test to run during the evals? I see there's a diff url, but no way to know which file in the diff is the test file we'll want to run after the code generation is complete

finally, you store the commit sha of the merge commit, but we'll also need the commit from just before this PR went in, so we can init the sandbox with the codebase at that point in time

apps/open-swe/langbench/runEvals.ts
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env node
```
bracesproul: shouldn't be an excutable, instead could you make it a `*.eval.ts` file, and use the `import * as ls from "langsmith/vitest";` module to define tests to run these via

see this file for an example: `apps/open-swe/evals/langgraph.eval.ts`

apps/open-swe/langbench/static/langgraph_prs.json
bracesproul: ah looking elsewhere it looks like youre using the merge commit sha as the point to checkout the repo from. just double checking thatll be the commit directly before this PR was merged

apps/open-swe/langbench/runEvals.ts
```diff
@@ -0,0 +1,217 @@
...
+      repo: prData.repo_name,
+      branch: "main",
+      baseCommit: undefined,
+    };
+    const repoDir = getRepoAbsolutePath(targetRepository);
+
+    // First, clone and checkout the merge commit to get the parent
+    await cloneAndCheckoutRepo(sandbox, prData, prData.merge_commit_sha);
+
+    // Get the pre-merge commit (parent of merge commit)
```
bracesproul: ah youre doing it here. would be better to just store this on the dataset, and when cloning the repo pass that pre-merge commit as the `baseCommit` so it checks out the repo from that commit

dont want to make this req every time when we can just hard code it

apps/open-swe/langbench/static/langgraph_prs.json
Palashio: here's what i was thinking: 

we go to the `diff_url` (just raw html), identify what test files exist in the diff -> and then run that file. going to the `diff_url` and manually looking for test files is how i initially filtered the PRs as well.

Palashio: i'd rather keep it as a json for now and then move it over to a langsmith dataset in the future? it's just easier to work with as i'm iterating -- commenting out different cases, seeing the diffs etc.

apps/open-swe/langbench/runEvals.ts
Palashio: what i'll do is that i'll store the merge commit, get the pre-merge commit, and then use that so we only have to clone once without the checkout. the github info doesn't come with the pre merge commit and so we'd have to fetch and store it manually.

apps/open-swe/langbench/static/langgraph_prs.json
bracesproul: > identify what test files exist in the diff -> and then run that file.

this is fine, but we'll need to ensure it'll still work if the agent makes updates to the same test files. we'll want to run any tests it writes, but also run the tests changed/added in the PR independently from the tests it wrote

apps/open-swe/langbench/runEvals.eval.ts
```diff
@@ -0,0 +1,160 @@
...
+    branch: "main",
+    baseCommit: targetCommit,
+  };
+
+  logger.info(
+    `Cloning repository: ${prData.repo_owner}/${prData.repo_name} at commit ${targetCommit}`,
+  );
+
+  await cloneRepo(sandbox, targetRepository, {
+    githubInstallationToken: process.env.GITHUB_TOKEN || "dummy_token",
```
bracesproul: throw err if no token found in env?

apps/open-swe/src/utils/github/api.ts
bracesproul, line 1: why are we still doing this programmatically instead of hard coding the values? they'll never change, so idk why we can't just hard code them and never make this req again

apps/open-swe/langbench/runEvals.eval.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    logger.info(`Found pre-merge commit: ${preMergeCommit}`);
+    result.pre_merge_sha = preMergeCommit;
+
+    const targetRepository: TargetRepository = {
+      owner: prData.repo_owner,
+      repo: prData.repo_name,
+      branch: "main",
+      baseCommit: prData.merge_commit_sha,
```
bracesproul: if you set this to the `preMergeCommit`, and set `branch` to `undefined`, the `cloneRepo` command will clone the repo & checkout that commit. then you won't even need the `cloneAndCheckoutRepo` func since you can just call `cloneRepo` inline here

bracesproul (changes requested):
looks like you missed my comments about the pre merge commit stuff. once you resolve these comments it lgtm and can merge

bracesproul (approved):
lgtm. i pushed up one commit renaming `GITHUB_TOKEN` to `GITHUB_PAT` to match other usage of pat in the repo

PR #708, opened 2025-08-07
feat: update langbench to run extracted tests

Palashio:
This PR makes updates to langbench

- updates each test cases with the test files that are modified in it 
- updates the evaluation pipeline to setup the enviroment, install dependencies, and run the tests
- the evaluated test outputs are then processed to understand how many test cases passed
- the entire repository is reverted to before the merge commit but the tests are checkout out post merge commit. this is done so that when we kick off the langgraph job the agent has access to the test cases it needs to pass.

### DIFF
--- a/apps/open-swe/langbench/evaluator.eval.ts
+++ b/apps/open-swe/langbench/evaluator.eval.ts
@@ -96,12 +96,12 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
       logger.info(
         `Checking out test files from merge commit: ${prData.merge_commit_sha}`,
       );
-      await checkoutFilesFromCommit(
+      await checkoutFilesFromCommit({
         sandbox,
         repoDir,
-        prData.merge_commit_sha,
-        testFiles,
-      );
+        commitSha: prData.merge_commit_sha,
+        filePaths: testFiles,
+      });
     }
 
     // Run tests on detected test files
--- a/apps/open-swe/langbench/evaluator.eval.ts
+++ b/apps/open-swe/langbench/evaluator.eval.ts
@@ -30,12 +30,12 @@ logger.info(`Starting evals over ${DATASET.length} PRs...`);
  */
 async function processPR(prData: PRData): Promise<PRProcessResult> {
   const result: PRProcessResult = {
-    pr_number: prData.pr_number,
-    repo_name: prData.repo_name,
+    prNumber: prData.prNumber,
+    repoName: prData.repoName,
     success: false,
-    evals_found: false,
-    evals_files: [],
-    test_files: [],
+    evalsFound: false,
+    evalsFiles: [],
+    testFiles: [],
   };
   const daytona = new Daytona({
     organizationId: process.env.DAYTONA_ORGANIZATION_ID,
@@ -43,11 +43,11 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
   let sandbox: Sandbox | undefined;
 
   try {
-    logger.info(`Processing PR #${prData.pr_number}: ${prData.title}`);
+    logger.info(`Processing PR #${prData.prNumber}: ${prData.title}`);
 
     // Use test files from PR data (already fetched and stored)
-    const testFiles = prData.test_files || [];
-    result.test_files = testFiles;
+    const testFiles = prData.testFiles || [];
+    result.testFiles = testFiles;
     // Create sandbox
     sandbox = await daytona.create(DEFAULT_SANDBOX_CREATE_PARAMS);
 
@@ -58,17 +58,17 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
       );
     }
 
-    result.workspace_id = sandbox.id;
+    result.workspaceId = sandbox.id;
     logger.info(`Created sandbox: ${sandbox.id}`);
 
     // Use the hardcoded pre-merge commit SHA from the dataset
-    const preMergeCommit = prData.pre_merge_commit_sha;
+    const preMergeCommit = prData.preMergeCommitSha;
     logger.info(`Using pre-merge commit: ${preMergeCommit}`);
-    result.pre_merge_sha = preMergeCommit;
+    result.preMergeSha = preMergeCommit;
 
     const targetRepository: TargetRepository = {
-      owner: prData.repo_owner,
-      repo: prData.repo_name,
+      owner: prData.repoOwner,
+      repo: prData.repoName,
       branch: undefined,
       baseCommit: preMergeCommit,
     };
@@ -94,12 +94,12 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
     // Checkout test files from the merge commit to get the updated test files
     if (testFiles.length > 0) {
       logger.info(
-        `Checking out test files from merge commit: ${prData.merge_commit_sha}`,
+        `Checking out test files from merge commit: ${prData.mergeCommitSha}`,
       );
       await checkoutFilesFromCommit({
         sandbox,
         repoDir,
-        commitSha: prData.merge_commit_sha,
+        commitSha: prData.mergeCommitSha,
         filePaths: testFiles,
       });
     }
@@ -115,23 +115,23 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
         repoDir,
         timeoutSec: 300,
       });
-      result.test_results = testResults;
+      result.testResults = testResults;
 
-      logger.info(`Test execution completed for PR #${prData.pr_number}`, {
+      logger.info(`Test execution completed for PR #${prData.prNumber}`, {
         totalTests: testResults.totalTests,
         passedTests: testResults.passedTests,
         failedTests: testResults.failedTests,
         success: testResults.success,
       });
     } else {
-      logger.info(`No test files to run for PR #${prData.pr_number}`);
+      logger.info(`No test files to run for PR #${prData.prNumber}`);
     }
 
     result.success = true;
-    logger.info(`Successfully processed PR #${prData.pr_number}`);
+    logger.info(`Successfully processed PR #${prData.prNumber}`);
   } catch (error) {
     result.error = error instanceof Error ? error.message : String(error);
-    logger.error(`Failed to process PR #${prData.pr_number}:`, { error });
+    logger.error(`Failed to process PR #${prData.prNumber}:`, { error });
   } finally {
     // Cleanup sandbox
     if (sandbox) {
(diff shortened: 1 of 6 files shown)


apps/open-swe/langbench/runEvals.eval.ts
```diff
@@ -42,10 +45,29 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
...
     logger.info(`Processing PR #${prData.pr_number}: ${prData.title}`);
 
+    // Use test files from PR data (already fetched and stored)
+    const testFiles = prData.test_files || [];
+    result.test_files = testFiles;
+
+    if (testFiles.length > 0) {
+      logger.info(
+        `Found ${testFiles.length} test files modified in PR #${prData.pr_number}:`,
+      );
```
bracesproul: is this supposed to represent the test files the agent modified? If im understanding correctly, it's just getting this from the hard coded test files list in the dataset?

(asking b/c i want to avoid confusing logs)

apps/open-swe/langbench/runEvals.eval.ts
```diff
@@ -78,6 +100,42 @@ async function processPR(prData: PRData): Promise<PRProcessResult> {
...
+    if (testFiles.length > 0) {
+      logger.info(
+        `Checking out test files from merge commit: ${prData.merge_commit_sha}`,
+      );
+      await checkoutFilesFromCommit(
+        sandbox,
+        repoDir,
+        prData.merge_commit_sha,
+        testFiles,
+      );
```
bracesproul: okay so this will overwrite all those files in the agent's PR with their state in the merge commit sha. makes sense

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+import { createLogger, LogLevel } from "../src/utils/logger.js";
+import { ENV_CONSTANTS } from "../src/utils/env-setup.js";
+
+const logger = createLogger(LogLevel.DEBUG, "Langbench Utils");
+
+/**
+ * Fetch diff content from a diff URL and extract test file names, this function is used in one-off situtations to get the test files from the diff url.
+ *
+ */
+export async function getTestFilesFromDiff(diffUrl: string): Promise<string[]> {
```
bracesproul, line 11: not used anywhere?

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+  command: string;
+  workingDir: string;
+  env?: Record<string, string>;
+  timeoutSec: number;
+}
+
+/**
+ * Run pytest on specific test files and return structured results
+ */
+export const runPytestOnFiles = async (
```
bracesproul: prefer doing `function` unless you're defining a func inside another func

top level: normal `function`
nested: arrow function (`const xyz - () => {}`)

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+  sandbox: Sandbox,
```
bracesproul: also a good rule of thumb for JS funcs:
- 3 or less args, can use positional args
- more than 3, either put all inside a named arg object, or all args after the 3rd go into a named arg object

makes it hard to keep track of if you have more than 3 positional args

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+// Use shared constants from env-setup utility
+const { RUN_PYTHON_IN_VENV, RUN_PIP_IN_VENV } = ENV_CONSTANTS;
+
+// Installation commands for pytest and dependencies
+const PYTEST_INSTALL_COMMANDS = [
+  `${RUN_PIP_IN_VENV} install pytest pytest-mock pytest-asyncio syrupy pytest-json-report`,
+  `${RUN_PIP_IN_VENV} install -e ./libs/langgraph`,
+];
+
+export interface TestResult {
```
bracesproul: duplicated. you alr have `TestResults` which is the same type

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+  success: boolean;
+  error: string | null;
+  totalTests: number;
+  passedTests: number;
+  failedTests: number;
+  testDetails: string[];
+}
+
+export interface ExecOptions {
```
bracesproul: same here

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Run pytest on specific test files and return structured results
+ */
+export const runPytestOnFiles = async (
+  sandbox: Sandbox,
+  testFiles: string[],
+  repoDir: string,
+  timeoutSec: number = 300,
+): Promise<TestResult> => {
+  if (testFiles.length === 0) {
+    logger.info("No test files provided, skipping pytest execution");
```
bracesproul: nit
```suggestion
    logger.warn("No test files provided, skipping pytest execution");
```

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  // Join test files for pytest command
+  const testFilesArg = testFiles.join(" ");
+  const command = `${RUN_PYTHON_IN_VENV} -m pytest ${testFilesArg} -v --tb=short --json-report --json-report-file=/tmp/pytest_report.json`;
+  logger.info("Running pytest command", { command });
+
+  logger.info(
+    "Installing pytest, pytest-mock, pytest-asyncio, syrupy, pytest-json-report, and langgraph in virtual environment...",
+  );
+  const installCommand = PYTEST_INSTALL_COMMANDS.join(" && ");
```
bracesproul: prefer executing two separate commands so we can identify errors easier

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+      command,
+      repoDir,
+      undefined,
+      timeoutSec,
+    );
+
+    // Read the JSON report file
+    let parsed: Omit<TestResult, "success" | "error">;
+    try {
+      const jsonReportResult = await sandbox.process.executeCommand(
```
bracesproul: prefer using `readFile` here if possible to reuse existing functions

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+      testDetails: [],
+    };
+  }
+};
+
+/**
+ * Parse pytest JSON report to extract test results
+ */
+export const parsePytestJsonReport = (
+  jsonReport: any,
```
bracesproul: can we omit `any`?
we should know ahead of time what the structure should look like, so could we type it as that object?

apps/open-swe/src/utils/github/git.ts
```diff
@@ -613,3 +613,44 @@ async function performClone(
...
+  }
+
+  logger.info(
+    `Checking out ${filePaths.length} files from commit ${commitSha}`,
+  );
+
+  for (const filePath of filePaths) {
+    try {
+      const result = await sandbox.process.executeCommand(
+        `git checkout ${commitSha} -- "${filePath}"`,
```
bracesproul: could this result in any errors if there's merge conflicts? not sure.

bracesproul: if there's a way to force it to overwrite the existing files in favor of the ones from the merge commit that would be ideal

apps/open-swe/src/utils/github/git.ts
```diff
@@ -613,3 +613,44 @@ async function performClone(
...
+        repoDir,
+        undefined,
+        30, // 30 second timeout for git checkout
```
bracesproul:
```suggestion
        30,
```

bracesproul (changes requested):
mostly nits

apps/open-swe/langbench/utils.ts
Palashio: i use this when processing json files and creating test cases. so wanted to leave it in here for the future!

apps/open-swe/langbench/runEvals.eval.ts
Palashio: you're right this was used before when i wasn't storing the test files

apps/open-swe/langbench/runEvals.eval.ts
Palashio: yep

Palashio: so we'll have old version of the repo with the new test cases that should pass after the agent is done going to work

apps/open-swe/src/utils/github/git.ts
Palashio: added --force should work now

apps/open-swe/langbench/evaluator.eval.ts
```diff
@@ -28,34 +30,45 @@ logger.info(`Starting evals over ${DATASET.length} PRs...`);
...
+    const testFiles = prData.testFiles || [];
+    result.testFiles = testFiles;
     // Create sandbox
     sandbox = await daytona.create(DEFAULT_SANDBOX_CREATE_PARAMS);
-    result.workspace_id = sandbox.id;
 
+    // Validate sandbox was created properly
+    if (!sandbox || !sandbox.id) {
+      throw new Error(
+        `Failed to create valid sandbox. Sandbox object: ${JSON.stringify(sandbox)}`,
```
bracesproul: i wouldn't log the sandbox. if it's undefined, just throw err
```suggestion
        "Failed to create valid sandbox",
```

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,227 @@
+import { createLogger, LogLevel } from "../src/utils/logger.js";
+import { ENV_CONSTANTS } from "../src/utils/env-setup.js";
+import { TestResults, PytestJsonReport, RunPytestOptions } from "./types.js";
+import { readFile } from "../src/utils/read-write.js";
+
+const logger = createLogger(LogLevel.DEBUG, "Langbench Utils");
+
+/**
+ * Fetch diff content from a diff URL and extract test file names, this function is used in one-off situtations to get the test files from the diff url.
+ *
```
bracesproul:
```suggestion
```

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,227 @@
...
+  } catch (error) {
+    logger.error(`Failed to fetch or parse diff from ${diffUrl}:`, { error });
+    return [];
+  }
+}
+
+/**
+ * Check if a file path represents a test file in libs/langgraph/tests/
+ */
+export function isLangGraphTestFile(filePath: string): boolean {
```
bracesproul: nit, drop export if only used in this file
```suggestion
function isLangGraphTestFile(filePath: string): boolean {
```

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,227 @@
...
+    };
+  }
+
+  logger.info(`Running pytest on ${testFiles.length} test files`, {
+    testFiles,
+  });
+
+  // Join test files for pytest command
+  const testFilesArg = testFiles.join(" ");
+  const command = `${RUN_PYTHON_IN_VENV} -m pytest ${testFilesArg} -v --tb=short --json-report --json-report-file=/tmp/pytest_report.json`;
```
bracesproul, line 83:
```suggestion
  const runTestCommand = `${RUN_PYTHON_IN_VENV} -m pytest ${testFilesArg} -v --tb=short --json-report --json-report-file=/tmp/pytest_report.json`;
```

You have a var also defined as `command` below (`L93`). want to avoid duplicates (separate: we should enable the eslint rule that errors on stuff like this. idk the name but im sure there is one)

apps/open-swe/langbench/utils.ts
```diff
@@ -0,0 +1,227 @@
...
+  logger.info("Running pytest command", { command });
+
+  logger.info(
+    "Installing pytest, pytest-mock, pytest-asyncio, syrupy, pytest-json-report, and langgraph in virtual environment...",
+  );
+
+  for (const [index, command] of PYTEST_INSTALL_COMMANDS.entries()) {
```
bracesproul: Can you give each of the values in this arr its own variable, and execute them one at a time instead of iterating through an array like this?

bracesproul (approved):
couple nits, after resolving you can merge. ty!

PR #724, opened 2025-08-09
feat: Add restart from last checkpoint button

### LINKED ISSUE #710: Feature Request: Restart Threads that have ran into an error.
Feature: Add the ability to start, stop (pause) and restart threads. Especially with threads that have generated an error that has been corrected by the user.

This feature would be greatly appreciated if you could implement it.

bracesproul:
Fixes #710


bracesproul:
@greptileai

PR #745, opened 2025-08-11
fix: adjust JWT issued-at time to handle clock skew in local environment

### LINKED ISSUE #744: JWT authentication fails locally due to clock skew between server and client
When running the application locally, requests to protected endpoints fail with a 401 Unauthorized error. This happens because the JWT validation process considers the token expired, even though it was freshly generated. We are getting the error message “Expiration time' claim ('exp') is too far in the future”.

Image

### Actual Behavior

- Generated tokens are rejected as expired due to a clock skew between token iat/exp claims and the server’s current time.\

### Possible Cause

- Local development environments sometimes have small time discrepancies due to differences in system clocks between client and server.

### Proposed Solution

- Expiration time of the JWT must be no more than 10 minutes into the future.

- Allow a small leeway (e.g., 60 seconds) when validating tokens to tolerate minor clock differences.

### Environment

- Local OS: Windows 11
- Node.js version: v23.6.0
- JWT library: jsonwebtoken ^9.0.2
- Production Environment: Works fine.

57hemanth:
This PR modifies JWT generation for GitHub App authentication to account for potential clock skew between the local machine and GitHub's servers.

### Changes:

- Updated generateJWT to subtract 60 seconds from the current time when setting the iat field.
- Ensures the token is not considered "issued in the future" if local time is slightly ahead.
- Keeps the exp value within GitHub’s required 10-minute maximum window.

### Rationale:

In local development, JWT authentication was failing when the local machine’s clock was slightly ahead of GitHub's. This caused the GitHub API to reject the token as invalid. By subtracting 60 seconds from the iat value, we ensure reliable authentication without compromising security.

This adjustment follows the recommendation in [GitHub’s official documentation on generating JWTs for a GitHub App](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app?apiVersion=2022-11-28&versionId=free-pro-team%40latest&productId=rest), which uses the same offset to prevent clock-skew-related authentication failures. The example code provided in GitHub’s documentation, written in Ruby, demonstrates this approach.

Screenshot 2025-08-11 152541

### References:
Fixes #744

### DIFF
--- a/packages/shared/src/jwt.ts
+++ b/packages/shared/src/jwt.ts
@@ -5,10 +5,12 @@ import jsonwebtoken from "jsonwebtoken";
  */
 export function generateJWT(appId: string, privateKey: string): string {
   const now = Math.floor(Date.now() / 1000);
+  const issuedAtTime = now - 60;
+  const expirationTime = issuedAtTime + 10 * 60;
 
   const payload = {
-    iat: now,
-    exp: now + 10 * 60,
+    iat: issuedAtTime,
+    exp: expirationTime,
     iss: appId,
   };
 


bracesproul:

PR #792, opened 2025-08-22
feat: modify cli for deep agents + formatting 

Palashio:


apps/cli/src/logger.ts
```diff
@@ -31,22 +68,71 @@ function formatToolCallArgs(tool: ToolCall): string {
...
+      // Truncate long commands (more than 160 characters)
+      if (command.length > 160) {
+        return `${toolName}: ${command.substring(0, 160)}...`;
       }
-      return `${toolName}: ${tool.args.command || ""}`;
+      return `${toolName}: ${command}`;
     }
 
-    case "grep": {
+    case "write_file": {
```
bracesproul, line 87: are these tools from deep agents?

bracesproul (approved):
if you tested locally and it works, lgtm w/ one question

bracesproul:
@open-swe can you plz review this PR for any code quality or security issues

bracesproul:
kk can prob ignore most of those. they seem like low pri tasks

apps/cli/src/logger.ts
Palashio: yep, they're all tools defined in deep agents

PR #797, opened 2025-08-24
feat: open swe v2 in js!

Palashio:
This PR creates the coding agent in javascript. It uses a version of the deep agents package that's in an RC build.


apps/open-swe-v2/src/agent.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getCodingInstructions } from "./prompts.js";
+import { createCodingAgentPostModelHook } from "./post-model-hook.js";
+import { CodingAgentState } from "./state.js";
+import { executeBash, httpRequest, webSearch } from "./tools.js";
+
+if (process.env.LANGCHAIN_TRACING_V2 !== "false") {
+  process.env.LANGCHAIN_TRACING_V2 = "true";
+  if (!process.env.LANGCHAIN_PROJECT) {
+    process.env.LANGCHAIN_PROJECT = "coding_agent";
+  }
```
bracesproul: dont want to do this. should allow users to set their own

apps/open-swe-v2/src/agent.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const codingInstructions = getCodingInstructions();
+const postModelHook = createCodingAgentPostModelHook();
+
+const agent = createDeepAgent({
+  tools: [executeBash, httpRequest, webSearch],
+  instructions: codingInstructions,
```
bracesproul, line 14: missing a lot of tools which you define elsewhere. e.g. `str_replace_based_edit_tool` or `grep`

is that intentional?

apps/open-swe-v2/src/agent.ts
```diff
@@ -0,0 +1,29 @@
...
+  subagents: [code_reviewer_agent, test_generator_agent],
```
bracesproul: prefer camel case (i pushed a fix)

apps/open-swe-v2/src/agent.ts
```diff
@@ -0,0 +1,29 @@
...
+  isLocalFileSystem: true,
+  postModelHook: postModelHook,
+  stateSchema: CodingAgentState,
+}).withConfig({ recursionLimit: 1000 }) as any;
```
bracesproul, line 19: having to cast as `any` here & elsewhere when defining state is a very bad sign. this means something is def wrong in the deep agents package

apps/open-swe-v2/src/command-safety.ts
```diff
@@ -0,0 +1,113 @@
...
+  typeof CommandSafetyValidationSchema
+>;
+
+let anthropicClient: ChatAnthropic | null = null;
+
+try {
+  const anthropicApiKey = process.env.ANTHROPIC_API_KEY;
+  if (anthropicApiKey) {
+    anthropicClient = new ChatAnthropic({
+      model: "claude-3-5-haiku-latest",
```
bracesproul, line 33: i updated to haiku. shouldnt need more expensive/powerful models for this?

apps/open-swe-v2/src/constants.ts
bracesproul, line 1: ya none of these tools are provided to the actual agent except `execute_bash`?

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
+import { AIMessage, AIMessageChunk } from "@langchain/core/messages";
+import { interrupt } from "@langchain/langgraph";
+import { WRITE_COMMANDS } from "./constants.js";
+import { CodingAgentStateHelpers, type CodingAgentStateType } from "./state.js";
+
+interface StateType extends CodingAgentStateType {
+  todos: any[];
+  files: Record<string, string>;
+  messages: any[];
```
bracesproul: no bueno no bueno

use types

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+interface ToolCall {
+  name?: string;
+  args?: Record<string, any>;
+  [key: string]: any;
```
bracesproul: why extra keys? should be able to just use the imported `ToolCall` from langchain core

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export function createCodingAgentPostModelHook() {
+  /**
+   * Post model hook that checks for write tool calls and uses caching to avoid
+   * redundant approval prompts for the same command/directory combinations.
+   */
+  async function postModelHook(
+    state: StateType,
+    _model?: any,
```
bracesproul: can just drop if not used (i pushed a commit)

apps/open-swe-v2/src/post-model-hook.ts
bracesproul: instead this should extend the zod object and add the new fields, then convert to a type

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    if (
+      !(
+        lastMessage instanceof AIMessage ||
+        lastMessage instanceof AIMessageChunk
+      ) ||
+      !lastMessage.tool_calls
+    ) {
+      return state;
+    }
```
bracesproul: prefer the `isAIMessage` func

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    if (!state.approved_operations) {
+      state.approved_operations = { cached_approvals: new Set<string>() };
```
bracesproul: dont do this. instead define a new var `approved_operations` and return that.

you'll also need to define a reducer for these state fields

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+    const approvedToolCalls: ToolCall[] = [];
+
+    for (const toolCall of lastMessage.tool_calls) {
+      const toolName = toolCall.name || "";
+      const toolArgs = toolCall.args || {};
+
+      // Skip tool calls without a name
+      if (!toolCall.name) {
+        continue;
+      }
```
bracesproul: throw err here. shouldnt happen, so if it does that means something went wrong

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+        approvedToolCalls.some((approved) => approved.name === toolCall.name),
+      );
+
+      const MessageClass =
+        lastMessage instanceof AIMessageChunk ? AIMessageChunk : AIMessage;
+      const newMessage = new MessageClass({
+        content: lastMessage.content,
+        tool_calls: originalToolCalls,
+        additional_kwargs: lastMessage.additional_kwargs,
+      });
```
bracesproul: can just do `AIMessage` here

remember to include all the fields from `lastMessage` too
```suggestion
      const newMessage = new AIMessage({
        ...lastMessage
        tool_calls: originalToolCalls,
      });
```

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      const newMessages = [...messages.slice(0, -1), newMessage];
+      state.messages = newMessages;
```
bracesproul, line 94: same as above, and other comments left on deepagents pr

can just return the new message instead of returning all of the old messages state and re-assigning the state object

apps/open-swe-v2/src/types.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Type definitions for Open SWE V2 coding agent
+ */
+
+// Command argument types
+export interface FileEditCommandArgs {
+  file_path?: string;
+  path?: string;
+  [key: string]: any;
```
bracesproul: why these extra `any` keys?

package.json
```diff
@@ -27,5 +27,8 @@
     "langsmith": "^0.3.48",
     "jsonwebtoken": "^9.0.2"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "deepagents": "rc"
+  }
```
bracesproul: dont add deps to root

apps/open-swe-v2/src/agent.ts
Palashio: i also just looked and it's not used anywhere so i'm removing

apps/open-swe-v2/src/agent.ts
Palashio: the idea is that those tools are ways for a deep agent to work on a file system (inherent to the deep agent) but these are specific to the coding agent. this setup was implemented by harrison on python deep agents where he defined the tools for a local filesystem.

apps/open-swe-v2/src/agent.ts
Palashio: will look into this and update the deep agents package + here after (created issue)

apps/open-swe-v2/src/post-model-hook.ts
Palashio: awesome

apps/open-swe-v2/src/constants.ts
Palashio: they're both fed to the post model hook as a list of commands that we want to prompt for. makes it easy for us to change that globally. i'll change the comments so it's easier to understand.

apps/open-swe-v2/src/agent.ts
Palashio: those tools are defined as a prat of the deep agents rc build right now, going to merge in later

apps/open-swe-v2/src/post-model-hook.ts
Palashio: for some reason i'm getting an issue where it doesn't want me to return partial state and only the full state? might need to be something i need to change with the deep agents js package.

bracesproul: likely a typing issue. i see on `main` of deep agents youre properly setting the post model hook to `Partial<DeepAgentsState>` is that the right place?

apps/open-swe-v2/src/agent.ts
bracesproul: the `LANGCHAIN_PROJECT` env var is used to specify whihc langsmith tracing project the runs should go to

apps/open-swe-v2/src/agent.ts
```diff
@@ -0,0 +1,22 @@
+import "@langchain/langgraph/zod";
+import { createDeepAgent } from "deepagents";
+import "dotenv/config";
```
bracesproul: dont need if running in langgraph server
```suggestion
```

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,100 @@
+import { AIMessage, AIMessageChunk } from "@langchain/core/messages";
+import { interrupt } from "@langchain/langgraph";
+import { WRITE_COMMANDS } from "./constants.js";
+import { AgentStateHelpers, type CodingAgentStateType } from "./state.js";
+import { ToolCall } from "@langchain/core/messages/tool";
+import { ApprovedOperations } from "./types.js";
+
+type StateType = CodingAgentStateType;
```
bracesproul: why redefine?

apps/open-swe-v2/src/post-model-hook.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function createAgentPostModelHook() {
+  /**
+   * Post model hook that checks for write tool calls and uses caching to avoid
+   * redundant approval prompts for the same command/directory combinations.
+   */
+  async function postModelHook(state: StateType): Promise<StateType> {
```
bracesproul: ah the full vs partial issue might be here. node return types should always be:
```suggestion
  async function postModelHook(state: StateType): Promise<Partial<StateType>> {
```

apps/open-swe-v2/src/post-model-hook.ts
bracesproul: needs fix

bracesproul (approved):
lgtm w/ a couple comments. just make sure to test locally before merging

apps/open-swe-v2/src/post-model-hook.ts
Palashio: that didn't fix it, i'm going to check deep agents js to see what's up (figuring it's something inconsistent there)

PR #798, opened 2025-08-24
feat: Enhanced Ollama Provider Integration (builds on #780)

Chesars:


Chesars:
Just closed this PR. Tried it multiple models: qwen3:0.6b, qwen3:8b/phi3:3.8b (decent quality but 10-30s per call). Larger models like llama3.1:70b could work but require 16GB+ RAM, high-end CPUs/GPUs, and still take 30-60s per call. While technically possible with powerful hardware, the cost and latency make cloud APIs the better practical choice for agent workflows

PR #814, opened 2025-08-30
chore: Remove logs in restart run route

bracesproul:

PR #820, opened 2025-09-01
chore: Remove logs in restart run route

Chesars:
Small cleanup to the `restart-run` API route.
Removed temporary `console.log/dir` and a stray `console.error`, dropped the `eslint-disable no-console`, and switched to a bare `catch {}`. No functional changes—responses stay the same.


Chesars:
Hi, I'm closing this PR as duplicate of #814.

PR #822, opened 2025-09-02
docs: fix webhook URL path in development setup docs

ks093:
Github webhooks trigger on `webhooks/github` instead of `webhook/github` as mentioned in the documentation.

Code reference -> 
https://github.com/langchain-ai/open-swe/blob/998655385b192ab9711e87cc25113911b176318a/apps/open-swe/src/routes/github/unified-webhook.ts#L67


bracesproul (approved):
good catch, thank you!

PR #901, opened 2026-02-07
fix: add langgraph-cli dependency, fix ChatAnthropic import

aran-yogesh:
model: Opus-4.6, 20,00 tokens


apps/agent/agent/server.py
```diff
@@ -912,7 +913,9 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
 
     logger.info("Returning agent with sandbox for thread %s", thread_id)
     return create_deep_agent(
-        model=None,  # TODO: Actually pass a model here
+        model=ChatAnthropic(
+            model="claude-opus-4-6", max_tokens=20_000
+        ),  # TODO: Actually pass a model here
```
bracesproul, line 918: should drop the comment since you're passing the model now
```suggestion
        ),
```

PR #922, opened 2026-02-09
feat: add error normalization middleware for tool and model calls [Closes #904]

### LINKED ISSUE #904: Add middleware to catch tool errors, dont crash run
We should wrap all tool calls the agent makes via a middleware, and catch any errors from tools so they don't crash the run.

See [this example](https://github.com/langchain-ai/langchainplus/tree/main/agent-builder-graphs/agent_builder_graphs/shared_middleware/error.py) for a starting place on how to handle this.

We should use the `awrap_tool_call` (also implement sync version in `wrap_tool_call` that just wraps the tool call in a `try`/`except`, and if an error occurs it should return a `ToolMessage` with the error message in the `content` field, `tool_call_id` (see the linked file above for how to handle) and `status="error"`

aran-yogesh:
- Adds ErrorNormalizationMiddleware that wraps tool calls in try/except, converting unhandled exceptions into structured ToolMessage error responses instead of crashing the agent run
- Supports both sync (wrap_tool_call) and async (awrap_tool_call) tool execution paths
- Model call errors (invalid API key, rate limits, etc.) are logged but re-raised so they surface to the caller appropriately

### DIFF
--- a/apps/agent/agent/middleware/tool_error_handler.py
+++ b/apps/agent/agent/middleware/tool_error_handler.py
@@ -78,7 +78,7 @@ class ToolErrorMiddleware(AgentMiddleware):
     ) -> ToolMessage | Command:
         try:
             return handler(request)
-        except: 
+        except Exception as e: 
             logger.exception("Error during tool call handling; request=%r", request)
             data = _to_error_payload(e, request)
             return ToolMessage(
@@ -94,7 +94,7 @@ class ToolErrorMiddleware(AgentMiddleware):
     ) -> ToolMessage | Command:
         try:
             return await handler(request)
-        except:  
+        except Exception as e:  
             logger.exception("Error during tool call handling; request=%r", request)
             data = _to_error_payload(e, request)
             return ToolMessage(


apps/agent/agent/middleware/tool_error_handler.py
```diff
@@ -0,0 +1,104 @@
...
+    state_schema = AgentState
+
+    def wrap_tool_call(
+        self,
+        request: ToolCallRequest,
+        handler: Callable[[ToolCallRequest], ToolMessage | Command],
+    ) -> ToolMessage | Command:
+        try:
+            return handler(request)
+        except Exception as e:  # noqa: BLE001
```
bracesproul: if `e` isn't used i think you can just do this
```suggestion
        except:
```
same for below too

bracesproul (approved):

apps/agent/agent/middleware/tool_error_handler.py
aran-yogesh: e is being used below with  data = _to_error_payload(e, request)

PR #936, opened 2026-02-11
feat: add background execution guidance for long-running commands

aran-yogesh:


apps/agent/agent/prompt.py
```diff
@@ -10,15 +13,37 @@
 - Use `{working_dir}` as your working directory for all operations
 
 
+---
+##  Dependency Installation
```
bracesproul, line 13: dont do this, the agent should be able to figure out which package manager/install command is used in the project

it also knows how to install pip/npm/yarn so this info isn't really helpful

apps/agent/agent/prompt.py
```diff
@@ -1,3 +1,6 @@
+
```
bracesproul: unnecessary spaces

apps/agent/agent/prompt.py
```diff
@@ -10,15 +13,37 @@
...
 
 ### Code Style
 
 - NEVER add inline comments to code
 - Any docstrings on functions you add or modify must be VERY concise (1 line preferred)
 
-### Installing Dependencies
 
-Do **not** run dependency installation commands (e.g. `uv run poe install-deps`, `npm install`, `yarn install`) in this environment. They are not supported in the sandbox and will fail or behave unexpectedly. Assume dependencies are already available or skip setup steps that install them.
+### Running Long Commands
```
bracesproul: not sure what this means? would drop

PR #944, opened 2026-02-13
fix: resolve circular import in commit_and_open_pr

aran-yogesh:


apps/agent/agent/utils/sandbox.py
```diff
@@ -0,0 +1,55 @@
...
+    multiple agent invocations.
+
+    Args:
+        sandbox_id: Optional existing sandbox ID to connect to.
+                   If None, creates a new sandbox.
+
+    Returns:
+        SandboxBackendProtocol instance
+    """
+    from ..integrations.langsmith import LangSmithProvider
```
bracesproul: why not just put it in this file?

PR #947, opened 2026-02-13
feat: route LangChain OSS/Applied AI/Docs issues to correct repos

### LINKED ISSUE #885: Open SWE v3 - Deepagents
We're going to revamp Open SWE to now be built on top of [deepagents](https://github.com/langchain-ai/deepagents) & have much better support for running LangSmith inside an agent.

To start, we'll just focus on the Linear & GitHub integration, then towards running a full dev env inside the sandbox, and finally connecting it to a UI and improving the UX.

Tasks:
- [x] #888
- [x] #886
- [x] #896
- [x] #892
- [x] #887
- [x] #897
- [x] #904
- [x] #898
- [x] #899
- [x] #900
- [x] #934
- [x] #937
- [x] #946
- [x] #928
- [x] #961
- [x] #965
- [x] #959
- [x] #953
- [x] #890
- [x] #960
- [x] #968
- [x] #889
- [x] #989
- [x] #945 
- [x] #1000
- [x] #990
- [ ] Support parsing images from github comments
- [x] #997
- [x] #964

aran-yogesh:
closes #885

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
   check-spelling:
     name: Check code spelling
     runs-on: ubuntu-latest
+    if: hashFiles('apps/open-swe/package.json') != ''
     steps:
       - uses: actions/checkout@v4
       - uses: codespell-project/actions-codespell@v2
@@ -93,6 +94,7 @@ jobs:
   dev-server-check:
     name: Check dev server startup
     runs-on: ubuntu-latest
+    if: hashFiles('apps/open-swe/package.json') != ''
     defaults:
       run:
         working-directory: apps/open-swe


apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,72 @@ def get_service_jwt_token_for_user(
...
+        if "owner" in config and "name" in config:
+            return config
+        # New nested format: team_id maps to structure with projects
+        if "projects" in config and project_name:
+            return config["projects"].get(project_name)
+        if "default" in config:
+            return config["default"]
+
+    # Try team_name
+    if team_name and team_name in LINEAR_TEAM_TO_REPO:
```
bracesproul: this does the same thing, instead of having both `team_name` and `team_id` defined, can we just pass one? just figure out which is the right field to use, then only pass that one to this function

apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,72 @@ def get_service_jwt_token_for_user(
...
+        # Legacy flat format: team_name maps directly to repo config
+        if "owner" in config and "name" in config:
+            return config
+        # New nested format: team_name maps to structure with projects
+        if "projects" in config and project_name:
+            return config["projects"].get(project_name)
+        if "default" in config:
+            return config["default"]
+
+    return None
```
bracesproul: i think this should return the default value of langchainplus?
```suggestion
    return {"owner": "langchain-ai", "name": "langchainplus"}
```

apps/agent/agent/webapp.py
```diff
@@ -716,12 +776,28 @@ async def linear_webhook(  # noqa: PLR0911, PLR0912, PLR0915
...
-    if team_id and team_id in LINEAR_TEAM_TO_REPO:
-        repo_config = LINEAR_TEAM_TO_REPO[team_id]
-    elif team_name and team_name in LINEAR_TEAM_TO_REPO:
-        repo_config = LINEAR_TEAM_TO_REPO[team_name]
+    project = issue.get("project", {}) if issue else {}
+    project_name = ""
+    if isinstance(project, dict):
+        project_name = project.get("name", "") or ""
+    elif isinstance(project, str):
+        project_name = project
```
bracesproul: why have this type checking? just figure out if project will be a dict or a string, then only account for that

unless the linear api returns both dict and string (which i doubt) you should be able to just figure out which one it is and use that

apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,72 @@ def get_service_jwt_token_for_user(
...
+    },
+    "Docs": {
+        "default": {"owner": "langchain-ai", "name": "docs"}
+    },
 }
 
 
+def get_repo_config_from_team_mapping(
+    team_name: str, team_id: str, project_name: str = ""
+) -> dict[str, str] | None:
```
bracesproul: returning default value, not typed as None anymore
```suggestion
) -> dict[str, str]:
```

apps/agent/agent/webapp.py
bracesproul: can delete this if you return a default value in the `get_repo_config_from_team_mapping` method. also, not sure where the logic above where it pulls from issue names, but we never supported that
```suggestion
```

apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,64 @@ def get_service_jwt_token_for_user(
     return jwt.encode(payload, X_SERVICE_AUTH_JWT_SECRET, algorithm="HS256")
 
 
-LINEAR_TEAM_TO_REPO: dict[str, dict[str, str]] = {
+LINEAR_TEAM_TO_REPO: dict[str, dict[str, Any] | dict[str, str]] = {
+    # Test workspaces (legacy format for backward compatibility)
     "Brace's test workspace": {"owner": "langchain-ai", "name": "open-swe"},
     "Yogesh-dev": {"owner": "aran-yogesh", "name": "TalkBack"},
+
+    # Production team/project mappings
```
bracesproul: nit, unnecessary comment
```suggestion
```

apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,64 @@ def get_service_jwt_token_for_user(
     return jwt.encode(payload, X_SERVICE_AUTH_JWT_SECRET, algorithm="HS256")
 
 
-LINEAR_TEAM_TO_REPO: dict[str, dict[str, str]] = {
+LINEAR_TEAM_TO_REPO: dict[str, dict[str, Any] | dict[str, str]] = {
+    # Test workspaces (legacy format for backward compatibility)
```
bracesproul: same, can drop
```suggestion
```

apps/agent/agent/webapp.py
```diff
@@ -69,12 +69,64 @@ def get_service_jwt_token_for_user(
...
+
+    # Legacy flat format: team maps directly to repo config
+    if "owner" in config and "name" in config:
+        return config
+
+    # New nested format: team maps to structure with projects
+    if "projects" in config and project_name:
+        return config["projects"].get(project_name)
+
+    # Team has default repo (no project needed)
```
bracesproul:
```suggestion
    if "owner" in config and "name" in config:
        return config

    if "projects" in config and project_name:
        return config["projects"].get(project_name)

```

apps/agent/agent/webapp.py
```diff
@@ -714,27 +766,24 @@ async def linear_webhook(  # noqa: PLR0911, PLR0912, PLR0915
         return {"status": "ignored", "reason": "No issue data in comment"}
 
     team = issue.get("team", {})
-    team_id = team.get("id", "") if team else ""
     team_name = team.get("name", "") if team else ""
+    project = issue.get("project")
+    project_name = project.get("name", "") if project else ""
+
+    # Look up repository configuration from team/project mapping
```
bracesproul:
```suggestion
```

PR #950, opened 2026-02-14
fix: clone error when repo directory already exists

### LINKED ISSUE #946: Fix clone error if repo already exists
Right now, it automatically clones the repo when the agent starts
We should catch an error if one throws when cloning
Idk what the error message is, if it's one where we can reliably determine that it was because the repo is already cloned, then skip the next step
If not, we should run `ls` and check if the repo exists in the ouptut

Once we've confirmed the repo exists, we should `cd` into it, check if there are any uncommitted changes, if there are, do nothing, if there aren't, run `git pull` (use a util func for this)

aran-yogesh:
closes #946

### DIFF
--- a/apps/agent/agent/server.py
+++ b/apps/agent/agent/server.py
@@ -80,11 +80,12 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
         raise ValueError(msg)
 
     repo_dir = f"/workspace/{repo}"
+    auth_url = f"https://git:{token}@github.com/{owner}/{repo}.git"
+    clean_url = f"https://github.com/{owner}/{repo}.git"
 
     is_git_repo = await loop.run_in_executor(None, is_valid_git_repo, sandbox_backend, repo_dir)
 
     if not is_git_repo:
-        # Directory missing or not a valid git repo - remove and clone
         logger.warning(
             "Repo directory missing or not a valid git repo at %s, removing", repo_dir
         )
@@ -98,9 +99,7 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
         except Exception:
             logger.exception("Failed to remove invalid directory")
             raise
-        # Fall through to clone below
     else:
-        # Valid git repo exists, check for uncommitted changes
         logger.info("Repo exists at %s, checking for uncommitted changes", repo_dir)
         has_changes = await loop.run_in_executor(
             None, git_has_uncommitted_changes, sandbox_backend, repo_dir
@@ -110,13 +109,8 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
             logger.warning("Repo has uncommitted changes at %s, skipping pull", repo_dir)
             return repo_dir
 
-        # No uncommitted changes, safe to pull
         logger.info("Repo is clean, pulling latest changes from %s/%s", owner, repo)
 
-        auth_url = f"https://git:{token}@github.com/{owner}/{repo}.git"
-        clean_url = f"https://github.com/{owner}/{repo}.git"
-
-        # Set authenticated URL for private repos, then restore clean URL.
         try:
             await loop.run_in_executor(
                 None,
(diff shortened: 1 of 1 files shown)


apps/agent/agent/utils/github.py
```diff
@@ -20,6 +20,28 @@ def _run_git(sandbox_backend: Any, repo_dir: str, command: str) -> Any:
     return sandbox_backend.execute(f"cd {repo_dir} && {command}")
 
 
+def repo_directory_exists(sandbox_backend: Any, repo_dir: str) -> bool:
```
bracesproul: add proper types, not `Any`

apps/agent/agent/server.py
```diff
@@ -76,44 +82,54 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
...
-        if status_result.exit_code == 0 and not status_result.output.strip():
+            if has_changes:
+                logger.warning("Repo has uncommitted changes at %s, skipping pull", repo_dir)
+                return repo_dir
+
+            # No uncommitted changes, safe to pull
+            logger.info("Repo is clean, pulling latest changes from %s/%s", owner, repo)
+
+            # CRITICAL: Ensure remote URL doesn't contain token (clean up from previous runs)
+            clean_url = f"https://github.com/{owner}/{repo}.git"
```
bracesproul: this will fail if it's a private repo. you need to use the prev pattern we implement where it adds the token to the url so that it can clone, then after cloning reset the url to this clean one without the token

apps/agent/agent/utils/github.py
```diff
@@ -20,6 +20,28 @@ def _run_git(sandbox_backend: Any, repo_dir: str, command: str) -> Any:
...
+    """Check if repository directory exists."""
+    safe_repo_dir = shlex.quote(repo_dir)
+    result = sandbox_backend.execute(f"test -d {safe_repo_dir} && echo exists")
+    return result.exit_code == 0 and "exists" in result.output
+
+
+def is_valid_git_repo(sandbox_backend: Any, repo_dir: str) -> bool:
```
bracesproul: i would remove the `repo_directory_exists` method and just use the is valid git

if it's not a valid git repo, or it doesnt exist, is valid git will catch that

if it exists and its a valid git repo, it'll succeed

apps/agent/agent/server.py
```diff
@@ -76,88 +81,105 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
...
-        )
-    except Exception:
-        logger.exception("Failed to execute check command in sandbox")
-        raise
+    is_git_repo = await loop.run_in_executor(None, is_valid_git_repo, sandbox_backend, repo_dir)
 
-    if check_result.exit_code == 0 and "exists" in check_result.output:
-        logger.info("Repo already exists at %s, pulling latest changes", repo_dir)
+    if not is_git_repo:
+        # Directory missing or not a valid git repo - remove and clone
```
bracesproul: these comments arent necessary, please remove all

apps/agent/agent/server.py
```diff
@@ -76,88 +81,105 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
...
-                None,
-                sandbox_backend.execute,
-                f"cd {repo_dir} && git remote set-url origin {clean_url}",
-            )
-        except Exception:
-            logger.exception("Failed to set remote URL after clone")
-            raise
+    # Directory doesn't exist or was removed - clone it
+    logger.info("Cloning repo %s/%s to %s", owner, repo, repo_dir)
+    clone_url = f"https://git:{token}@github.com/{owner}/{repo}.git"
```
bracesproul: this url is already defined above, reuse it?

apps/agent/agent/server.py
```diff
@@ -76,88 +81,105 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
...
+        logger.exception("Failed to execute git clone")
+        raise
+
+    if result.exit_code != 0:
+        msg = f"Failed to clone repo {owner}/{repo}: {result.output}"
+        logger.error(msg)
+        raise RuntimeError(msg)
+
+    # Clean up the remote URL to remove token
+    clean_url = f"https://github.com/{owner}/{repo}.git"
```
bracesproul: same here, already defined above

bracesproul (approved):

PR #954, opened 2026-02-17
fix: restore sandbox backend via thread metadata when cache misses

### LINKED ISSUE #928: Stop using in memory `SANDBOX_BACKENDS`, instead fetch from thread metadata
Currently, we cache the sandbox backend in `SANDBOX_BACKENDS` var. we shouldn't do this. instead we should be fetching the sandbox id from the thread metadata, then connecting to it via `_create_langsmith_sandbox`

caching it this way will lead to issues down the line, better to just get the id from metadata which we know will always be defined, then load the sandbox when we need it

aran-yogesh:
Closes #928

### DIFF
--- a/apps/agent/agent/middleware/open_pr.py
+++ b/apps/agent/agent/middleware/open_pr.py
@@ -141,7 +141,7 @@ I've {action} pull request to address this issue:
 
 {last_message_content}"""
                 await comment_on_linear_issue(linear_issue_id, comment)
-            raise ValueError("threadId not found")
+            raise ValueError("No thread_id found in config")
 
         repo_config = configurable.get("repo", {})
         repo_owner = repo_config.get("owner")


apps/agent/agent/middleware/open_pr.py
```diff
@@ -147,7 +147,7 @@ async def open_pr_if_needed(
         repo_owner = repo_config.get("owner")
         repo_name = repo_config.get("name")
 
-        sandbox_backend = SANDBOX_BACKENDS.get(thread_id)
+        sandbox_backend = await get_sandbox_backend(thread_id) if thread_id else None
```
bracesproul: dont return none. if no thread id we should prob error since thats very had

apps/agent/agent/utils/sandbox_state.py
```diff
@@ -2,7 +2,46 @@
...
+client = get_client()
+
 # Thread ID -> SandboxBackend mapping, shared between server.py and middleware
 SANDBOX_BACKENDS: dict[str, Any] = {}
+
+
+async def _get_sandbox_id_from_metadata(thread_id: str) -> str | None:
+    """Fetch sandbox_id from thread metadata."""
+    try:
+        thread = await client.threads.get(thread_id=thread_id)
```
bracesproul: dont do this, instead import `get_config` from `langgraph.config`, then you can do `config.get("metadata", {}).get("sandbox_id")`

apps/agent/agent/utils/sandbox_state.py
```diff
@@ -2,7 +2,46 @@
...
+
+
+async def get_sandbox_backend(thread_id: str) -> Any | None:
+    """Get sandbox backend from cache, or connect using thread metadata."""
+    sandbox_backend = SANDBOX_BACKENDS.get(thread_id)
+    if sandbox_backend:
+        return sandbox_backend
+
+    sandbox_id = await _get_sandbox_id_from_metadata(thread_id)
+    if not sandbox_id:
```
bracesproul, line 36: i would honestly throw an error here. if no ID then that's a major problem and the app wont be able to continue

apps/agent/agent/server.py
bracesproul: duplicated func now. delete this and import from the util file

apps/agent/agent/server.py
```diff
@@ -235,13 +223,29 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
...
 
-    if sandbox_id == SANDBOX_CREATING:
+    if sandbox_id == SANDBOX_CREATING and not sandbox_backend:
         logger.info("Sandbox creation in progress, waiting...")
         sandbox_id = await _wait_for_sandbox_id(thread_id)
 
-    if sandbox_id is None:
+    if sandbox_backend:
+        logger.info("Using cached sandbox backend for thread %s", thread_id)
+        thread = await client.threads.get(thread_id=thread_id)
```
bracesproul: same as above, should use `get_config` here instead of making a request to get the thread

apps/agent/agent/middleware/open_pr.py
```diff
@@ -141,13 +141,13 @@ async def open_pr_if_needed(
 
 {last_message_content}"""
                 await comment_on_linear_issue(linear_issue_id, comment)
-            return None
+            raise ValueError("threadId not found")
```
bracesproul:
```suggestion
            raise ValueError("No thread_id found in config")
```

PR #955, opened 2026-02-18
feat: enforce timeouts on execute tool commands

### LINKED ISSUE #953: auto wrap commands in a timeout
- add a new middleware that wraps tool calls
- in the middleware, check to see if the tool call is the execute tool
- if so, modify the command it's execuring to be wrapped in a timeout, always add `timeout 300s ` to the command
  - ensure you do a regex check to make sure the command the llm generated doesnt already have a timeout set
- add something to the prompt that tells the llm there is always a 5m timeout. if it runs a command that times out and it wants to run it again with a larger timeout, it should append `timeout Ns` to the command

aran-yogesh:
Complete Trace:
https://dev.smith.langchain.com/public/1346a20a-bf93-4e5b-9cb2-54f1c49e6043/r

specific execute:
https://dev.smith.langchain.com/public/34f69c4f-6b42-4cc1-a9b0-b697dd115ebd/r
https://dev.smith.langchain.com/public/e287a858-e69c-45a8-96a0-0bb2416de372/r
https://dev.smith.langchain.com/public/f5564523-9fac-4e47-9d16-cf1f732fdad1/r
https://dev.smith.langchain.com/public/0f8e8000-7a4d-47ec-a4f3-14acaead0398/r

Closes #953

### DIFF
--- a/apps/agent/agent/middleware/timeout_execute_tool.py
+++ b/apps/agent/agent/middleware/timeout_execute_tool.py
@@ -59,7 +59,7 @@ def _overwrite_request_if_needed(request: ToolCallRequest) -> ToolCallRequest:
     args = dict(tool_call.get("args", {}))
     args["command"] = wrapped
     tool_call["args"] = args
-    logger.info("Wrapped execute command with timeout")
+    logger.debug("Wrapped execute command with timeout")
     return request.override(tool_call=tool_call)
 
 


apps/agent/agent/middleware/timeout_execute_tool.py
```diff
@@ -0,0 +1,95 @@
...
+        handler: Callable[[ToolCallRequest], ToolMessage | Command],
+    ) -> ToolMessage | Command:
+        if _get_tool_name(request) != "execute":
+            return handler(request)
+
+        command = _get_command_arg(request)
+        if not command:
+            return handler(request)
+
+        wrapped = _wrap_command(command)
```
bracesproul: if the tool name is `execute` these should never be none

i would just have one function which gets the wrapped command, or return none

also the logic inside `wrap_tool_call` and `awrap_tool_call` is the exact same except one is async and the other is sync. so refactor all the logic out and you should be able to make the `wrap_tool_call` and `awrap_tool_call` method body two lines. something like this:

```python
request = overwrite_or_return_request(request)
return await handler(request)
```
and
```python
request = overwrite_or_return_request(request)
return handler(request)
```

bracesproul (changes requested):

PR #956, opened 2026-02-18
feat: prompt review

aran-yogesh:


apps/agent/agent/prompt.py
```diff
@@ -1,32 +1,246 @@
...
 
 You are operating in a **remote Linux sandbox** at `{working_dir}`.
 
 All code execution and file operations happen in this sandbox environment.
 
-
 **Important:**
 - Use `{working_dir}` as your working directory for all operations
+- The `execute` tool enforces a 5-minute timeout by default (`timeout 300s`)
+- If a command times out and needs longer, rerun it by explicitly appending `timeout Ns`"""
```
bracesproul: it now uses a timeout arg, should update the prompt to reflect that instead of timeout command

apps/agent/agent/prompt.py
```diff
@@ -1,32 +1,246 @@
...
+
+
+# --------------------------------------------------
+# Section: Current Task Overview
+# --------------------------------------------------
+TASK_OVERVIEW_SECTION = """---
+
+### Current Task Overview
+
+You are currently executing a specific task from a pre-generated plan. You have access to:
```
bracesproul: you dont pre-generate a plan, so this is wrong

apps/agent/agent/prompt.py
```diff
@@ -1,32 +1,246 @@
...
+
+# --------------------------------------------------
+# Section: Task Execution Guidelines
+# --------------------------------------------------
+TASK_EXECUTION_SECTION = """---
+
+### Task Execution Guidelines
+
+- You are executing a task from the plan.
+- Previous completed tasks and their summaries contain crucial context — always review them first.
```
bracesproul: same as above, there's no explicit plan step it has to follow

apps/agent/agent/prompt.py
```diff
@@ -1,32 +1,246 @@
...
+# Section: File & Code Management
+# --------------------------------------------------
+FILE_MANAGEMENT_SECTION = """---
+
+### File & Code Management
+
+- **Repository location:** `{working_dir}`
+- All changes are auto-committed — no manual commits needed, and you should never create backup files.
+- Work only within the existing Git repository.
+- Use `install_dependencies` to install dependencies (skip if installation fails). Only call this tool if the task REQUIRES installing dependencies."""
```
bracesproul: this tool doesnt exist

PR #957, opened 2026-02-19
fix: add sandbox_id to run metadata and config

aran-yogesh:
trace link:
https://dev.smith.langchain.com/public/cc8869b6-c846-4d21-a96c-441bf1360fce/r


apps/agent/agent/server.py
```diff
@@ -329,6 +329,10 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
                 raise
 
     SANDBOX_BACKENDS[thread_id] = sandbox_backend
+    # Ensure run metadata includes the active sandbox id for trace visibility.
+    config.setdefault("metadata", {})["sandbox_id"] = sandbox_backend.id
```
bracesproul, line 326: we should be calling `client.threads.update` setting it in the metadata field

i dont think it will be persisted across threads here

apps/agent/agent/server.py
```diff
@@ -329,6 +329,15 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
...
+    # Also add to configurable so it shows on node/tool traces that surface config.
+    config.setdefault("configurable", {})["sandbox_id"] = sandbox_backend.id
+    # Persist on thread metadata as well.
```
bracesproul: dont need these comments
```suggestion
    config.setdefault("metadata", {})["sandbox_id"] = sandbox_backend.id
    config.setdefault("configurable", {})["sandbox_id"] = sandbox_backend.id
```

and are you sure you need to set it on both? i think it should only be set in one (ideally config.configurable)

plz try only setting it on configurable (via `config.setdefault` and `threads.update`) and check if it shows up in the trace

apps/agent/agent/integrations/langsmith.py
```diff
@@ -64,11 +64,45 @@ def _create_langsmith_sandbox(
...
 
     provider = LangSmithProvider(api_key=api_key)
-    return provider.get_or_create(
+    backend = provider.get_or_create(
         sandbox_id=sandbox_id,
         template=template_name,
         template_image=template_image,
     )
+    if sandbox_id is None:
+        _update_thread_sandbox_metadata(backend.id)
```
bracesproul: either always update it, or go `get_config` to see if it's already set in config

the sandbox id arg being none doesnt reliably tell us if it's already set in configurable

apps/agent/agent/integrations/langsmith.py
```diff
@@ -64,11 +64,45 @@ def _create_langsmith_sandbox(
...
+        config = get_config()
+        thread_id = config.get("configurable", {}).get("thread_id")
+        if not thread_id:
+            return
+        client = get_client()
+
+        async def _update() -> None:
+            await client.threads.update(
+                thread_id=thread_id,
+                metadata={"sandbox_id": sandbox_id},
```
bracesproul, line 93: need to standardize on either configurable or metadata

above youre trying to read it from configurable but here youre setting it in metadata. need to pick one. see where we already read it from in the codebase and make sure it's all standard

apps/agent/agent/server.py
```diff
@@ -329,6 +329,12 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
                 raise
 
     SANDBOX_BACKENDS[thread_id] = sandbox_backend
+    config.setdefault("metadata", {})["sandbox_id"] = sandbox_backend.id
+    config.setdefault("configurable", {})["sandbox_id"] = sandbox_backend.id
```
bracesproul: same comment as above, dont set it in both - pick one

apps/agent/agent/server.py
```diff
@@ -329,6 +329,12 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
...
+    await client.threads.update(
```
bracesproul: you dont need to do this because youre already setting it when you create sandbox backend in `_create_langsmith_sandbox` above

PR #963, opened 2026-02-19
feat: add multimodal image support for Linear comments and queued run

### LINKED ISSUE #890: Support multimodal inputs
We should extract any images in linear tickets, and convert them to human messages
to do this we need to:
- add a regex check in the linear webhook that checks to see if any image URLs exist in a ticket (ensure we're supporting multiple images)
- if so, the regex should extract the URLs of the linear images
- convert the image URLs to base64 (i believe there's a way to just make a request to the image url, get back base64)
- append new image blocks to the message we're sending to the agent. see this code:
```python
from langchain_core.messages.content import create_image_block
```

the final content block should look something like this (ensure you properly support multiple images, this code is just a demo to show what it looks like for one)
```python
image_url = regex_code_to_get_url(ticket_content)
image_base_64 = fetch_base_64_from_image_url(image_url)
media_type = get_mime_type_from_image_url()

content = [
  {"type": "text", "text": ticket_content,
  create_image_block(base64=image_base_64, mime_type=media_type)
]
```

aran-yogesh:
follow up comments: 

trace : https://dev.smith.langchain.com/public/35e98d93-cc7f-4388-93a7-5f65b6ca0d4a/r

Queued messages trace:

https://dev.smith.langchain.com/public/08f799ea-6a9c-4e97-b8b9-1a7974ad81fb/r

Closes #890

### DIFF
--- a/apps/agent/agent/utils/multimodal.py
+++ b/apps/agent/agent/utils/multimodal.py
@@ -80,5 +80,4 @@ async def fetch_image_block(
 
 
 def dedupe_urls(urls: list[str]) -> list[str]:
-    deduped: set[str] = set(urls)
-    return list(deduped)
+    return list(dict.fromkeys(urls))


apps/agent/agent/middleware/check_message_queue.py
```diff
@@ -80,11 +84,55 @@ async def check_message_queue_before_model(  # noqa: PLR0911
             thread_id,
         )
 
-        content_blocks = [
-            {"type": "text", "text": msg.get("content", "")}
-            for msg in queued_messages
-            if msg.get("content")
-        ]
+        async def _build_blocks_from_payload(
```
bracesproul: i dont think this method should be nested inside this function. would refactor it to be outside the func

apps/agent/agent/utils/multimodal.py
```diff
@@ -0,0 +1,86 @@
...
+logger = logging.getLogger(__name__)
+
+IMAGE_MARKDOWN_RE = re.compile(r"!\[[^\]]*\]\((https?://[^\s)]+)\)")
+IMAGE_URL_RE = re.compile(
+    r"(https?://[^\s)]+\.(?:png|jpe?g|gif|webp|bmp|tiff)(?:\?[^\s)]+)?)",
+    re.IGNORECASE,
+)
+
+
+def extract_image_urls(text: str) -> list[str]:
```
bracesproul, line 24: can you add unit tests for this?

apps/agent/agent/utils/multimodal.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def _dedupe_urls(urls: list[str]) -> list[str]:
+    seen: set[str] = set()
+    deduped: list[str] = []
+    for url in urls:
+        if url in seen:
+            continue
+        seen.add(url)
+        deduped.append(url)
```
bracesproul: why not just track seen? do we need to track both?

apps/agent/agent/utils/multimodal.py
```diff
@@ -0,0 +1,86 @@
...
+    urls.extend(IMAGE_MARKDOWN_RE.findall(text))
+    urls.extend(IMAGE_URL_RE.findall(text))
+
+    deduped = _dedupe_urls(urls)
+    if deduped:
+        logger.debug("Extracted %d image URL(s)", len(deduped))
+    return deduped
+
+
+def strip_image_urls(text: str, image_urls: list[str]) -> str:
```
bracesproul: i actually think you should leave the image urls in there, bc a user might say something like
```
look at this image:
<url>
```

and the model should be smart enough to understand the attached images are the ones from the url in the text

apps/agent/agent/middleware/check_message_queue.py
```diff
@@ -80,11 +84,55 @@ async def check_message_queue_before_model(  # noqa: PLR0911
...
+                        else:
+                            logger.warning(
+                                "LINEAR_API_KEY not set; cannot authenticate image fetch for %s",
+                                image_url,
+                            )
+                    image_block = await fetch_image_block(
+                        image_url, client, headers=headers
+                    )
+                    if image_block:
+                        blocks.append(image_block)
```
bracesproul: this should all live inside the `fetch_image_block` method

apps/agent/agent/utils/multimodal.py
aran-yogesh: I used a seen set for fast checks and a list to preserve the original order of URLs. Returning only seen would lose ordering. If you prefer, I can simplify to `list(dict.fromkeys(urls))`

bracesproul: ya thats better. make deduped a set, only track that, then convert to a list before returning

apps/agent/agent/utils/multimodal.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.debug("Extracted %d image URL(s)", len(deduped))
+    return deduped
+
+
+
+async def fetch_image_block(
+    image_url: str,
+    client: httpx.AsyncClient,
+    *,
+    linear_api_key: str | None = None,
```
bracesproul: dont pass as an arg, read from env var instead

apps/agent/agent/utils/multimodal.py
```diff
@@ -0,0 +1,88 @@
...
+                logger.warning(
+                    "LINEAR_API_KEY not set; cannot authenticate image fetch for %s",
+                    image_url,
+                )
+        response = await client.get(image_url, headers=headers)
+        response.raise_for_status()
+        content_type = response.headers.get("Content-Type", "").split(";")[0].strip()
+        if not content_type:
+            guessed, _ = mimetypes.guess_type(image_url)
+            content_type = guessed or "application/octet-stream"
```
bracesproul: this isnt an acceptable content type for llm content blocks. just say if `mimetypes.guess_type` doesnt work, ignore the image and log a warning

apps/agent/agent/webapp.py
```diff
@@ -131,6 +133,8 @@ def get_repo_config_from_team_mapping(
     return {"owner": "langchain-ai", "name": "langchainplus"}
 
 
+
+
```
bracesproul: unnecessary spaces? always review your PRs before opening for review to catch simple things like this
```suggestion
```

apps/agent/agent/webapp.py
```diff
@@ -619,6 +678,27 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
         "Please analyze this issue and implement the necessary changes. "
         "When you're done, commit and push your changes."
     )
+    content_blocks: list[dict[str, Any]] = [{"type": "text", "text": prompt}]
```
bracesproul: use the actual content block type for this

apps/agent/agent/webapp.py
```diff
@@ -619,6 +678,27 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
...
+    if image_urls:
```
bracesproul, line 680: this is duplicated code

apps/agent/agent/webapp.py
```diff
@@ -619,6 +678,27 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
...
+        for url in image_urls:
+            if url in seen_urls:
+                continue
+            seen_urls.add(url)
+            deduped_urls.append(url)
+        image_urls = deduped_urls
+        logger.info("Preparing %d image(s) for multimodal content", len(image_urls))
+        logger.debug("Image URLs: %s", image_urls)
+
+        async with httpx.AsyncClient() as client:
```
bracesproul, line 685: this is duplicated code too. you should refactor it out of here and our of your util in the other file so it only has to be defined once

bracesproul (changes requested):

PR #970, opened 2026-02-23
feat: inject AGENTS.md into system prompt

### LINKED ISSUE #959: Add support for agents.md
After cloning a repo we should look to see if a file named `AGENTS.md` exists in the root. If it does, we should read the contents, and set them inside the main system prompt. We should wrap them in XML tags to make it clear what it is to the agent, something like this:
```
The following text is pulled from the repositories AGENTS.md file. It may contain specific instructions and guidelines for how to manage this repository.
<agents_md>
{AGENTS_MD}
</agents_md>
```

Ensure we do this check every time a new run starts so that it's always present in the system prompt.

aran-yogesh:
https://dev.smith.langchain.com/public/009aa054-aac3-46a6-92e2-836681b2b6b0/r

Close #959

### DIFF
--- a/apps/agent/agent/server.py
+++ b/apps/agent/agent/server.py
@@ -325,7 +325,7 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
     SANDBOX_BACKENDS[thread_id] = sandbox_backend
 
     if not repo_dir:
-        msg = "repo_dir is not defined; cannot operate without a cloned repository"
+        msg = "Cannot proceed: no repo was cloned. Set 'repo.owner' and 'repo.name' in the configurable config"
         raise RuntimeError(msg)
 
     linear_issue = config["configurable"].get("linear_issue", {})


apps/agent/agent/utils/agents_md.py
```diff
@@ -0,0 +1,30 @@
...
+    loop = asyncio.get_event_loop()
+    result = await loop.run_in_executor(
+        None,
+        sandbox_backend.execute,
+        f"test -f {safe_agents_path} && cat {safe_agents_path}",
+    )
+    if result.exit_code != 0:
+        return None
+    content = result.output or ""
+    content = content.strip()
```
bracesproul, line 33: if code is not 0, or content is none, run `ls` inside the repo dir, then add a log saying
```
No AGENTS.md found at {repo_dir}/AGENTS.md. ls output: {ls_output}
```

apps/agent/agent/server.py
```diff
@@ -327,6 +330,7 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
     linear_issue = config["configurable"].get("linear_issue", {})
     linear_project_id = linear_issue.get("linear_project_id", "")
     linear_issue_number = linear_issue.get("linear_issue_number", "")
+    agents_md = await read_agents_md_in_sandbox(sandbox_backend, repo_dir)
```
bracesproul, line 334: throw an error above if `repo_dir` isnt defined

aran-yogesh: I kept it non‑fatal if `repo_dir` is missing, since there are valid runs without a repo (or before clone completes). In those cases we just skip AGENTS.md and proceed. I also just added miniaml logging

aran-yogesh: I kept a minimal logging because there might be runs without `repo_dir` (like no repo configured or clone not done yet). In those cases we just skip AGENTS.md and move on, so I didn’t throw.

bracesproul: but if repo dir isnt defined, how would the agent work? it wont be able to clone anything right?

apps/agent/agent/utils/agents_md.py
```diff
@@ -0,0 +1,36 @@
...
+        sandbox_backend.execute,
+        f"test -f {safe_agents_path} && cat {safe_agents_path}",
+    )
+    if result.exit_code != 0:
+        logger.debug("AGENTS.md not found at %s", safe_agents_path)
+        return None
+    content = result.output or ""
+    content = content.strip()
+    if content:
+        logger.info("Loaded AGENTS.md from %s", safe_agents_path)
```
bracesproul: i wouldnt log if it succeeds, only if it fails

apps/agent/agent/server.py
```diff
@@ -139,6 +140,7 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
         logger.info("Repo updated at %s", repo_dir)
         return repo_dir
 
+    # If we reach here, repo doesn't exist or is invalid - clone it
```
bracesproul: unnecessary comment
```suggestion
```

apps/agent/agent/prompt.py
```diff
@@ -72,9 +73,20 @@ def construct_system_prompt(
     working_dir: str,
     linear_project_id: str = "",
     linear_issue_number: str = "",
+    agents_md: str | None = None,
```
bracesproul: can you default this to a string so it matches the other args
```suggestion
    agents_md: str = "",
```

apps/agent/agent/server.py
aran-yogesh: yea thats actually right if the repo_dir isnt there at that stage the agent cant do useful work anyway

apps/agent/agent/server.py
```diff
@@ -324,9 +324,14 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
 
     SANDBOX_BACKENDS[thread_id] = sandbox_backend
 
+    if not repo_dir:
+        msg = "repo_dir is not defined; cannot operate without a cloned repository"
```
bracesproul: the actual issue is that the configurable object doesn't have the `repo` object defined, right? issue message should be something telling users that, bc this wont be very helpful to users

PR #971, opened 2026-02-24
fix: collect only comments since last agent response for follow-up triggers

aran-yogesh:
Linear returns comments newest-first but the old logic assumed oldest-first, causing follow-up @openswe comments to be ignored in favor of the original trigger.

Now sorts by createdAt descending, collects user comments until the first bot message (most recent agent response), then reverses for chronological order.

example branch commits:
https://github.com/aran-yogesh/TalkBack/pull/105

### DIFF
--- a/apps/agent/agent/webapp.py
+++ b/apps/agent/agent/webapp.py
@@ -603,15 +603,11 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
     comment_ids: set[str] = set()
     comment_id_to_index: dict[str, int] = {}
     if comments:
-        last_bot_comment_idx = -1
         for i, comment in enumerate(comments):
             comment_id = comment.get("id", "")
             if comment_id:
                 comment_ids.add(comment_id)
                 comment_id_to_index[comment_id] = i
-            body = comment.get("body", "")
-            if any(body.startswith(prefix) for prefix in bot_message_prefixes):
-                last_bot_comment_idx = i
 
         relevant_comments = []
         trigger_index = None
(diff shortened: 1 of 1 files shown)


apps/agent/agent/webapp.py
```diff
@@ -586,29 +586,30 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
...
+        )
+
+        # Collect user comments since the last agent response.
+        # Iterate newest-first and stop as soon as we see a bot message.
+        recent_user_comments = []
+        for comment in sorted_comments:
             body = comment.get("body", "")
             if any(body.startswith(prefix) for prefix in bot_message_prefixes):
-                last_bot_comment_idx = i
+                break  # Everything after this is from before the last agent response
```
bracesproul: if this breaks after the first mention of the openswe prefix, wouldnt this break after only one comment?

e.g. if it's
comment_1
comment_2
open_swe_comment_3

wouldnt it reverse so open_swe_comment_3 is first, then break right away?

also can you verify this _won't_ trigger the agent if it's not a comment with `@openswe`?

e.g. if the comments look like this:
@openswe (first trigger)
some random comment

will `some random comment` trigger the agent?

bracesproul (commented):
can you refactor this logic into a method `get_recent_comments` then write unit tests that simulate comments to ensure it works as intended

have it return None if there are no comments to process, or a list of comments if there are some to process

apps/agent/agent/webapp.py
aran-yogesh: If the newest comment is a user comment (even with @openswe), it collects comments until it hits a bot response.

A random comment won’t trigger the agent because the webhook

PR #984, opened 2026-02-26
fix: recover from sandbox connection failures by recreating sandbox and recloning repo 

### LINKED ISSUE #960: Recover from sandbox connection failures by recreating the sandbox, updating the stored sandbox ID, and re-cloning the repo
If you create a ticket, tag openswe, let it complete, then go into langsmith and delete the sandbox from that ticket, then send a followup, it'll try to reconnect to the sandbox you just deleted and it'll fail. we should be catching these sandbox not found errors and if they occur, recreate the sandbox, clone the repo, then continue

aran-yogesh:
Closes #960

### DIFF
--- a/apps/agent/agent/server.py
+++ b/apps/agent/agent/server.py
@@ -174,12 +174,13 @@ async def _recreate_sandbox(
     thread_id: str,
     repo_owner: str,
     repo_name: str,
+    *,
     github_token: str | None,
 ) -> tuple[SandboxBackendProtocol, str]:
     """Recreate a sandbox and clone the repo after a connection failure.
 
     Clears the stale cache entry, sets the SANDBOX_CREATING sentinel,
-    creates a fresh sandbox, clones the repo, and updates repo_dir metadata.
+    creates a fresh sandbox, and clones the repo.
     """
     SANDBOX_BACKENDS.pop(thread_id, None)
     await client.threads.update(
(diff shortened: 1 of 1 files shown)


apps/agent/agent/server.py
```diff
@@ -242,6 +243,29 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
                 repo_dir = await _clone_or_pull_repo_in_sandbox(
                     sandbox_backend, repo_owner, repo_name, github_token
                 )
+            except SandboxClientError:
```
bracesproul, line 275: can you refactor this into a util function?

apps/agent/agent/server.py
```diff
@@ -169,6 +170,38 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
...
+        metadata={"sandbox_id": SANDBOX_CREATING},
+    )
+    try:
+        sandbox_backend = await asyncio.to_thread(create_langsmith_sandbox)
+        repo_dir = await _clone_or_pull_repo_in_sandbox(
+            sandbox_backend, repo_owner, repo_name, github_token
+        )
+        await client.threads.update(
+            thread_id=thread_id,
+            metadata={"repo_dir": repo_dir},
```
bracesproul: do we need to update this again? will it already have been set elsewhere?

apps/agent/agent/server.py
```diff
@@ -169,6 +170,38 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
     return repo_dir
 
 
+async def _recreate_sandbox(
+    thread_id: str,
+    repo_owner: str,
+    repo_name: str,
+    github_token: str | None,
```
bracesproul, line 178: i like to follow a pattern where my functions never have >3 positional args. it makes it hard to follow the more positional args are added

can you update this so the 4rd arg (`github_token`) is a named arg and not positional?

PR #992, opened 2026-03-01
fix: skip redundant config check in getRepoAbsolutePath

### LINKED ISSUE #842: skip `config` check
`config` is mostly undefined(causing the check to always fail) and the is already handled in `isLocalMode()`
https://github.com/langchain-ai/open-swe/blob/19debf1b62ab6a3fe67432f99411e0df825d3c2b/packages/shared/src/git.ts#L13

Br1an67:
## Summary

Remove the `config &&` guard in `getRepoAbsolutePath()` so that `isLocalMode()` can fall back to environment variables when `config` is not passed.

## Problem

As noted in #842, the `config` parameter is `undefined` in the majority of call sites (only 2 out of ~30 callers pass it). The current guard `if (config && isLocalMode(config))` short-circuits when `config` is `undefined`, preventing `isLocalMode()` from ever checking the `OPEN_SWE_LOCAL_MODE` environment variable. This means local mode only works for the two callers that happen to pass `config`.

## Solution

Change `if (config && isLocalMode(config))` → `if (isLocalMode(config))`.

`isLocalMode()` already handles `undefined` config by falling back to `isLocalModeFromEnv()`, so this is safe.

## Testing

Added unit tests for `getRepoAbsolutePath` covering:
- Sandbox mode (default path)
- Local mode via environment variable (without config)
- Local mode via config object
- Missing repo name error

All existing and new tests pass.

Fixes #842

### DIFF
--- a/packages/shared/src/__tests__/git.test.ts
+++ b/packages/shared/src/__tests__/git.test.ts
@@ -1,0 +1,54 @@
+import { getRepoAbsolutePath } from "../git.js";
+import { SANDBOX_ROOT_DIR } from "../constants.js";
+
+describe("getRepoAbsolutePath", () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    process.env = { ...originalEnv };
+    delete process.env.OPEN_SWE_LOCAL_MODE;
+    delete process.env.OPEN_SWE_LOCAL_PROJECT_PATH;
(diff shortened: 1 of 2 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

bracesproul:

PR #999, opened 2026-03-02
chore: improve GitHub auth failure logging with exception type 

aran-yogesh:


apps/agent/agent/webapp.py
```diff
@@ -222,6 +222,7 @@ async def get_github_token_for_user(ls_user_id: str, tenant_id: str) -> dict[str
     except httpx.HTTPStatusError as e:
         return {"error": f"HTTP error: {e.response.status_code} - {e.response.text}"}
```
bracesproul, line 225: can you add an error log here too

apps/agent/agent/webapp.py
bracesproul, line 187: and here

apps/agent/agent/webapp.py
```diff
@@ -222,6 +222,7 @@ async def get_github_token_for_user(ls_user_id: str, tenant_id: str) -> dict[str
...
     except Exception as e:  # noqa: BLE001
+        logger.warning("GitHub auth API call failed: %s: %s", type(e).__name__, str(e))
```
bracesproul: should be an error bc it's a failure
```suggestion
        logger.error("GitHub auth API call failed: %s: %s", type(e).__name__, str(e))
```

apps/agent/agent/webapp.py
```diff
@@ -558,7 +559,11 @@ async def process_linear_issue(  # noqa: PLR0912, PLR0915
                 await comment_on_linear_issue(issue_id, comment)
                 return
             else:
-                logger.warning("Auth result has neither token nor auth_url: %s", auth_result)
+                logger.warning(
+                    "Auth result has neither token nor auth_url: %s (error=%r)",
+                    auth_result,
+                    auth_result.get("error"),
```
bracesproul: auth result will have the error field so no need to duplicate it. revert this change here

PR #1002, opened 2026-03-04
chore: resolve github token in get_agent and read from thread metadata

aran-yogesh:


apps/agent/agent/utils/sandbox_state.py
```diff
@@ -15,6 +15,9 @@
 # Thread ID -> SandboxBackend mapping, shared between server.py and middleware
 SANDBOX_BACKENDS: dict[str, Any] = {}
 
+# Thread ID -> plaintext GitHub token, shared between server.py, tools, and middleware
+GITHUB_TOKENS: dict[str, str] = {}
```
bracesproul: dont do this. we shouldnt change how we store & fetch the github tokens

apps/agent/agent/server.py
```diff
@@ -231,28 +233,109 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+    source: str,
+    issue_id: str,
+    *,
+    auth_url: str | None = None,
+    error: str | None = None,
+    email: str | None = None,
+) -> None:
+    """Post an auth failure comment to the appropriate source (e.g. Linear)."""
+    if source == "linear" and issue_id:
+        if auth_url:
```
bracesproul: all of these comments should be the same no matter where we're posting them, so the logic should be:
- construct comment
- then have if statement which calls the specific method to leave the comment

right now you have the comment contents nested inside the linear handler, which means youll have to either refactor later when we add support for more providers, or duplicate the code, both of which we dont want

apps/agent/agent/server.py
```diff
@@ -231,28 +233,109 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
-        github_token = decrypt_token(encrypted_token)
-
     if thread_id is None or not graph_loaded_for_execution(config):
         logger.info("No thread_id or not for execution, returning agent without sandbox")
         return create_deep_agent(
             system_prompt="",
             tools=[],
         ).with_config(config)
 
+    # --- GitHub token resolution ---
```
bracesproul, line 251: there's too much code here. the code in this file should be much simplier:
- call a single function `save_encrypted_token_from_email`
- in that function you should:
1. take the email, exchange it for an id (this should be one function)
2. take the id, exchange it for a token (this should be a second function)
3. take the token, encrypt it, store it in the thread metadata (this should be a 3rd function)

then, for the failure cases where you want to send a linear comment alerting the user, they should also all be abstracted into a method (`leave_failure_comment(source, message)`) and inside of that method it should have the if/else statements for where to send the message

the code should end up looking something like this:
```python
await save_encrypted_token_from_email(email, source)
```
then inside the file with the `save_encrypted_token_from_email` func:
```python
ls_user_id = await get_ls_user_id_from_email(email)
if not ls_user_id:
    err_comment = "Whatever the comment is"
    await _handle_auth_failure(err_comment, source)
    raise ValueError("No ls_user_id found from email")

auth_res = await get_token_or_auth_url_from_id(ls_user_id)
if auth_res.get("auth_url"):
    auth_url_comment = "Plz auth: {auth_url}"
    await _handle_auth_failure(err_comment, source)
    raise ValueError("User not authenticated.")

if not auth_res.get("token"):
    missing_token_comment = "No token found"
    await _handle_auth_failure(missing_token_comment, source)
    raise ValueError("No token found")

encrypted_token = encrypt_token(token)
await client.threads.update(...) # set the token on the thread
```

apps/agent/agent/utils/sandbox_state.py
bracesproul: also it doesnt even look like youre using this anywhere? why is it here?

apps/agent/agent/middleware/open_pr.py
```diff
@@ -153,6 +163,17 @@ async def open_pr_if_needed(
                 body=pr_body,
             )
 
+        linear_issue = configurable.get("linear_issue", {})
```
bracesproul: dont do this. the agent should be the one commenting not a hard-coded comment

apps/agent/agent/middleware/open_pr.py
bracesproul, line 1: remove all changes to this file

apps/agent/agent/tools/commit_and_open_pr.py
```diff
@@ -168,10 +170,27 @@ def commit_and_open_pr(
                     "pr_url": None,
                 }
 
-        encrypted_token = configurable.get("github_token_encrypted")
+        encrypted_token = None
```
bracesproul: lets refactor this into a function `get_github_token` which handles trying to get it from config, falling back to fetching the token from the client, and decrypting all in one. i suspect it'll be useful elsewhere too

apps/agent/agent/utils/auth.py
```diff
@@ -0,0 +1,151 @@
...
+
+logger = logging.getLogger(__name__)
+
+LANGSMITH_API_KEY = os.environ.get("LANGSMITH_API_KEY_PROD", "")
+LANGSMITH_API_URL = os.environ.get("LANGSMITH_ENDPOINT", "https://api.smith.langchain.com")
+LANGSMITH_HOST_API_URL = os.environ.get("LANGSMITH_HOST_API_URL", "https://api.host.langchain.com")
+GITHUB_OAUTH_PROVIDER_ID = os.environ.get("GITHUB_OAUTH_PROVIDER_ID", "")
+X_SERVICE_AUTH_JWT_SECRET = os.environ.get("X_SERVICE_AUTH_JWT_SECRET", "")
+
+logger.info(
```
bracesproul: debug, dont need this logging in prod
```suggestion
logger.debug(
```

apps/agent/agent/utils/auth.py
```diff
@@ -0,0 +1,151 @@
...
+        logger.error("GitHub auth failed: GITHUB_OAUTH_PROVIDER_ID is not configured")
+        return {"error": "GITHUB_OAUTH_PROVIDER_ID not configured"}
+
+    try:
+        service_token = get_service_jwt_token_for_user(ls_user_id, tenant_id)
+
+        headers = {
+            "X-Service-Key": service_token,
+            "X-Tenant-Id": tenant_id,
+            "X-User-Id": ls_user_id,
```
bracesproul: dont pass this, its very confusing but the tldr is langsmith has two concept of user ids

`user_id`/`X-User-Id` is actually a different value from `ls_user_id`

you can just avoid passing and it should work (i think)
```suggestion
```

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
 DEFAULT_RECURSION_LIMIT = 1_000
 
 
+async def leave_failure_comment(source: str, issue_id: str, message: str) -> None:
+    """Leave an auth failure comment for the appropriate source."""
+    if source == "linear" and issue_id:
+        logger.info("Posting auth failure comment to Linear issue %s (source=%s)", issue_id, source)
+        await comment_on_linear_issue(issue_id, message)
```
bracesproul: lets add an `else` which throws an error with a message like "Unknown source: {source}"

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+    """Encrypt a GitHub token and store it on the thread metadata."""
+    encrypted = encrypt_token(token)
+    await client.threads.update(
+        thread_id=thread_id,
+        metadata={"github_token_encrypted": encrypted},
+    )
+    return encrypted
+
+
+async def save_encrypted_token_from_email(
```
bracesproul: move this to the `auth.py` file

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+    user_info = await get_ls_user_id_from_email(email)
+    return user_info.get("ls_user_id"), user_info.get("tenant_id")
+
+
+async def get_token_or_auth_url_from_id(ls_user_id: str, tenant_id: str) -> dict[str, Any]:
+    """Resolve a GitHub token or auth URL from LangSmith user + tenant ids."""
+    return await get_github_token_for_user(ls_user_id, tenant_id)
+
+
+async def persist_encrypted_github_token(thread_id: str, token: str) -> str:
```
bracesproul: move this to the `auth.py` file

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+
+async def get_langsmith_user_from_email(email: str) -> tuple[str | None, str | None]:
+    """Resolve LangSmith user + tenant ids from an email."""
+    user_info = await get_ls_user_id_from_email(email)
+    return user_info.get("ls_user_id"), user_info.get("tenant_id")
+
+
+async def get_token_or_auth_url_from_id(ls_user_id: str, tenant_id: str) -> dict[str, Any]:
+    """Resolve a GitHub token or auth URL from LangSmith user + tenant ids."""
+    return await get_github_token_for_user(ls_user_id, tenant_id)
```
bracesproul: these dont need to be wrapped. just call the underlying functions directly

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
 DEFAULT_RECURSION_LIMIT = 1_000
 
 
+async def leave_failure_comment(source: str, issue_id: str, message: str) -> None:
```
bracesproul: move this to a util file. prob file to put in `auth.py`

bracesproul: also, instead of accepting issue_id as an arg, you should just get the id from config inside the linear if statement in this method.

apps/agent/agent/server.py
```diff
@@ -231,28 +234,132 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+        raise RuntimeError(msg)
+    linear_issue = config["configurable"].get("linear_issue", {})
+    linear_issue_id = linear_issue.get("id", "")
+
+    try:
+        github_token, new_encrypted = await save_encrypted_token_from_email(
+            user_email,
+            source,
+            linear_issue_id,
+            thread_id,
```
bracesproul: see my comment above but dont pass linear issue id here, and same with thread id. just get both from the config when you actually need to use them

apps/agent/agent/tools/commit_and_open_pr.py
```diff
@@ -23,6 +24,24 @@
 logger = logging.getLogger(__name__)
 
 
+def get_github_token(config: dict[str, Any], thread_id: str | None) -> str | None:
```
bracesproul: dont accept config as an arg, just call `get_config()` in the thread
you dont need to accept thread_id either because you can get that from the config returned from thread id

bracesproul: also this function should be in a separate util file

apps/agent/agent/server.py
```diff
@@ -234,25 +234,35 @@ def graph_loaded_for_execution(config: RunnableConfig) -> bool:
...
+        logger.error("Missing source for thread %s; cannot route auth failure responses", thread_id)
+        msg = f"GitHub auth failed for thread {thread_id}: missing source"
+        raise RuntimeError(msg)
+    try:
+        github_token, new_encrypted = await save_encrypted_token_from_email(user_email, source)
+    except ValueError as exc:
+        logger.error("GitHub auth failed for thread %s: %s", thread_id, str(exc))
+        raise RuntimeError(str(exc)) from exc
+
+    config["configurable"]["github_token_encrypted"] = new_encrypted
```
bracesproul:
```suggestion
    config["metadata"]["github_token_encrypted"] = new_encrypted
```

PR #1004, opened 2026-03-04
fix: add strict tool calling and search persistence prompting  Closes [OPE-17]

aran-yogesh:


agent/prompt.py
```diff
@@ -41,6 +41,8 @@
 
 ### Task Execution
 
+> **STRICT REQUIREMENT:** Every task MUST end with a `linear_comment` or `slack_thread_reply` tool call — no exceptions. The user cannot see your text responses. The ONLY way to communicate with the user is via these tools. If you finish work without calling one of them, the user sees nothing and the task is considered failed.
```
bracesproul, line 44: instead of explicitly stating the tool names, bc it will be annoying/hard to main in the future, just say you must call a tool to leave a comment from the source you were invoked from. then, since we should already say where they were invoked from (confirm we actually do) the LLM will be smart enough to figure out what the right tool is to call

agent/prompt.py
```diff
@@ -50,13 +52,13 @@
 1. **Understand** — Read the issue/task carefully. Explore relevant files before making any changes.
 2. **Implement** — Make focused, minimal changes. Do not modify code outside the scope of the task.
 3. **Verify** — Run tests and linters to confirm correctness before submitting.
-4. **Submit** — Call `commit_and_open_pr`.
-5. **Comment** — Call `linear_comment` or `slack_thread_reply` with a summary and the PR link.
+4. **Submit** — Call `commit_and_open_pr`. You MUST call this before the comment step.
+5. **Comment** — Call `linear_comment` or `slack_thread_reply` with a summary and the PR link. **This step is mandatory. Never skip it.**
```
bracesproul, line 56: same here, dont mention specific tool names

agent/prompt.py
```diff
@@ -50,13 +52,13 @@
...
 
 For questions or status checks (no code changes needed):
 
 1. **Answer** — Gather the information needed to respond.
-2. **Comment** — Call `linear_comment` or `slack_thread_reply` with your answer. Never leave a question unanswered."""
+2. **Comment** — Call `linear_comment` or `slack_thread_reply` with your answer. **Never leave a question unanswered. Always call one of these — even if you are blocked or cannot complete the task, you must still call one to explain why.**"""
```
bracesproul, line 61: same here

agent/prompt.py
```diff
@@ -87,12 +89,20 @@
 
 ### Tool Usage Best Practices
 
+> **STRICT REQUIREMENT:** You MUST call at least one tool in every single response. There are no exceptions. If you have nothing obvious to do, run `ls` or a quick `rg --files` search. A response without a tool call is a failure.
```
bracesproul, line 92: im worried about saying `in every single response`. could this cause the model to send a comment every time it generates a message?

bracesproul: also this line doesnt make sense `If you have nothing obvious to do, run `ls` or a quick `rg --files` search. A response without a tool call is a failure.`

if it has nothing to do, it should end...?

agent/prompt.py
```diff
@@ -87,12 +89,20 @@
...
+
 - **Search:** Use `execute` to run search commands (`grep`, `find`, etc.) in the sandbox.
 - **Dependencies:** Use the correct package manager; skip if installation fails.
 - **History:** Use `git log` and `git blame` via `execute` for additional context when needed.
 - **Parallel Tool Calling:** Call multiple tools at once when they don't depend on each other.
 - **URL Content:** Use `fetch_url` to fetch URL contents. Only use for URLs the user has provided or discovered during exploration.
-- **Scripts may require dependencies:** Always ensure dependencies are installed before running a script."""
+- **Scripts may require dependencies:** Always ensure dependencies are installed before running a script.
+- **When a search returns no results, never give up — always escalate:**
+    1. Run `ls -la /workspace` to understand the full repo structure first, including hidden files and directories.
```
bracesproul, line 101: note we dont always use `/workspace`, ensure these are dynamic now

agent/prompt.py
```diff
@@ -125,6 +135,8 @@
 
 - **Persistence:** Keep working until the current task is completely resolved. Only terminate when you are certain the task is complete.
 - **Accuracy:** Never guess or make up information. Always use tools to gather accurate data about files and codebase structure.
+- **Tooling:** Every response **MUST** include at least one tool call — no exceptions. If unsure what to do next, run a lightweight `execute` search (e.g., `rg -n` or `rg --files`) or `ls` to gather context. **Never output a response without calling a tool.** Do not claim you will search; actually call a tool. A response with no tool call is always wrong.
+- **User Communication:** The user CANNOT see your text responses. The ONLY way to communicate with the user is by calling `linear_comment` (Linear tasks) or `slack_thread_reply` (Slack tasks). Always end every task — success, failure, or blocked — with one of these calls so the user knows what happened.
```
bracesproul, line 139: same abt tool names

bracesproul (changes requested):

johannes117:
closing, will re-implement

PR #1033, opened 2026-03-09
fix: handle out-of-sync push by pulling and re-committing before retry [closes OPE-5]

aran-yogesh:


bracesproul:
going to close. msgd in slack

PR #1056, opened 2026-03-12
fix: reduce token usage by improving agent and subagent efficiency

aran-yogesh:
Problem                                                                                                                                                      
A production run on langchain-ai/smith-frontend consumed 12 million tokens to fix a single UI flicker bug. Trace analysis revealed a cascade of inefficiencies rooted in how the agent dispatches and uses subagents, and how it explores codebases.                                                                        

https://smith.langchain.com/o/ebbaf2eb-769b-4505-aca2-d11de10372a4/host/deployments/a2a0ec69-99f0-4303-8d39-2791b2a66138?peek=019cd988-22bb-7060-bb7c-e0deca8576b7&peeked_trace=019cd988-22bb-7060-bb7c-e0deca8576b7

 Root Cause

  Step 1 — Bad subagent task descriptions

  The main agent dispatched 2 parallel subagents to investigate the flicker. Subagent 1 was told to search for:
  "Messages" and "Details" as string literals in tab components
  These are visible UI labels. On a large frontend codebase they match hundreds of unrelated files — menus, tooltips, tests, docs. The subagent spent 186 seconds
   grepping and found nothing useful.

  Step 2 — Subagent returned unusable output

  Subagent 1 returned 226 characters:
  The summary above covers everything. Let me know if you'd like me to dig deeper...
  It referenced "the summary above" — context that doesn't exist from the parent agent's perspective. The parent received zero actionable data.

  Step 3 — Failed greps compounded into wasted turns

  With no useful subagent output, the main agent took over and repeated the same bad search strategy. When grepping for useRunV2Streaming, it guessed the
  filename (useRunsV2.ts) instead of mapping the filesystem first. The file didn't exist at that path — so the grep returned nothing. Instead of stopping, the
  agent ran 5 variations of the same failing grep across 5 separate turns:

 ```
 grep "useRunV2Streaming" glob="**/hooks/useRunsV2.ts*"  → no results
  grep "export function useRunV2Streaming" glob="**/useRunsV2.ts"  → no results
  grep "useRunV2Streaming" glob="**/useRunsV2.ts"  → no results
  grep "useRunV2Streaming" path="/workspace/.../src"  → no results
  grep "export function useRunV2Streaming" path="/workspace/.../src"  → finally found
```

  Each failed turn re-sent the full growing conversation history. The agent also directly grepped for "Messages" as a string literal — the same flood strategy as
   subagent 1 — returning hundreds of useless matches it then had to process.

  Step 4 — Main agent recovered with 94 extra turns

  With no useful subagent output and failed searches, the main agent spent 94 turns on exploration. The same files were read 3-6 times each without modification
  between reads. git diff --stat was run twice with no changes between the calls.

  Step 5 — Context compounding

  Every turn re-sends the entire conversation history. 94 extra turns at an average 50k input tokens = ~4.7M extra tokens from the recovery phase alone. Combined
   with the subagents, the run hit 12M tokens.

  Step 6 — Secondary issues found in follow-up traces

  - Subagents dispatched for file retrieval instead of using read_file directly
  - Subagent returned prose summary when asked for exact verbatim code
  - 3 parallel subagents assigned overlapping file scopes → 3× duplicate output
  - write_todos called on every single step, adding permanent context overhead
  - README and unrelated package.json read and held in context permanently

grep fix:
https://dev.smith.langchain.com/public/029c273f-8489-40b7-b3be-be2f1b8f8d17/r

### DIFF
--- a/agent/prompt.py
+++ b/agent/prompt.py
@@ -50,7 +50,7 @@ If you make changes, communicate updates in the source channel:
 
 For tasks that require code changes, follow this order:
 
-1. **Understand** — Read the issue/task carefully. Explore relevant files before making any changes.
+1. **Understand** — Read the issue/task carefully. Before reading any file or grepping anything, map the repo structure first by running `find . -type f | grep -v node_modules | grep -v dist | grep -v ".git" | head -150` from `{working_dir}`. This gives you the complete real file tree so you never have to guess filenames. Subagent output is findings, not a filesystem map — always build your own map first.
 2. **Implement** — Make focused, minimal changes. Do not modify code outside the scope of the task.
 3. **Verify** — Run tests and linters to confirm correctness before submitting.
 4. **Submit** — Call `commit_and_open_pr` to push changes to the existing PR branch.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: agent/prompt.py [hunk 1]
```diff
@@ -50,7 +50,7 @@
 
 For tasks that require code changes, follow this order:
 
-1. **Understand** — Read the issue/task carefully. Explore relevant files before making any changes.
+1. **Understand** — Read the issue/task carefully. Before reading any file or grepping anything, map the repo structure first by running `find . -type f | grep -v node_modules | grep -v dist | grep -v ".git" | head -150` from `{working_dir}`. This gives you the complete real file tree so you never have to guess filenames. Subagent output is findings, not a filesystem map — always build your own map first.
```
bracesproul: update suggested command to:
```
git ls-files | tree --fromfile -L 3
```
but prompt it to add more or less levels, suggest 3 as a good starting point

johannes117:
I think this is overprompting. deepagents should handle this

PR #1070, opened 2026-03-17
feat: authenticate git operations via sandbox proxy instead of credential files [closes: OPE-20]

aran-yogesh:
## Deployment Checklist                                                                                                                      
                                                                                                                                               
  - [ ] Set `DEFAULT_SANDBOX_TEMPLATE_NAME=open-swe-new` in the revision env vars                                                              
  - [ ] Deploy the revision (first run will auto-create the `open-swe-new` template if missing)                                                
  - [ ] Update the sandbox pool to use the new `open-swe-new` template


agent/integrations/langsmith.py
```diff
@@ -43,18 +48,68 @@ def _get_sandbox_template_config() -> tuple[str | None, str | None]:
     return template_name, template_image
 
 
+def _configure_github_proxy(sandbox_name: str, github_token: str, api_key: str) -> None:
```
bracesproul: sandboxes support service keys now, so for our internal use can you cut a service key for the user, but also ensure you fallback to api key if they dont have the service key secret set

agent/integrations/langsmith.py
```diff
@@ -128,7 +187,7 @@ def delete(
 
 
 # Default template configuration
-DEFAULT_TEMPLATE_NAME = "open-swe"
+DEFAULT_TEMPLATE_NAME = "open-swe-new"
```
bracesproul, line 168: why is this changing?

agent/server.py
```diff
@@ -197,6 +171,21 @@ async def _clone_or_pull_repo_in_sandbox(  # noqa: PLR0915
     return repo_dir
 
 
+async def _create_sandbox_with_proxy(github_token: str | None) -> SandboxBackendProtocol:
```
bracesproul: why is github token str or none? and why are we accepting it as an arg if we get the installation token? can we just drop this arg and only use the installation token?

agent/server.py
```diff
@@ -342,19 +328,19 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
...
-            # Reset sandbox_id and create a new sandbox
+            # Reset sandbox_id and create a new sandbox with proxy auth configured
             await client.threads.update(
                 thread_id=thread_id,
                 metadata={"sandbox_id": SANDBOX_CREATING},
             )
 
             try:
-                sandbox_backend = await asyncio.to_thread(create_sandbox)
+                sandbox_backend = await _create_sandbox_with_proxy(github_token)
```
bracesproul: if you get rid of create_sandbox and instead use this new method, people using non-langsmith sandboxes will break. you should continue using the create_sandbox method and if they have the sandbox type set to langsmith, create the langsmith sandbox using the proxy server

agent/integrations/langsmith.py
aran-yogesh: this is because proxy server wasnt working with old template mukil asked to create a new template
Its also in the deployment check list

agent/integrations/langsmith.py
aran-yogesh: Talked to Mukil, going with API key for now

agent/integrations/langsmith.py
bracesproul: will the name changing cause any problems?

agent/integrations/langsmith.py
bracesproul: nit dont pass api key via args, just read it when you need it

agent/integrations/langsmith.py
aran-yogesh: No it wont cause any problems Already test it locally 
But we should be updating the sandbox pool as well

PR #1072, opened 2026-03-17
feat: add GitHub image parsing support for webhook events [closes: OPE-6]

aran-yogesh:
https://dev.smith.langchain.com/public/281ef135-8aaa-435c-acdc-9d1b41f9853b/r


agent/utils/multimodal.py
bracesproul, line 1: why does the `fetch_image_block` method have a check for if its a github url?

agent/webapp.py
```diff
@@ -1315,9 +1315,20 @@ async def process_github_pr_comment(payload: dict[str, Any], event_type: str) ->
         return
 
     prompt = build_pr_prompt(comments, pr_url)
+    image_urls = dedupe_urls(
```
bracesproul: this logic is duplicated below. please refactor to a util function so we dont need to duplicate logic

bracesproul (changes requested):

agent/utils/multimodal.py
aran-yogesh: GitHub doesn’t send the image directly — it redirects to another link (CDN). So we need `follow_redirects=True ` to actually fetch the image instead of getting a redirect response.

Image

PR #1073, opened 2026-03-18
feat: add Slack image support with url_private auth and content-type validation [closes: OPE-23]

aran-yogesh:
## Tasks                                                                                                                     
  - [ ] Add `files:read` scope to Slack bot token scopes at api.slack.com/apps → OAuth & Permissions → Bot Token Scopes        
  - [ ] Reinstall app to workspace after adding the scope to regenerate the bot token 

https://dev.smith.langchain.com/public/bd2e3332-c2da-46c6-9ab0-b1880abadd89/r

https://dev.smith.langchain.com/public/af1aa355-92da-4316-abab-1830b3f2e364/r

### DIFF
--- a/agent/utils/multimodal.py
+++ b/agent/utils/multimodal.py
@@ -46,7 +46,7 @@ async def fetch_image_block(
         logger.debug("Fetching image from %s", image_url)
         headers = None
         host = (urlparse(image_url).hostname or "").lower()
-        if host.endswith("uploads.linear.app"):
+        if host == "uploads.linear.app" or host.endswith(".uploads.linear.app"):
             linear_api_key = os.environ.get("LINEAR_API_KEY", "")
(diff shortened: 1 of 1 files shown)


agent/utils/multimodal.py
```diff
@@ -44,7 +45,8 @@ async def fetch_image_block(
     try:
         logger.debug("Fetching image from %s", image_url)
         headers = None
-        if "uploads.linear.app" in image_url:
+        host = (urlparse(image_url).hostname or "").lower()
+        if host == "uploads.linear.app" or host.endswith(".uploads.linear.app"):
```
bracesproul, line 49: can prob simplify this by only checking:
```suggestion
        if host.endswith("uploads.linear.app"):
```
same for all below

PR #1142, opened 2026-03-27
fix: check for running runs instead of thread status in is_thread_active

vethan:
`is_thread_active` previously checked thread status, which can get stuck as  "busy" if a run crashes (e.g. sandbox goes down mid-run). Now it checks for actual running runs via `runs.list(status="running")`.  This fixes issues where Slack messages were queued but never processed after a sandbox failure


johannes117:
could you update this to treat pending runs as active too? otherwise a webhook can race a pending run and create a second run instead of queueing.

PR #1143, opened 2026-03-27
fix: recover from stale sandbox creation sentinel instead of deadlocking     

vethan:
Fix deadlock when `__creating__` sandbox sentinel is left over from a previous crashed/interrupted run. With max_workers=1, the single worker would block for 180s polling for a sandbox that no other run could create.   `_wait_for_sandbox_id` now returns `None` on timeout (or if sentinel is reset by a failed run), resets the stale sentinel, and lets the caller fall through to create a new sandbox.


agent/server.py
```diff
@@ -235,25 +235,34 @@ async def _recreate_sandbox(
...
 
     Polls thread metadata until sandbox_id is set to a real value
-    (not the creating sentinel).
-
-    Raises:
-        TimeoutError: If sandbox creation takes too long
+    (not the creating sentinel).  If no other run is actively creating
+    the sandbox (stale sentinel), resets the sentinel and returns None
+    so the caller can create it.
     """
```
johannes117, line 245: can we re-read the metadata right before clearing this? otherwise a slow creator could write the real sandbox id just before this update and we would overwrite it.

PR #1144, opened 2026-03-27
fix: add Python version and git to Modal sandbox image

hung-ngm:
## Summary

- Pin Modal sandbox image to Python 3.12 — without this, `debian_slim()` inherits the local Python version which may be unsupported (e.g. 3.13 fails with `InvalidError: Unsupported Python version`)
- Install git via `apt_install("git")` — required by `_clone_or_pull_repo_in_sandbox` which runs `git clone` inside the sandbox

## Context

The current `modal.Sandbox.create(app=app)` uses the default `debian_slim()` image which:
1. Doesn't include git (causes `bash: line 1: git: command not found`)
2. Defaults to the local Python version which may not be supported by Modal's Image builder

All other sandbox providers have git pre-installed (LangSmith uses `python:3`, Daytona uses a pre-built snapshot, Runloop has devboxes with git, local uses host git). Modal needs it explicitly due to its minimal base images.

## Test plan

- [ ] Set `SANDBOX_TYPE=modal` and `MODAL_APP_NAME=open-swe` in `.env`
- [ ] Trigger `@openswe` from a GitHub issue comment
- [ ] Verify sandbox creates successfully (no Python version error)
- [ ] Verify repo clones into sandbox (no git not found error)
- [ ] Verify agent responds to the issue


ArshVermaGit (commented):
Nice catch on the environment mismatch here — pinning the Modal image to Python 3.12 is a smart, defensive move that prevents subtle failures when local Python versions drift ahead of what Modal supports. Explicitly installing git via `apt_install("git")` also makes the sandbox behavior consistent with the other providers and avoids the confusing `git: command not found` failure during repo setup. I like that this keeps the change minimal while solving two real-world reliability issues that would otherwise be hard to debug. The test plan is practical and clearly proves end-to-end functionality from sandbox creation through repo cloning and agent response. Overall, a clean and pragmatic fix that improves portability and developer confidence. 👍

PR #1146, opened 2026-03-27
chore: add PR review evals with LLM-as-judge scoring [closes: OPE-28]

aran-yogesh:


agent/tools/github_review.py
```diff
@@ -250,6 +261,13 @@ def submit_pr_review(
     if not repo_config:
         return {"success": False, "error": "No repo config found"}
 
+    config = get_config()
+    if config.get("configurable", {}).get("eval_mode"):
```
bracesproul: could you make this a util `is_eval_mode` and inside that call `get_config`

this way your if statement can just be `if is_eval_mode():`

also, i would refactor it to be `mode: "eval"` this way we can easily add more modes later

make sure you're checking the value is `eval` too after this change

johannes117:
closing, will re-implement

PR #1154, opened 2026-03-31
fix: clarify before acting when context is insufficient

aran-yogesh:


ArshVermaGit (commented):
This PR improves the agent’s decision-making flow by encouraging clarification when the available context isn’t sufficient, which should help prevent unnecessary or incorrect changes. The added guidance in `prompt.py` strikes a nice balance — it nudges the agent to pause and ask questions when requirements are ambiguous, while still emphasizing forward progress when the intent is clear. I especially like the focus on being deliberate before modifying code, as that helps reduce noisy refactors and improves overall reliability in shared components. Even though the change is small, it meaningfully improves how the agent approaches tasks, making its behavior feel more thoughtful and aligned with real-world developer expectations.

PR #1156, opened 2026-04-01
feat: add ACP connector for remote Open SWE threads

johannes117:
## Summary
- add a dual-auth ACP connector flow so desktop clients can authenticate either with a LangSmith API key or with deployment-managed Open SWE ACP API keys
- exchange LangSmith-authenticated ACP sessions into the caller's GitHub identity, stamp ACP thread ownership metadata, and enforce `session/list` and `session/load` against that authenticated subject
- support self-hosted ACP API keys that can either map to user context or fall back to shared GitHub App execution, and document the new setup with focused ACP/auth tests

## Test plan
- [x] `uv run pytest tests/test_acp_connector.py tests/test_auth_sources.py -q`
- [x] `uv run ruff check agent/acp_connector.py agent/utils/auth.py tests/test_acp_connector.py tests/test_auth_sources.py`

Made with [Cursor](https://cursor.com)


ArshVermaGit (approved):
This PR introduces an ACP connector to support remote Open SWE threads, and overall it feels like a thoughtful addition that improves how authentication and session ownership are handled across environments. The dual-auth flow is particularly useful, giving flexibility for desktop clients to authenticate either through LangSmith or deployment-managed ACP keys, while still maintaining clear ownership metadata tied to the caller’s GitHub identity. I also like that the implementation considers both self-hosted and shared execution scenarios, which makes the feature more adaptable across setups. The included tests around auth sources and connector behavior add confidence that the flow works as expected and won’t regress easily. Overall, this looks like a well-scoped feature that meaningfully improves interoperability and makes the Open SWE workflow more robust and practical to use.

PR #1159, opened 2026-04-02
feat: stop auto-cloning and let agent manage repo setup [closes OPE-21]

aran-yogesh:
https://dev.smith.langchain.com/public/2f0ba864-c649-4b1d-b3df-5936ee182052/r


ArshVermaGit (commented):
Interesting change — delegating repo setup to the agent instead of auto-cloning feels like a solid step toward making the workflow more flexible and easier to run in different environments (local, sandboxed, or remote). It should also reduce some hidden side effects during startup and give the agent clearer control over repo lifecycle.
The addition of utilities like get_branch_name and list_repos makes sense in this context, since the agent now needs more explicit visibility into repository state. I also like that this keeps responsibilities more modular rather than embedding repo assumptions directly into the runtime.
One thing worth double-checking is how this impacts first-run UX and failure modes — for example:
what happens if the repo isn’t present locally?
do we provide clear guidance or logs when setup is required?
are existing workflows relying on auto-clone going to break silently?
If those cases are handled (or documented), this looks like a clean architectural improvement that should make the system more predictable and easier to extend.

JiwaniZakir (commented):
The shell command in step 4 of `REPO_SETUP_SECTION` in `prompt.py` has a subtle operator precedence bug:

```bash
cd /workspace/<name> && git fetch origin && git checkout <branch_name> 2>/dev/null || git checkout -b <branch_name>
```

Because `&&` and `||` are left-associative with equal precedence, if `cd` or `git fetch` fails, the shell falls through to `git checkout -b <branch_name>` executing in whatever the current directory is — not `/workspace/<name>`. The intent is clearly to only create the branch if the checkout fails, so it should be grouped: `cd /workspace/<name> && git fetch origin && (git checkout <branch_name> 2>/dev/null || git checkout -b <branch_name>)`.

Additionally, moving `AGENTS.md` ingestion from a guaranteed server-side read (`read_agents_md_in_sandbox` + injecting into `construct_system_prompt`) to an agent-managed step is a behavioral regression worth calling out. Previously, AGENTS.md content was always present in the system prompt before the agent took any action. Now it's advisory — the agent might skip step 5, or complete steps out of order, and silently proceed without repo-specific rules. A stronger enforcement mechanism (e.g., a tool hook or middleware that verifies AGENTS.md was read) would preserve the original reliability guarantee.

agent/prompt.py
```diff
@@ -26,16 +25,41 @@
...
+Before starting any task, you must set up the repository in your sandbox. Follow these steps in order:
+
+1. **Find the repo** — Call `list_repos` to get the list of available repositories. Match the repo to your task context (e.g. the Linear team/project or issue description). If the repo is not in the common list, call `list_repos(org="<org_name>")` to search a GitHub org. If you are still unsure which repo to use, ask the user for confirmation before proceeding.
+
+2. **Clone the repo** — Clone it into `/workspace`:
+   ```
+   git clone https://github.com/<owner>/<name>.git /workspace/<name>
+   ```
+
+3. **Get your branch** — Call `get_branch_name` to get the branch name for this thread. Never hard-code the branch name.
```
bracesproul: what does hard coding the branch name mean?

agent/prompt.py
```diff
@@ -26,16 +25,41 @@
...
+### Repository Setup
+
+Before starting any task, you must set up the repository in your sandbox. Follow these steps in order:
+
+1. **Find the repo** — Call `list_repos` to get the list of available repositories. Match the repo to your task context (e.g. the Linear team/project or issue description). If the repo is not in the common list, call `list_repos(org="<org_name>")` to search a GitHub org. If you are still unsure which repo to use, ask the user for confirmation before proceeding.
+
+2. **Clone the repo** — Clone it into `/workspace`:
+   ```
+   git clone https://github.com/<owner>/<name>.git /workspace/<name>
+   ```
```
bracesproul: the agent knows how to clone, dont need this
```suggestion
```
also `/workspace` is unique to LangSmith sandboxes, not all providers use that as the root. there should be a util func you can use to get the root

agent/prompt.py
bracesproul: should we give it a tool to get branch name, or should it just be able to run git commands to get the name? Whats the reasoning for the tool?

agent/prompt.py
```diff
@@ -26,16 +25,41 @@
...
+
+3. **Get your branch** — Call `get_branch_name` to get the branch name for this thread. Never hard-code the branch name.
+
+4. **Checkout your branch** — Always fetch and checkout your branch before making any changes:
+   ```
+   cd /workspace/<name> && git fetch origin && (git checkout <branch_name> 2>/dev/null || git checkout -b <branch_name>)
+   ```
```
bracesproul: same here, it knows how to checkout branches
```suggestion
```

agent/prompt.py
```diff
@@ -26,16 +25,41 @@
...
+
+You MUST complete ALL of these steps before doing any other work. The sandbox starts clean — no repo is pre-cloned."""
 
 
 FILE_MANAGEMENT_SECTION = """---
 
 ### File & Code Management
 
-- **Repository location:** `{working_dir}`
+- **Repository location:** `/workspace/<repo_name>` (clone the repo here first — see Repository Setup)
```
bracesproul: yeah you deleted the {working_dir} code here. dont do that. as stated above `workspace` is unique to langsmith, not all providers

agent/prompt.py
```diff
@@ -271,31 +302,14 @@
...
 )
 
 
 def construct_system_prompt(
-    working_dir: str,
     linear_project_id: str = "",
     linear_issue_number: str = "",
-    agents_md: str = "",
 ) -> str:
-    agents_md_section = ""
```
bracesproul, line 287: ooc why not hard code the agents.md anymore? i see you added some prompting to always read instead of inject here, but whats the reasoning behind that?

agent/server.py
```diff
@@ -290,26 +181,14 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
...
         logger.info("Using cached sandbox backend for thread %s", thread_id)
-        metadata = get_config().get("metadata", {})
-        repo_dir = metadata.get("repo_dir")
-
-        if repo_owner and repo_name:
-            logger.info("Pulling latest changes for repo %s/%s", repo_owner, repo_name)
-            try:
-                repo_dir = await _clone_or_pull_repo_in_sandbox(
-                    sandbox_backend, repo_owner, repo_name
-                )
```
bracesproul, line 311: can we make this a util? `check_or_recreate_sandbox`

agent/prompt.py
aran-yogesh: i meant "don't make up your own branch name, use the tool." i reworded it

aran-yogesh: The branch name follows `open-swe/{thread_id} ` the thread ID comes from metadata, not git. Agent has no way to know its own thread ID without this tool.

agent/prompt.py
aran-yogesh:
```
 Before (on main):                                    
  1. Server auto-clones the repo into the sandbox      
  2. Server reads AGENTS.md from the cloned repo       
  3. Server injects it into the system prompt via agents_md param                                
  4. Agent starts with AGENTS.md already in context    
```

```
Now (this PR):                                       
1. Server starts the agent (sandbox is clean, no repo yet)                                                
2. Agent clones the repo itself                      
3. Agent reads AGENTS.md itself (per step 5 in the prompt)     
```                                      
                                                       
Since the repo isn't cloned when construct_system_prompt is called, there's no AGENTS.md to read yet. That's why i removed the param — it would always be empty.
We could maybe add a middleware that reads the AGENTS.md from the sandbox after the first clone and then keeps injecting it into the system prompt for the next turns. What do you think?

agent/tools/list_repos.py
```diff
@@ -0,0 +1,69 @@
...
+        for entry in entries:
+            key = (entry["owner"], entry["name"])
+            if key not in seen:
+                seen.add(key)
+                repos.append({"owner": entry["owner"], "name": entry["name"]})
+
+    return repos
+
+
+async def list_repos(org: str | None = None) -> dict[str, Any]:
```
bracesproul: should support pagination b/c some orgs (like ours) have 100's of repos


```suggestion
async def list_repos(organization_name: str | None = None) -> dict[str, Any]:
```

also itll only ever be for a single org, so return value should just be a list of strings of the repo names & any additional pagination info

but big thing here is to support more args for pagination/filtering/limits

agent/tools/list_repos.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Returns common repos from the configured repo map.
+    Pass org to also search that GitHub org via the API.
+    If unsure which repo to use, ask the user for confirmation.
+    """
+    common_repos = _get_common_repos()
+    result: dict[str, Any] = {
+        "common_repos": common_repos,
+        "default_org": DEFAULT_GITHUB_ORG,
+    }
```
bracesproul: dont do this, should only fetch from the github api

agent/prompt.py
```diff
@@ -26,16 +26,35 @@
...
+
+1. **Find the repo** — Call `list_repos` to get the list of available repositories. Match the repo to your task context (e.g. the Linear team/project or issue description). If the repo is not in the common list, call `list_repos(org="<org_name>")` to search a GitHub org. If you are still unsure which repo to use, ask the user for confirmation before proceeding.
+
+2. **Clone the repo** — Clone it into `{working_dir}`.
+
+3. **Get your branch** — Always call the `get_branch_name` tool to get the branch name for this thread.
+
+4. **Checkout your branch** — Always fetch and checkout your branch before making any changes.
+
+5. **Read and follow AGENTS.md** — After cloning, check if `AGENTS.md` exists at the repository root (`{working_dir}/<name>/AGENTS.md`). If it exists, you MUST read it immediately and treat its contents as **mandatory rules** for all work in that repository. AGENTS.md contains project-specific conventions, coding standards, and constraints that override your default behavior. Violating AGENTS.md rules is equivalent to violating the system prompt. If AGENTS.md does not exist, skip this step.
```
bracesproul:
```suggestion
5. **Read and follow AGENTS.md** — After cloning, check if `AGENTS.md` exists at the repository root (`{working_dir}/<repo>/AGENTS.md`). If it exists, you MUST read it immediately and treat its contents as **mandatory rules** for all work in that repository. AGENTS.md contains project-specific conventions, coding standards, and constraints that override your default behavior. Violating AGENTS.md rules is equivalent to violating the system prompt. If AGENTS.md does not exist, skip this step.
```

agent/tools/list_repos.py
```diff
@@ -0,0 +1,55 @@
...
+async def list_repos(
+    organization_name: str,
+    page: int = 1,
+    per_page: int = 100,
+    sort: str = "updated",
+    name_filter: str | None = None,
+) -> dict[str, Any]:
+    """List GitHub repositories for an organization via the GitHub API.
+
+    Note: This uses the /orgs/{name}/repos endpoint, which only works for
```
bracesproul: did you confirm this? id be surprised if github doesnt offer an api for listing repos for personal orgs...

my guess is its called `orgs`, but still accepts regular user account names

bracesproul (approved):
lgtm but plz confirm the org stuff, and if it works for users update the docstring

PR #1160, opened 2026-04-03
chore: bump some langchain deps

ccurme:
`langchain-anthropic` -> 1.4.0 (picks up prompt caching improvements)

Also bumps deepagents to 0.5.0a3 (let me know if you'd prefer not to install pre-release)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.14'",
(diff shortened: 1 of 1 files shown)


ArshVermaGit (commented):
Looks good overall 👍 — dependency-only changes are usually low risk, and the upgrade to langchain-anthropic 1.4.0 is especially nice since it brings prompt caching improvements, which should help performance and reduce token costs in longer-running threads.
The only thing worth double-checking is the bump to deepagents 0.5.0a3, since it’s an alpha release. Pre-release versions can sometimes introduce breaking changes or unstable APIs. If this was intentional for a specific feature or fix, I’m fine with it — but if not, we may want to pin to the latest stable version to avoid unexpected behavior in production.
If tests are passing and there are no regressions in agent workflows, I’m comfortable approving 🚀

johannes117:
superceded by main

PR #1165, opened 2026-04-05
fix: reset stale sandbox creation sentinel after restart

### LINKED ISSUE #1116: Bug: Stale SANDBOX_CREATING sentinel causes indefinite timeout after server restart
## Description

When a server restarts while a sandbox is being created, the `SANDBOX_CREATING` sentinel value (`__creating__`) remains in the thread metadata but the in-memory `SANDBOX_BACKENDS` cache is lost. On the next request for that thread, the server enters `_wait_for_sandbox_id()` and waits up to 3 minutes before timing out — even though no worker is actively creating the sandbox.

## Steps to Reproduce

1. Start the server
2. Trigger a webhook that begins sandbox creation (metadata gets set to `__creating__`)
3. Restart the server before sandbox creation completes
4. Trigger another webhook for the same thread
5. Server waits 3 minutes then raises `TimeoutError`

## Expected Behavior

The server should detect the stale sentinel (sandbox_id == `SANDBOX_CREATING` but no in-memory backend cached) and reset the metadata to `None`, allowing fresh sandbox creation.

## Proposed Fix

In `get_agent()`, when `sandbox_id == SANDBOX_CREATING` and `not sandbox_backend` (no in-memory cache), reset the metadata instead of waiting:

```python
if sandbox_id == SANDBOX_CREATING and not sandbox_backend:
    logger.warning(
        "Found stale SANDBOX_CREATING for thread %s with no cached backend, resetting",
        thread_id,
    )
    await client.threads.update(thread_id=thread_id, metadata={"sandbox_id": None})
    sandbox_id = None
```

This preserves the existing wait behavior for the normal multi-worker case (where another worker IS actively creating the sandbox and has the backend in its memory).

I have a PR ready with the fix and tests.

Gujiassh:
## Summary
- reset stale `SANDBOX_CREATING` thread metadata when a worker restarts and no cached sandbox backend exists anymore
- preserve the normal wait path only for the true multi-worker case where another worker is still actively creating the sandbox
- add a focused regression test covering the stale-sentinel reset path in `get_agent()`

Fixes #1116.

## Testing
- `python3 -m pytest tests/test_stale_sandbox_creating.py`
- `python3 -m py_compile agent/server.py tests/test_stale_sandbox_creating.py`

### DIFF
--- a/agent/server.py
+++ b/agent/server.py
@@ -284,6 +284,13 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
     sandbox_id = await get_sandbox_id_from_metadata(thread_id)
 
     if sandbox_id == SANDBOX_CREATING and not sandbox_backend:
+        logger.warning(
+            "Found stale SANDBOX_CREATING for thread %s with no cached backend, resetting",
+            thread_id,
+        )
+        await client.threads.update(thread_id=thread_id, metadata={"sandbox_id": None})
+        sandbox_id = None
+    elif sandbox_id == SANDBOX_CREATING:
         logger.info("Sandbox creation in progress, waiting...")
         sandbox_id = await _wait_for_sandbox_id(thread_id)
 
--- a/tests/test_stale_sandbox_creating.py
+++ b/tests/test_stale_sandbox_creating.py
@@ -1,0 +1,215 @@
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+MODULE_PATH = REPO_ROOT / "agent" / "server.py"
+
+
+def load_server_module():
+    fake_client = types.SimpleNamespace(threads=types.SimpleNamespace(update=None))
+
+    agent_pkg = types.ModuleType("agent")
+    agent_pkg.__path__ = [str(REPO_ROOT / "agent")]
+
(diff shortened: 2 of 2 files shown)


ArshVermaGit (commented):
This is a very solid reliability fix that addresses a subtle but impactful edge case in the sandbox lifecycle. Resetting stale `SANDBOX_CREATING` metadata when no backend actually exists prevents workers from getting stuck in a false waiting state, while still preserving the intended coordination behavior for genuine multi-worker scenarios. I especially like that the change is narrowly scoped and supported by a focused regression test that explicitly validates the stale-sentinel recovery path in `get_agent()`. The inclusion of compile checks alongside pytest adds extra confidence that the change is safe and clean. Overall, this is a thoughtful and pragmatic fix that strengthens resilience in distributed worker environments without introducing unnecessary complexity. Great catch and well executed.

Gujiassh:
Revalidated on the current branch tip:\n\n- python3 -m pytest tests/test_stale_sandbox_creating.py\n- Result: 1 passed\n\nThe fix still addresses the stale restart case correctly: if the thread metadata is stuck on SANDBOX_CREATING but there is no cached backend after restart, the code now clears the stale sentinel instead of waiting forever.

Gujiassh:
Revalidated on 2026-04-11:\n\n- python3 -m pytest tests/test_stale_sandbox_creating.py\n- Result: 1 passed\n\nThe stale SANDBOX_CREATING sentinel reset path remains correct on the current branch tip.

PR #1178, opened 2026-04-09
fix: refresh GitHub proxy token on sandbox reuse to prevent git auth failures

aran-yogesh:
Follow-up messages in the same Slack thread reuse the existing sandbox. The sandbox proxy was configured with a GitHub installation token only at creation time — never refreshed on reuse. When the token expired (1 hour), all git operations failed:

Git pull failed with exit code 1: fatal: could not read Username for 'https://github.com': No such device or address

Screenshot 2026-04-09 at 12 10 20 PM

Screenshot 2026-04-09 at 12 14 45 PM

Not SSO related — the error is missing credentials in the proxy, not a GitHub-side rejection.

### DIFF
--- a/agent/server.py
+++ b/agent/server.py
@@ -354,7 +354,6 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
         try:
             sandbox_backend = await asyncio.to_thread(create_sandbox, sandbox_id)
             logger.info("Connected to existing sandbox %s", sandbox_id)
-            await _refresh_github_proxy_if_needed(sandbox_backend)
         except Exception:
             logger.warning("Failed to connect to existing sandbox %s, creating new one", sandbox_id)
             # Reset sandbox_id and create a new sandbox with proxy auth configured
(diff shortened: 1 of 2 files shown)


agent/server.py
```diff
@@ -192,6 +192,24 @@ async def _create_sandbox_with_proxy() -> SandboxBackendProtocol:
     return sandbox_backend
 
 
+async def _refresh_github_proxy_if_needed(
```
bracesproul: this won't refresh when needed. itll refresh every time. misleading name

agent/server.py
```diff
@@ -351,6 +370,8 @@ async def get_agent(config: RunnableConfig) -> Pregel:  # noqa: PLR0915
                 await client.threads.update(thread_id=thread_id, metadata={"sandbox_id": None})
                 raise
 
+        await _refresh_github_proxy_if_needed(sandbox_backend)
```
bracesproul: why are you calling this twice? once on line 304 and again here?

bracesproul (changes requested):

agent/server.py
aran-yogesh: They're in different branches of the if/elif/else, so only one runs per request
not both

bracesproul (approved):

PR #1189, opened 2026-04-15
Security: Server-Side Request Forgery (SSRF) risk in image fetching pipeline

tuanaiseo:
## Problem

Image URLs from queued messages are fetched directly without network safety validation. In `check_message_queue_before_model`, untrusted `image_urls` are passed to `fetch_image_block`, which performs `client.get(..., follow_redirects=True)` on arbitrary URLs. An attacker could supply internal/metadata endpoints (e.g., cloud instance metadata) or redirect chains to access internal resources.

**Severity**: `high`
**File**: `agent/utils/multimodal.py`

## Solution

Apply the same URL safety controls used in `agent/tools/http_request.py` before any fetch: enforce http/https only, resolve and block private/loopback/link-local/reserved IPs, and validate every redirect target before following. Consider allowlisting trusted hosts for image ingestion.

## Changes

- `agent/utils/multimodal.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/agent/utils/multimodal.py
+++ b/agent/utils/multimodal.py
@@ -3,12 +3,14 @@
 from __future__ import annotations
 
 import base64
+import ipaddress
 import logging
 import mimetypes
 import os
 import re
+import socket
 from typing import Any
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
 
 import httpx
 from langchain_core.messages.content import create_image_block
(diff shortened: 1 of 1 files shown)


jkennedyvz:
Thanks for the report, but this isn't exploitable in practice. Image URLs are extracted via a regex that requires image file extensions (.png, .jpg, etc.), responses are validated against an image content-type allowlist, and the results are base64-encoded  into LLM image blocks with no path back to the caller. There's no feasible way to use this for metadata access or data exfiltration.

Closing as not applicable.

PR #1200, opened 2026-04-16
feat: give agent ability to read cross-posted Slack message links [closes OPE-37]

aran-yogesh:
## Summary                                                                      
  - Add `read_slack_message` agent tool to resolve cross-posted Slack message   
  links on demand                                                                 
  - Auto-resolve Slack links in webhook preprocessing so the agent gets message   
  content upfront    

## Trace

https://dev.smith.langchain.com/public/0c688e43-4f82-4165-a126-c53b64a3493a/r


agent/webapp.py
```diff
@@ -758,16 +760,62 @@ async def process_slack_mention(event_data: dict[str, Any], repo_config: dict[st
     )
     trigger_user = user_name or (f"<@{user_id}>" if user_id else "Unknown user")
 
+    # Auto-resolve cross-posted Slack message links in context
+    resolved_links_section = ""
+    image_urls_from_links: list[str] = []
+    all_context_text = " ".join(msg.get("text", "") for msg in context_messages)
+    slack_links = extract_slack_message_urls(all_context_text)
+    seen_urls: set[str] = set()
+    if slack_links:
```
bracesproul: all this logic should be abstracted out. dont want to include this much business logic within an already very large function

agent/webapp.py
```diff
@@ -758,16 +760,62 @@ async def process_slack_mention(event_data: dict[str, Any], repo_config: dict[st
...
     prompt = (
         "You were mentioned in Slack.\n\n"
         f"## Repository\n{repo_config.get('owner')}/{repo_config.get('name')}\n\n"
         f"## Triggered by\n{trigger_user}\n\n"
         f"## Slack Thread\n- Channel: {channel_id}\n- Thread TS: {thread_ts}\n"
         f"- Context starts at: {context_source}\n\n"
         f"## Conversation Context\n{context_text}\n\n"
-        f"## Latest Mention Request\n{clean_text}\n\n"
+        f"## Latest Mention Request\n{clean_text}"
+        f"{resolved_links_section}\n\n"
```
bracesproul: shouldn't actually include this section unless there's contents (there likely won't be most of the time)

agent/tools/read_slack_message.py
```diff
@@ -0,0 +1,77 @@
...
+        return None, ""
+    user_id = message.get("user", "")
+    author = user_id
+    if user_id:
+        names = await get_slack_user_names([user_id])
+        author = names.get(user_id, user_id)
+    return message, author
+
+
+def read_slack_message(url: str) -> dict[str, Any]:
```
bracesproul: this should be `slack_read_thread_messages`

and it should accept args for getting any message not just messages from urls:
- message_ts
- channel_ts
the agent will be smart enough to convert the url to this format, and then it can also be used in other instances to get messages in slack

PR #1225, opened 2026-04-28
fix(http_request): pin DNS resolution to defeat SSRF rebinding bypass

### LINKED ISSUE #1186: SSRF protection bypass via DNS rebinding in http_request.py
## Bug Description

The SSRF filter in `_is_url_safe()` resolves the target hostname via `socket.getaddrinfo()` to block private/internal addresses. However, `requests.request()` performs a second, independent DNS resolution when opening the TCP connection. An attacker controlling a short-TTL DNS record can pass the first check (resolving to a public IP) and then switch the record to an internal address before the actual connection is made.

## Location

`agent/tools/http_request.py:23–37` — DNS resolution for validation  
`agent/tools/http_request.py:69` — TCP connection (second DNS resolution)

## Reproduction

1. Register a domain with a DNS record pointing to a public IP and TTL ≤ 1 second.
2. Pass that URL to the `http_request` tool.
3. `_is_url_safe` resolves to the public IP — check passes.
4. Before `requests.request` opens the socket, update the DNS record to `192.168.1.1` (or `169.254.169.254`).
5. The actual TCP connection goes to the internal address, bypassing the SSRF filter.

## Impact

Agents running user-supplied URLs can be redirected to internal network services (AWS metadata endpoint, Redis, internal APIs) despite the SSRF protection. The bypass requires controlling DNS infrastructure but is reliable with sub-second TTLs.

## Suggested Fix

Resolve the hostname once, pin the resolved IP, and pass it directly to the transport layer so no second DNS lookup occurs. One approach is a custom `requests` adapter that overrides the socket connect to use the pre-validated IP rather than re-resolving the hostname.

---
Found via automated codebase analysis. Happy to submit a PR if this is confirmed.

voidborne-d:
Closes #1186.

## Summary

The SSRF guard in `agent/tools/http_request.py` resolves the hostname twice — once for validation, and again when `requests.request` opens the TCP connection. An attacker controlling a short-TTL DNS record can answer the first query with a public IP (so `_is_url_safe` returns `True`) and the second with an internal address (e.g. `169.254.169.254`, `10.0.0.0/8`), bypassing the filter.

This PR closes the gap by **pinning** DNS resolution: validation does the lookup once, captures the approved IP set, and the actual request runs inside a context that returns those exact IPs for the target hostname — so the connect step never performs a fresh, attacker-influenced DNS query.

## Approach

1. `_is_url_safe(url)` now returns `(is_safe, reason, validated_ips)`. The IP list is the deduped set of public addresses that passed the policy.
2. New `_pinned_dns_resolution(target_hostname, pinned_ips)` context manager swaps `socket.getaddrinfo` for one that:
   - Returns synthetic addr-info tuples (IPv4 or IPv6, correctly shaped) for `target_hostname`.
   - Falls through to the original resolver for any other host, so unrelated DNS in the same call stack is unaffected.
   - Restores the original resolver on exit.
3. `_request_with_safe_redirects` now wraps each `requests.request(...)` call inside `_pinned_dns_resolution(hostname, validated_ips)`. Each redirect hop revalidates **and** re-pins for the new target hostname, so a redirect to a fresh host gets that host's freshly-validated IPs (verified in `test_redirect_repins_to_new_target_host`).
4. A module-level `threading.Lock` serialises overlapping pin scopes — the swap is process-global, so concurrent pins must not interleave.

This works for HTTPS too: connecting to the pinned IP and TLS-handshaking with the original hostname (SNI / cert validation) is exactly what `urllib3` already does when given a hostname that resolves to a single IP.

## Tests

`tests/test_http_security.py` gains 9 cases on top of the 2 existing ones:

- `_pinned_dns_resolution` returns pinned IPs for the target host without hitting the system resolver.
- Falls through for unrelated hosts.
- Restores `socket.getaddrinfo` on exit.
- Builds correct IPv6 addr-info tuples.
- **`test_dns_rebinding_attack_uses_pinned_ip_during_connect`** — the headline test: a `getaddrinfo` mock returns `8.8.8.8` for the first lookup (validation) and `192.168.1.1` for any subsequent lookup (the rebind). Inside `requests.request` we re-resolve and assert the result is `8.8.8.8`. Without the pin this test fails (the attacker's rebind would be honoured).
- Each redirect re-pins to its own target host's validated IPs.
- `_is_url_safe` now returns the dedup'd validated IP list, and still blocks when *any* resolved IP is private (mixed-A-record case).

## Local gates

- `PYTHONPATH=. pytest tests/` → **116 passed** (108 pre-existing + 9 new pinning + 2 pre-existing http_security tests still passing — no regressions).
- `ruff check agent/tools/http_request.py tests/test_http_security.py` → All checks passed.
- `ruff format --check agent/tools/http_request.py tests/test_http_security.py` → 2 files already formatted.
- `python -m py_compile agent/tools/http_request.py` → clean.

## Notes

- The `Closes #1186` link covers `agent/tools/http_request.py`. The same `_request_with_safe_redirects` is reused by `agent/tools/fetch_url.py`, so the pinning fix applies to that path too — no further changes needed.
- The pin only intercepts the **target** hostname; lookups for proxies, DNS-over-HTTPS endpoints, etc., go through the system resolver normally.
- The lock serialises pinned requests within a single process. `requests.request` already does its own DNS work synchronously inside a single Session, so this doesn't change practical concurrency for any single agent run.

### DIFF
--- a/agent/tools/http_request.py
+++ b/agent/tools/http_request.py
@@ -1,5 +1,8 @@
+import contextlib
 import ipaddress
 import socket
+import threading
+from collections.abc import Iterator
 from typing import Any
 from urllib.parse import urljoin, urlparse
 
@@ -7,23 +10,36 @@ import requests
 
 _MAX_REDIRECTS = 5
 
+# Serialises pinned DNS resolution. _pinned_dns_resolution swaps the global
+# socket.getaddrinfo, so concurrent pin scopes inside the same process must not
+# overlap. The lock is held only for the duration of a single HTTP request.
+_DNS_PIN_LOCK = threading.Lock()
 
-def _is_url_safe(url: str) -> tuple[bool, str]:
-    """Check if a URL is safe to request (not targeting private/internal networks)."""
+
+def _is_url_safe(url: str) -> tuple[bool, str, list[str]]:
+    """Check if a URL is safe to request (not targeting private/internal networks).
+
+    Returns ``(is_safe, reason, validated_ips)``. ``validated_ips`` is the list of
+    resolved IP addresses that passed the SSRF policy; empty when ``is_safe`` is
+    False. The list is what callers must pin DNS resolution to during the actual
+    request — re-resolving on connect lets a short-TTL attacker DNS-rebind to a
+    private address after this check passes.
+    """
     try:
         parsed = urlparse(url)
         if parsed.scheme not in {"http", "https"}:
-            return False, f"Unsupported URL scheme: {parsed.scheme or '<missing>'}"
+            return False, f"Unsupported URL scheme: {parsed.scheme or '<missing>'}", []
 
         hostname = parsed.hostname
         if not hostname:
-            return False, "Could not parse hostname from URL"
+            return False, "Could not parse hostname from URL", []
 
         try:
             addr_infos = socket.getaddrinfo(hostname, None)
         except socket.gaierror:
-            return False, f"Could not resolve hostname: {hostname}"
+            return False, f"Could not resolve hostname: {hostname}", []
 
+        validated_ips: list[str] = []
         for addr_info in addr_infos:
             ip_str = addr_info[4][0]
             try:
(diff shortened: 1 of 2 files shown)


agent/tools/http_request.py
```diff
@@ -56,23 +141,32 @@ def _request_with_safe_redirects(
...
-            current_method,
-            current_url,
-            timeout=timeout,
-            allow_redirects=False,
-            **request_kwargs,
-        )
+        parsed = urlparse(current_url)
+        hostname = parsed.hostname or ""
+
+        with _pinned_dns_resolution(hostname, validated_ips):
```
johannes117, line 214: this does not cover proxy envs: requests resolves the proxy locally, then the proxy resolves the target host after validation

agent/tools/http_request.py
```diff
@@ -32,11 +48,80 @@ def _is_url_safe(url: str) -> tuple[bool, str]:
                 continue
 
             if ip.is_private or ip.is_loopback or ip.is_link_local or ip.is_reserved:
```
johannes117: this still allows some non-public/special-use ranges like 100.64.0.0/10; safer to require globally routable ips and test edge ranges

agent/tools/http_request.py
```diff
@@ -32,11 +48,80 @@ def _is_url_safe(url: str) -> tuple[bool, str]:
...
+        original = socket.getaddrinfo
+
+        def _pinned(host, port, *args, **kwargs):  # type: ignore[no-untyped-def]
+            if host and str(host).lower() == target:
+                infos = _build_for_port(int(port) if port else 0)
+                if infos:
+                    return infos
+            return original(host, port, *args, **kwargs)
+
+        socket.getaddrinfo = _pinned  # type: ignore[assignment]
```
johannes117, line 164: this mutates socket.getaddrinfo process-wide; the lock only protects this helper, so unrelated threads can still observe the pinned resolver during the request

voidborne-d:
Thanks @johannes117 — all three are real and now fixed in `ba5ff3c`. Quick summary per point:

**1. CGN (and friends): switched to `ip.is_global`.**
The old `is_private | is_loopback | is_link_local | is_reserved` chain misses several special-use ranges. The biggest is RFC 6598 carrier-grade NAT (`100.64.0.0/10`), but `0.0.0.0` (unspecified) and `255.255.255.255` (broadcast) also slip through some of them depending on Python version. `ip.is_global` rejects all of them in one shot. Multicast (`224.0.0.0/4`) is still globally scoped per `ipaddress` so I block it explicitly. Tests added covering CGN low/high, `224.0.0.1`, `239.255.255.250`, `0.0.0.0`, `255.255.255.255`.

**2. Process-wide mutation → thread-scoped resolver.**
The installed `_pinned` now compares `threading.get_ident()` to the pin-issuing thread and falls through to the original resolver for any other thread, even when they look up the pinned target hostname. So the swap of `socket.getaddrinfo` is still visible process-wide, but it's a no-op for unrelated threads. The module lock is kept to serialise *overlapping pin scopes* inside the same process — without it, two pin-issuers can clobber each other's `_pinned` mid-request and a request can run with a stranger's resolver. New test spins up a worker thread, has it resolve the pinned hostname inside the pin window, and asserts the worker saw the original resolver, not the pin.

**3. Proxy path: refuse rather than silently bypass.**
You're right that the pin is only authoritative for *direct* DNS — when `HTTP_PROXY`/`HTTPS_PROXY` routes the request through a proxy, the destination is resolved on the proxy side. Rather than pretend the guard still holds, the request is now refused with a clear reason when a proxy is configured for the target URL (via `requests.utils.get_environ_proxies` + `should_bypass_proxies`). Operators who explicitly want to allow proxies can add the host to `NO_PROXY` or unset the proxy env for SSRF-sensitive calls — the proxy's own egress filter then becomes the authoritative gate, and that's a deliberate operator decision rather than an accidental local-pin gap. Tests cover (a) `HTTP_PROXY` set → blocked, (b) `NO_PROXY` exempts the host → proceeds, (c) no proxy env → proceeds.

Net diff: `+204 / -7` (1 logic file, 10 new tests). 21/21 security tests, 126/126 full suite, `ruff check` and `ruff format --check` both clean. Happy to split any of these into a separate PR if you'd prefer to land the CGN/`is_global` change first and then iterate on the threading model + proxy posture.

voidborne-d:
Good catch — the IDNA mismatch is exactly what would have happened. Pushed `28ba16f` on the same branch:

- `_pinned_dns_resolution` now tracks `{raw, raw.encode('idna')}` and the comparison uses `in targets` instead of `==`.
- Added a regression test `test_pin_matches_idna_encoded_hostname_for_idn_targets` that pins `例え.jp` and asserts the pin matches both the Unicode and Punycode (`xn--r8jz45g.jp`) lookups.

Full suite (22 tests) passes locally on Python 3.14 / urllib3 2.7.0.

PR #1260, opened 2026-05-08
fix(open-swe): surface reviewer diff-prep failures instead of swallowing

johannes117:
## Summary

When the reviewer's repo prep (`_ensure_repo_checked_out`) or diff computation (`compute_diff_in_sandbox`) failed, the bare `except Exception: logger.exception(...)` in `get_reviewer_agent` swallowed the error, handed the agent an empty diff, and the agent obediently published a misleading "no issues found" review.

This lands minimal instrumentation so we can actually see the failure on the next run:

- `_ensure_repo_checked_out` and `compute_diff_in_sandbox` now check the sandbox `exit_code` and raise `RuntimeError` with the sandbox `output` (combined stdout+stderr) on non-zero exit.
- The prep block in `get_reviewer_agent` re-raises after `logger.exception`, so the failure surfaces in the LangSmith run trace instead of producing a fake clean review.

Once we trigger another reviewer run on a private PR, the trace will tell us whether it's `git fetch <sha>`, `gh repo clone`, a stale warm-sandbox clone, or something else. Real fix lands in a follow-up PR after we see the error.

## Test plan

- [x] `make lint`
- [x] `uv run pytest -x tests/ -k reviewer` (65 passed)
- [ ] Trigger a reviewer run on a private PR and confirm the failure shows up in the LangSmith trace with a useful message

### DIFF
--- a/agent/reviewer.py
+++ b/agent/reviewer.py
@@ -179,7 +179,14 @@ async def _ensure_repo_checked_out(
     )
     import asyncio
 
-    await asyncio.to_thread(sandbox_backend.execute, script)
+    result = await asyncio.to_thread(sandbox_backend.execute, script)
+    exit_code = getattr(result, "exit_code", None)
+    if exit_code not in (0, None):
+        output = getattr(result, "output", "") or ""
+        raise RuntimeError(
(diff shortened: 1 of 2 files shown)


johannes117:
@open-swe review

PR #1271, opened 2026-05-08
fix: restore Co-authored-by trailer for triggering user

johannes117:
## Summary
- Investigation: the `Co-authored-by` trailer (and the "_Opened collaboratively by …_" PR-body note) silently disappeared in #1238, when we moved off the custom `commit_and_open_pr` tool / `open_pr` middleware to the gh CLI. Those were the only callers of the helpers in `agent/utils/authorship.py`, so since then the agent has driven commits and PRs entirely without user attribution.
- Fix: resolve the triggering user's identity in `get_agent` (reusing the existing `resolve_triggering_user_identity` helper, which already handles the GitHub/Slack/Linear paths) and inject a new `### Collaborative Attribution` section into the system prompt with the exact trailer and PR-body line to use. The section is only rendered when an identity is resolvable, so runs without a known triggering user behave exactly as before.

## Test plan
- [ ] `make test` (added two new prompt-rendering tests covering the with/without-identity cases)
- [ ] After deploy, trigger a Slack-initiated run and confirm the resulting commit message contains `Co-authored-by: <user> <email>` and the draft PR body ends with `_Opened collaboratively by <user> and open-swe._`

### DIFF
--- a/agent/prompt.py
+++ b/agent/prompt.py
@@ -2,6 +2,7 @@ import logging
 import os
 from pathlib import Path
 
+from .utils.authorship import CollaboratorIdentity
 from .utils.github_comments import UNTRUSTED_GITHUB_COMMENT_OPEN_TAG
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


johannes117:
@open-swe review

PR #1438, opened 2026-06-06
feat: validate LLM API keys on startup

### LINKED ISSUE #1437: Enhancement: Validate LLM API keys at startup to prevent silent failures
Hey team! 👋

I noticed while getting set up that if you forget to add your LLM API key (like `OPENAI_API_KEY` or `GOOGLE_API_KEY`) to the `.env` file, the backend server still starts up perfectly fine. It only crashes later when you actually try to trigger the agent and it attempts to make an API call. 

**Proposal:**
It would be great to have a quick validation check in `agent/server.py` (or the lifespan hook) when the server boots up. It could check the configured `DEFAULT_MODEL_ID` and ensure the corresponding API key is present in the environment. If it's missing, it should throw a clear `ValueError` immediately on startup so new developers aren't left scratching their heads later.

I'd love to take a crack at fixing this as a good first issue if you're open to it!

SurajsinghBayas:
Fixes #1437

### Description
Added a `validate_llm_startup_config()` check that runs during the FastAPI `lifespan` hook. This immediately validates that the API key for the configured `DEFAULT_MODEL_ID` (or `LLM_MODEL_ID` override) is present in the environment variables.

Previously, if a developer forgot to configure their LLM API keys, the server would start normally and silently fail later when the agent attempted to process a task. This catches the configuration error instantly at boot time.


agent/utils/model.py
```diff
@@ -168,3 +168,22 @@ def provider_model_kwargs(
         if effort is not None:
             kwargs["model_kwargs"] = {"reasoning_effort": effort}
     return kwargs
+
+
+def validate_llm_startup_config() -> None:
+    """Validate that the required API keys are present for the configured LLM model."""
+    import os
```
johannes117: move these imports to the top of the file

SurajsinghBayas: Done! Let me know if you need any other changes!

SurajsinghBayas:
@johannes117 any updates

johannes117 (changes requested):
not safe to merge: startup validation checks llm_model_id/default_model_id, but runtime model selection comes from team/profile/thread config. this can fail valid deployments using non-openai defaults and still miss invalid configured runtime/fallback models. please align validation with the actual model resolution path or scope it clearly to local dev defaults.

SurajsinghBayas:
Great point! To avoid failing valid deployments, I went with your suggestion and scoped this strictly to local dev defaults (by checking if the DASHBOARD_BASE_URL points to localhost).
I also renamed the function to validate_local_dev_llm_config() and updated the docstring so it's completely clear to future developers that this check only runs locally @johannes117

SurajsinghBayas:
@johannes117 let me know if this works !

johannes117 (approved):

PR #1475, opened 2026-06-09
feat(open-swe): stream agent chat via @langchain/react v2 protocol

christian-bromann:
## Summary

- **Replaces the custom thread stream with `@langchain/react`.** The dashboard Agents chat now uses `StreamProvider` + `HttpAgentServerAdapter` (`AgentThreadStreamProvider`) instead of the deleted `useThreadStream` hook and the 2s React Query polling loop. Live transcript updates come from `stream.messages`, projected client-side via `streamMessagesToUi` (mirroring server `state_messages_to_ui`).

- **Adds authenticated LangGraph v2 proxy routes on the dashboard API.** New endpoints under `/dashboard/api/threads/{id}/` forward `state`, `stream/events`, `commands`, and `history` to the LangGraph runtime with session cookies. `send_dashboard_message` is narrowed to **mid-run queueing only** (returns 409 when idle); new runs are started by the UI through `stream.submit` / `run.start` commands enriched server-side with GitHub token, repo, model/effort, and profile context.

- **Fixes SDK stream subscription edge cases.** `get_dashboard_thread_state` drops `next` while a run is pending/running so `@langchain/react` treats the thread as active and opens the event stream after out-of-band run creation. Stream modes are expanded (`messages`, `tools`, `checkpoints`, `events`, …) for v2 protocol compatibility. `langgraph.json` pins `api_version: 0.10.0rc3`; `deepagents` bumped to `>=0.6.8`.

- **Improves live transcript UX.** Markdown rendering switches from `react-markdown` to **Streamdown** with a latched “live” target (`useLiveMarkdownMessageId`) so token streaming doesn’t flicker between graph steps. The thinking spinner and prompt bar gain settle/stop behavior (`RunStopButton` via `stream.stop()`). Pending prompts from thread creation now carry model/effort and flush through `submitAgentPrompt` once the stream is ready.

- **Refactors diff rendering and stats.** `DiffView` drops the hand-rolled shiki/diff-lines implementation in favor of `@pierre/diffs` `MultiFileDiff`; line-change counts are centralized in `diffStats.ts` (also used by `ToolExecution` and ported git UI). `formatRelativeTime` moves to shared `lib/utils` with `Intl.RelativeTimeFormat`.

- **Minor docs/tooling.** `INSTALLATION.md` link/table touch-ups; `ui/package.json` adds `@langchain/core`, `@langchain/react`, `streamdown` and removes `diff`, `react-markdown`, remark plugins; `SourceControlTile.tsx` excluded from `tsconfig` with an explanatory comment.

- renders reasoning token and subagent cards:

Screenshot 2026-06-11 at 12 13 58 AM


agent/dashboard/thread_api.py
```diff
@@ -519,39 +540,157 @@ async def _start_agent_run(
...
+        login,
+        metadata,
+        overrides=overrides,
     )
-    run_id = run.get("run_id") if isinstance(run, dict) else getattr(run, "run_id", None)
-    await client.threads.update(
-        thread_id=thread_id,
-        metadata={"latest_run_id": run_id, "latest_run_status": "pending", "updated_at_ms": now_ms},
+    if isinstance(client_configurable, dict):
+        merged_configurable = {**merged_configurable, **client_configurable}
```
johannes117: client configurable wins this merge, so a logged-in dashboard user can set github_login/user_email in run.start and the run resolves another user's stored oauth token (resolve_github_token keys off configurable github_login). they can also override repo/source. reverse the merge order or allowlist only agent_model_id/agent_effort from the client

agent/dashboard/thread_api.py
```diff
@@ -704,6 +801,129 @@ async def delete_dashboard_thread(thread_id: str, login: str, *, email: str | No
...
+    await _authorized_thread_metadata(thread_id, login, email=email)
+    url = f"{langgraph_url().rstrip('/')}/threads/{thread_id}/commands"
+    headers = {"Content-Type": content_type}
+
+    outgoing = body
+    try:
+        parsed = json.loads(body)
+    except json.JSONDecodeError:
+        parsed = None
+    if isinstance(parsed, dict):
```
johannes117: non-dict bodies (e.g. a json array of batched commands) skip enrichment entirely and are forwarded with fully client-controlled config/configurable. reject non-dict bodies with 400

agent/dashboard/thread_api.py
```diff
@@ -704,6 +801,129 @@ async def delete_dashboard_thread(thread_id: str, login: str, *, email: str | No
...
+) -> AsyncIterator[bytes]:
+    await _authorized_thread_metadata(thread_id, login, email=email)
+    url = f"{langgraph_url().rstrip('/')}/threads/{thread_id}/stream/events"
+    headers = {
+        "Content-Type": content_type,
+        "Accept": "text/event-stream",
+    }
+
+    try:
+        async with httpx.AsyncClient(timeout=httpx.Timeout(None)) as client:
```
johannes117: these raw httpx proxies send no auth headers, unlike the sdk client which picks up LANGSMITH_API_KEY. against an auth-required deployment (LANGGRAPH_URL_PROD) all three new proxies 403 while the sdk paths work

agent/dashboard/thread_api.py
```diff
@@ -704,6 +801,129 @@ async def delete_dashboard_thread(thread_id: str, login: str, *, email: str | No
...
+            async with client.stream("POST", url, content=body, headers=headers) as response:
+                response.raise_for_status()
+                async for chunk in response.aiter_bytes():
+                    yield chunk
+    except httpx.HTTPStatusError:
```
johannes117: an upstream 4xx/5xx here becomes an empty 200 sse stream - the client sees the stream open and close with no data and no error. check status before committing the stream, or emit an sse error event

agent/dashboard/thread_api.py
```diff
@@ -704,6 +801,129 @@ async def delete_dashboard_thread(thread_id: str, login: str, *, email: str | No
...
+    outgoing = body
+    try:
+        parsed = json.loads(body)
+    except json.JSONDecodeError:
+        parsed = None
+    if isinstance(parsed, dict):
+        enriched = await _enrich_run_start_command(thread_id, login, parsed, email=email)
+        outgoing = json.dumps(enriched).encode()
+
+    async with httpx.AsyncClient(timeout=httpx.Timeout(None)) as client:
```
johannes117: timeout(none) makes sense for the sse proxy but commands/history are plain request/response - a hung langgraph leaves the dashboard request open forever. use a finite timeout

agent/dashboard/thread_api.py
```diff
@@ -591,64 +729,23 @@ async def send_dashboard_message(
...
-        queued = await queue_message_for_thread(
-            thread_id,
-            queue_payload,
-        )
-        if not queued:
-            raise HTTPException(502, "failed to queue follow-up message")
-        thread = await client.threads.get(thread_id)
-        return _thread_summary(
-            thread if isinstance(thread, dict) else {"thread_id": thread_id, "metadata": metadata}
+    if not await is_thread_active(thread_id):
```
johannes117: is_thread_active fails open (returns false on any exception), so a transient langgraph error now becomes a 409 telling the ui the thread is idle, prompting a run.start against a possibly-busy thread. worth distinguishing couldnt-determine (502) from genuinely idle

agent/dashboard/thread_api.py
```diff
@@ -704,6 +801,129 @@ async def delete_dashboard_thread(thread_id: str, login: str, *, email: str | No
...
+    result = state if isinstance(state, dict) else dict(state)
+    # The SDK's `useStream` opens its live event subscription only when the
+    # hydrated `getState()` looks active (`next` non-empty / absent). When a
+    # run was just started out-of-band (our REST run-create), the latest
+    # checkpoint can still be the previous finished one with `next == []`,
+    # which the SDK reads as idle and never opens the stream. Drop `next`
+    # while a run is pending/running so the SDK treats the thread as active.
+    run_status = await _latest_run_status(thread_id)
+    if run_status in {"pending", "running"}:
+        result.pop("next", None)
```
johannes117, line 890: state and run status come from two separate calls so this is best-effort: a run starting between them reproduces the looks-idle bug with a narrower window. also costs an extra runs.list per state fetch

agent/dashboard/thread_api.py
```diff
@@ -519,39 +540,157 @@ async def _start_agent_run(
...
+    command: dict[str, Any],
+    *,
+    email: str | None = None,
+) -> dict[str, Any]:
+    if command.get("method") != "run.start":
+        return command
 
-    run = await client.runs.create(
+    client = langgraph_client()
+    thread = await client.threads.get(thread_id)
```
johannes117: proxy_dashboard_thread_commands already fetched this thread and asserted ownership via _authorized_thread_metadata - pass the metadata through instead of refetching

ui/src/lib/agents/provider/useSubmitAgentMessage.ts
```diff
@@ -0,0 +1,73 @@
...
+          content: vars.content,
+          images: vars.images,
+          model_id: vars.model_id,
+          effort: vars.effort,
+        });
+        return;
+      }
+
+      await stream.submit(
+        { messages: [{ type: "human", content: vars.content }] },
```
johannes117: vars.images never makes it into the submitted message, here and in submitAgentPrompt below - only the mid-run queue path sends images. attaching an image on an idle thread silently loses it, a regression vs the old sendMessage endpoint

ui/src/lib/agents/provider/useSubmitAgentMessage.ts
```diff
@@ -0,0 +1,73 @@
...
+ * injects the message into the *current* run before the next model call — the
+ * same mid-run follow-up path used by Slack, Linear, and GitHub webhooks.
+ */
+export function useSubmitAgentMessage(threadId: string) {
+  const queryClient = useQueryClient();
+  const stream = useAgentThreadStream();
+
+  return useMutation({
+    mutationFn: async (vars: SendAgentMessageVariables) => {
+      if (stream.isLoading) {
```
johannes117: queue-vs-submit is decided from local stream.isLoading only. a run started via slack/linear webhook or another tab reads as idle here and this starts a second concurrent run instead of queueing. the old endpoint let the server arbitrate

ui/src/lib/agents/streamMessagesToUi.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+  };
+
+  messages.forEach((raw, index) => {
+    const msgId = typeof raw.id === "string" && raw.id ? raw.id : `msg-${index}`;
+    const timestamp = new Date().toISOString();
+
+    if (HumanMessage.isInstance(raw)) {
+      flushAgentTurn();
+      const text = raw.text.trim();
```
johannes117, line 323: server projection (message_adapter) emits image chunks first and keeps image-only human messages; this drops them, so image-only user messages vanish during streaming and reappear when server data takes over after the run

ui/src/lib/agents/streamMessagesToUi.ts
```diff
@@ -0,0 +1,190 @@
...
+  const flushAgentTurn = () => {
+    if (!agentTurn) return;
+    uiMessages.push({ ...agentTurn, chunks: mergeTextChunks(agentTurn.chunks) });
+    agentTurn = null;
+  };
+
+  const appendAgentChunks = (timestamp: string, chunks: Array<Chunk>) => {
+    if (!agentTurn) {
+      agentTurn = {
+        id: `agent-turn-${++agentTurnSeq}`,
```
johannes117: server uses the first ai message's real id for the agent turn; these synthetic ids mean every message changes id when the view swaps to server-projected thread.messages at run completion - remounting keyed components and breaking the useLiveMarkdownMessageId latch right at the boundary it was built for

ui/src/lib/agents/streamMessagesToUi.ts
```diff
@@ -0,0 +1,190 @@
...
+      };
+    } else {
+      agentTurn.timestamp = timestamp;
+      agentTurn.chunks.push(...chunks);
+    }
+  };
+
+  messages.forEach((raw, index) => {
+    const msgId = typeof raw.id === "string" && raw.id ? raw.id : `msg-${index}`;
+    const timestamp = new Date().toISOString();
```
johannes117: regenerated on every projection call, so live timestamps continuously drift to now and then jump to created_at when server data takes over. use the message's created_at when present

ui/src/components/agents/ported/MessageView.tsx
```diff
@@ -115,7 +117,7 @@ function buildRenderItems(chunks: Chunk[]): RenderItem[] {
     flushExplored();
     items.push({
       type: "text-chunk",
-      key: getChunkRenderKey(chunk, i),
+      key: messageId ? `${messageId}-text` : getChunkRenderKey(chunk, i),
```
johannes117: a message interleaving text/tool/text produces duplicate `${messageId}-text` sibling keys and react can reconcile the second chunk into the first. add a per-chunk index to the key

ui/src/components/agents/AgentThreadView.tsx
```diff
@@ -73,71 +51,65 @@ export function AgentThreadView({ thread }: AgentThreadViewProps) {
...
-        isPendingPromptConfirmed(entry, thread.messages)
-      )
-      return next.length === prev.length ? prev : next
+    if (pendingPrompts.length === 0) return
+    if (stream.isLoading || stream.isThreadLoading) return
+    if (pendingSubmitStarted.current) return
+
+    const entry = pendingPrompts[0]
+    if (!entry) return
+    pendingSubmitStarted.current = true
```
johannes117: if submit rejects after the run was actually created server-side (connection drop mid-stream), the ref resets and this effect re-submits the same prompt. drop the entry on successful submit instead of relying on the text match below. also the .catch is dead code given the .finally

ui/src/components/agents/AgentThreadView.tsx
```diff
@@ -73,71 +51,65 @@ export function AgentThreadView({ thread }: AgentThreadViewProps) {
...
-                  pendingImageKey(p) === pendingImageKey(entry)
-              )
-            )
-          },
-        }
+  useEffect(() => {
+    setPendingPrompts((prev) => {
+      if (prev.length === 0) return prev
+      const next = dropPendingPrompts(thread.id, (entry) =>
+        userMessageTexts.has(entry.prompt)
```
johannes117: text-only confirmation: image-only prompts never match, and an earlier text-identical user message drops a pending image prompt prematurely. the old isPendingPromptConfirmed compared image payloads too

ui/src/components/agents/ported/CloudPromptBar.tsx
```diff
@@ -11,12 +14,52 @@ import {
...
+        )
+        void queryClient.invalidateQueries({ queryKey: agentThreadKeys.all })
+      }
+    } finally {
+      setStopping(false)
+    }
+  }
+
+  return (
+    <button
```
johannes117: icon-only button - use the design-system IconButton instead of a raw button

ui/src/lib/utils.ts
```diff
@@ -1,7 +1,54 @@
...
+    ["minute", 60 * 1000],
+    ["hour", 60 * 60 * 1000],
+    ["day", 24 * 60 * 60 * 1000],
+    ["week", 7 * 24 * 60 * 60 * 1000],
+  ];
+
+  for (const [unit, ms] of units) {
+    const value = Math.round(diffMs / ms);
+
+    if (Math.abs(value) < (unit === "week" ? Infinity : 60)) {
```
johannes117: the <60 cutoff applies to every unit, so 36 hours renders as 36 hours ago and 45 days as 45 days ago. use per-unit cutoffs (60 min / 24 h / 7 d). also no month/year units, very old timestamps show hundreds of weeks ago

ui/src/lib/agents/streamConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Explicit v2-protocol paths bound to the dashboard's auth proxy
+ * (`agent/dashboard/routes.py`). Using a custom transport with these
+ * paths keeps commands, the event stream, and `getState` hydration all
+ * flowing through the same credentialed fetch — the built-in `apiUrl`
+ * branch hydrates state via an internal client that doesn't carry the
+ * dashboard session cookie.
+ */
+export function agentStreamPaths(threadId: string) {
+  return {
+    commands: `/threads/${threadId}/commands`,
```
johannes117: threadId interpolated unencoded, unlike api.ts which encodeURIComponents everywhere

ui/src/components/agents/ported/ToolExecution.tsx
```diff
@@ -27,29 +27,8 @@ function getFileName(path: string): string {
...
-
-  for (const part of parts) {
-    const lineCount = countLines(part.value);
-    if (part.added) additions += lineCount;
-    else if (part.removed) deletions += lineCount;
-  }
-
-  return { additions, deletions };
+function countLineChangesForTool(originalContent: string | null | undefined, newContent: string, filePath: string) {
+  return countLineChanges(originalContent, newContent, filePath);
```
johannes117: 1:1 wrapper - call countLineChanges directly

ui/src/components/agents/ported/DiffView.tsx
```diff
@@ -236,11 +19,8 @@ export function DiffView({ diffData }: DiffViewProps) {
     );
   }
 
-  const displayLines = expanded ? hunkLines : hunkLines.slice(0, MAX_COLLAPSED_LINES);
-  const hasMoreLines = hunkLines.length > MAX_COLLAPSED_LINES;
-  const hiddenCount = hunkLines.length - MAX_COLLAPSED_LINES;
-
-  if (hunkLines.length === 0) {
+  const stats = countLineChanges(originalContent, newContent, filePath);
```
johannes117: parses the diff on every render and MultiFileDiff parses the same pair again internally - wrap the stats in useMemo

langgraph.json
```diff
@@ -1,13 +1,16 @@
 {
   "$schema": "https://langgra.ph/schema.json",
   "python_version": "3.12",
+  "api_version": "0.10.0rc3",
```
johannes117, line 4: intentional to pin a release candidate?

johannes117:
test coverage: the rewritten tests only cover the repo flags in _build_dashboard_configurable. nothing exercises ownership checks on the four new proxy endpoints, the run.start enrichment/merge (where the github_login override issue lives), or the new 409-on-idle path of send_dashboard_message

langgraph.json
christian-bromann: Yep, will remove once API is released.

ui/src/lib/agents/AgentThreadStreamProvider.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }, [queryClient]);
+
+  return (
+    <StreamProvider
+      apiUrl={agentStreamApiUrl}
+      assistantId={AGENT_ASSISTANT_ID}
+      fetch={dashboardFetch}
+      threadId={threadId ?? undefined}
+      onCreated={onCreated}
+      onCompleted={onCompleted}
```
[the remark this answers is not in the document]
christian-bromann: The built-in SSE `StreamProvider` does not use the legacy LGP `POST /threads` / `runs.stream` shape, it uses the same v2 agent protocol (`POST /threads/{id}/commands`, `/stream/events`, `GET /state`) that `HttpAgentServerAdapter` did, and the dashboard proxy still serves all of those. New threads are created lazily on the first `run.start`: the SDK mints the thread id client-side, and `proxy_dashboard_thread_commands` stamps the dashboard record server-side, which is why removing `POST /dashboard/api/threads` is intentional and covered by `test_proxy_commands_lazily_creates_missing_thread_only_for_run_start`.

agent/dashboard/routes.py
```diff
@@ -949,6 +973,73 @@ async def api_delete_thread(
...
+    request: Request,
+    session: dict[str, Any] = _SESSION_DEP,
+) -> Response:
+    body = await request.body()
+    status_code, content, media_type = await proxy_dashboard_thread_commands(
+        thread_id,
+        session["sub"],
+        body,
+        email=session.get("email"),
+        content_type=request.headers.get("content-type", "application/json"),
```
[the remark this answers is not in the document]
christian-bromann: Added router-level Origin/Referer validation on all `/dashboard/api` mutations (checked against `DASHBOARD_BASE_URL` + `DASHBOARD_ALLOWED_ORIGINS`) and require `Content-Type: application/json` on the thread proxy endpoints (`/commands`, `/history`, `/stream/events`) so cross-site POSTs can’t bypass CORS preflight with `text/plain`.

johannes117 (approved):

PR #1527, opened 2026-06-15
feat: live reviewer eval logs on a dedicated admin page

johannes117:
Moves the reviewer eval runner off the admin page onto its own `/admin/evals` page (linked from the admin tab like Review Style Prompts), and adds a live log tail.

## What changed
- **Backend** (`eval_jobs.py`): stream the eval subprocess stdout into a rolling `log_tail` and persist it on each heartbeat, instead of only capturing it at process exit. Experiment-URL detection is retained across the rolling window.
- **Frontend**: new `/admin/evals` page with the run controls, status, and a scrollable log viewer (auto-follow). Admin page now links to it instead of embedding the section inline.

The log tail updates roughly every 10s (heartbeat cadence) while the UI polls every 5s.

### DIFF
--- a/agent/dashboard/eval_jobs.py
+++ b/agent/dashboard/eval_jobs.py
@@ -45,6 +45,10 @@ _WORKER_ID = uuid.uuid4().hex
 # them. The store record is the source of truth across workers/requests.
 _PROCS: dict[str, asyncio.subprocess.Process] = {}
 
+# Rolling tail of subprocess output, kept by the owning worker so the heartbeat
+# loop can persist a live log tail to the store while the eval runs.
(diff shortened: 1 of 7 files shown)


johannes117:
Addressed in c8b1d4a7. The flagged `createStart` import was part of an SSR-register block that is local route-generator drift (it's not on `main`) and got swept into the first commit — I've stripped it so the generated file matches `main` plus only the new `/admin/evals` route.

Note the build wasn't actually failing on it: `routeTree.gen.ts` is `// @ts-nocheck`, so `noUnusedLocals` doesn't apply, and the Vercel preview deployed green with the import present. Removed it anyway to keep the diff clean.

PR #1628, opened 2026-06-27
fix: stop reviewer/chat agent factories from mutating the caller's RunnableConfig

### LINKED ISSUE #1584: Bug: get_reviewer_agent and get_chat_agent mutate the caller's RunnableConfig in-place
### Bug

[`agent/reviewer.py::get_reviewer_agent(config: RunnableConfig)`](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/reviewer.py#L815-L820) and [`agent/chat.py::get_chat_agent(config: RunnableConfig)`](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/chat.py#L106-L114) are agent factories — by the `langgraph.json` entrypoint contract their job is to take a per-run `RunnableConfig` and return a compiled `Pregel`. They both **mutate the caller's `config` dict in-place** rather than treating it as read-only.

Three concrete writes (all from `develop` HEAD `9949077`):

| File | Line | Mutation |
|---|---:|---|
| `agent/reviewer.py` | [819](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/reviewer.py#L819) | `config["recursion_limit"] = DEFAULT_RECURSION_LIMIT` |
| `agent/reviewer.py` | [1005](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/reviewer.py#L1005) | `config["configurable"]["diff_text"] = pr_diff_text` |
| `agent/reviewer.py` | [1006](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/reviewer.py#L1006) | `config["configurable"]["diff_line_set"] = pr_diff_line_set` |
| `agent/chat.py` | [109](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/chat.py#L109) | `config["recursion_limit"] = DEFAULT_RECURSION_LIMIT` |

`DEFAULT_RECURSION_LIMIT = 9_999` ([`agent/server.py:476`](https://github.com/langchain-ai/open-swe/blob/9949077bc7edb806efdffa8ea454090c2705bcf4/agent/server.py#L476)).

### Impact

1. **Caller's `recursion_limit` is silently overwritten.** Any caller that passes a deliberately lower ceiling — e.g. `langgraph_sdk.client.threads.create(..., config={"recursion_limit": 200})` for a bounded review, or `recursion_limit=25` for a fast-fail test — observes `9_999` after the factory returns. There is no warning, no log line, and the override looks accepted.

2. **PR diff data leaks across runs that share a base config dict.** Lines 1005-1006 write the freshly fetched PR diff (`diff_text` can easily be multi-MB) into the same `configurable` sub-dict the caller passed in. Any code path that reuses a base config (e.g. a webhook handler that builds a per-request config from a tenant default and overlays a few fields) will see one run's diff visible to the next. The two writes are not synchronized either, so two concurrent reviewer invocations that happen to share a config base race on `diff_text` / `diff_line_set`.

3. **Violates the LangGraph factory contract.** The rest of the LangGraph runtime treats `RunnableConfig` as immutable — `Pregel.invoke` constructs a fresh `merged_config` rather than mutating its input. Factory authors and `Runnable.with_config(...)` users expect the same.

### Reproducer

I extracted the mutations from the live source at HEAD `9949077` with a 50-line AST walker. No imports of open-swe required (open-swe is a langgraph app, not a PyPI package); this just parses `agent/reviewer.py` and `agent/chat.py` with `ast` and finds every `config[...] = ...` inside the two factories.

```python
# repro_live.py
import ast, sys
from pathlib import Path

def find_param_mutations(source, function_name, param_name):
    hits = []
    for node in ast.walk(ast.parse(source)):
        if not isinstance(node, (ast.AsyncFunctionDef, ast.FunctionDef)): continue
        if node.name != function_name: continue
        for sub in ast.walk(node):
            if not isinstance(sub, ast.Assign): continue
            for tgt in sub.targets:
                root = tgt
                while isinstance(root, ast.Subscript): root = root.value
                if isinstance(root, ast.Name) and root.id == param_name:
                    hits.append((sub.lineno, ast.unparse(tgt), ast.unparse(sub.value)[:60]))
    return hits

repo = Path(sys.argv[1])
for fn, path in [("get_reviewer_agent", repo/"agent/reviewer.py"),
                 ("get_chat_agent",     repo/"agent/chat.py")]:
    print(f"=== {path.name}::{fn} ===")
    for ln, tgt, val in find_param_mutations(path.read_text(), fn, "config"):
        print(f"  line {ln:4d}: {tgt} = {val}")
```

The script also checks for any **defensive copy** of `config` inside each factory (`copy.deepcopy(config)`, `dict(config)`, `{**config, …}`, `config.copy()`, `merged_config`) that would protect the caller's dict. Output against HEAD `9949077`:

```
=== reviewer.py::get_reviewer_agent ===
  Mutations of `config` (3 found):
    line  819: config['recursion_limit'] = DEFAULT_RECURSION_LIMIT
    line 1005: config['configurable']['diff_text'] = pr_diff_text
    line 1006: config['configurable']['diff_line_set'] = pr_diff_line_set
  Defensive copies of `config` (0 found):
    NONE — `config` is never deepcopy'd / rebound to a fresh dict.
    ∴ every mutation above lands on the caller's dict.

=== chat.py::get_chat_agent ===
  Mutations of `config` (1 found):
    line  109: config['recursion_limit'] = DEFAULT_RECURSION_LIMIT
  Defensive copies of `config` (0 found):
    NONE — `config` is never deepcopy'd / rebound to a fresh dict.
    ∴ every mutation above lands on the caller's dict.
```

That closes the loop: the assignments exist, and there is no rebind of `config` to a fresh dict anywhere in the function body, so Python's reference semantics guarantee the caller's dict is mutated.

### Suggested fix

Defensive copy at the top of each factory; use `setdefault` instead of unconditional `=` for the recursion-limit default so the caller's explicit choice wins:

```python
import copy

async def get_reviewer_agent(config: RunnableConfig) -> Pregel:
    config = copy.deepcopy(config)
    config.setdefault("recursion_limit", DEFAULT_RECURSION_LIMIT)
    ...
    config["configurable"]["diff_text"] = pr_diff_text
    config["configurable"]["diff_line_set"] = pr_diff_line_set
    return create_deep_agent(...).with_config(config)
```

The same one-line addition (`config = copy.deepcopy(config)`) at the top of `get_chat_agent` fixes the chat factory too.

`Runnable.with_config(config)` returns a new bound runnable rather than mutating, so once the copy lands you can drop the `config["configurable"][...] = ...` writes in favor of the LangChain-idiomatic shape — but a strict defensive copy is the smaller, safer change.

AnnaSuSu:
## What

`get_reviewer_agent` (`agent/reviewer.py`) and `get_chat_agent` (`agent/chat.py`) are langgraph.json entrypoint factories — by contract they take a per-run `RunnableConfig` and return a compiled graph, treating the config as read-only. Both instead **mutate the caller's `config` dict in-place**:

| File | Mutation |
|---|---|
| `agent/reviewer.py` | `config["recursion_limit"] = DEFAULT_RECURSION_LIMIT` (always overwrites) |
| `agent/reviewer.py` | `config["configurable"]["diff_text"] = …` (multi-MB PR diff) |
| `agent/reviewer.py` | `config["configurable"]["diff_line_set"] = …` |
| `agent/chat.py` | `config["recursion_limit"] = DEFAULT_RECURSION_LIMIT` |
| `agent/chat.py` | `config["configurable"]["chat_github_token"] = token` (resolved App token) |

Closes #1584.

## Impact

1. **A caller's explicit `recursion_limit` is silently overwritten** — e.g. `threads.create(..., config={"recursion_limit": 200})` ends up at `9_999` with no warning.
2. **Run-scoped data leaks across runs that share a base config** — a webhook handler that overlays a few fields onto a tenant-default config leaks one run's PR diff (and a GitHub App token) into the next; two concurrent runs race on `diff_text`/`diff_line_set`.
3. Violates the LangGraph factory contract (the runtime treats `RunnableConfig` as immutable).

## Fix

Shallow-copy `config` and its nested `configurable` dict at the top of each factory, so every subsequent write lands on the copy that is bound to the returned agent via `.with_config(config)`. The diff / token still reach the agent at runtime (the agent is built from the copy), but the caller's dict is never touched. `recursion_limit` uses `setdefault` so an explicit caller value wins instead of being silently overwritten (still defaults to `DEFAULT_RECURSION_LIMIT` when unset).

All writes were verified to be at most two levels deep (`config[k]` / `config["configurable"][k]`), so a two-level shallow copy fully isolates the caller's dict — no `deepcopy` needed.

## Tests

- New `tests/test_factory_config_isolation.py` — both factories leave the caller's config untouched (explicit `recursion_limit` preserved; default not written back).
- Updated two tests in `tests/test_reviewer.py` that asserted the pre-fix in-place mutation; they now assert the diff is bound to the **agent's** config (what `add_finding` reads at runtime) and that the caller's input config is left unmutated.

```
tests/test_factory_config_isolation.py ....                         [ 4 passed ]
tests/test_reviewer.py (diff_line_set + diff-fetch-fails)           [ 2 passed ]
full suite: 1212 passed, 1 pre-existing unrelated failure
```

(The single remaining failure — `test_plan_review.py::test_plan_routes_registered` — fails on `main` without this change and is unrelated.)

### DIFF
--- a/agent/chat.py
+++ b/agent/chat.py
@@ -164,8 +164,15 @@ async def _resolve_chat_model(configurable: dict) -> tuple[str, str]:
 
 async def get_chat_agent(config: RunnableConfig) -> Pregel:
     """Get a read-only PR chat agent. No sandbox; PR context comes via config."""
+    # Treat the caller's RunnableConfig as read-only. The body below stashes
+    # run-scoped values (recursion limit, GitHub App token) into the config; without
+    # a copy those writes mutate the caller's dict and leak across runs that share a
+    # base config. Shallow-copy config and the one nested dict we write into. #1584.
+    config = {**config}
+    config["configurable"] = {**config.get("configurable", {})}
+
     thread_id = config["configurable"].get("thread_id")
-    config["recursion_limit"] = DEFAULT_RECURSION_LIMIT
+    config.setdefault("recursion_limit", DEFAULT_RECURSION_LIMIT)
 
     if thread_id is None or not graph_loaded_for_execution(config):
         return create_deep_agent(system_prompt="", tools=[]).with_config(config)
--- a/agent/reviewer.py
+++ b/agent/reviewer.py
@@ -1206,9 +1206,15 @@ class PrepareReviewerRunMiddleware(BasePrepareRunMiddleware):
 
 async def get_reviewer_agent(config: RunnableConfig) -> Pregel:
     """Get or create a reviewer agent with checkpointed run prep."""
+    # Caller's RunnableConfig is a per-run entrypoint arg; keep it read-only. We set
+    # recursion_limit on it below, so shallow-copy config (and the nested configurable
+    # dict) to avoid leaking run-scoped state across runs that share a base config. #1584
+    config = {**config}
+    config["configurable"] = {**config.get("configurable", {})}
+
     thread_id = config["configurable"].get("thread_id", None)
 
-    config["recursion_limit"] = DEFAULT_RECURSION_LIMIT
+    config.setdefault("recursion_limit", DEFAULT_RECURSION_LIMIT)
 
     if thread_id is None or not graph_loaded_for_execution(config):
         logger.info("No thread_id or not for execution, returning reviewer agent without sandbox")
--- a/tests/test_factory_config_isolation.py
+++ b/tests/test_factory_config_isolation.py
@@ -1,0 +1,69 @@
+"""Regression tests for #1584.
+
+The ``get_reviewer_agent`` / ``get_chat_agent`` factories must treat the caller's
+``RunnableConfig`` as read-only: per the langgraph.json entrypoint contract they
+derive a compiled graph from a per-run config and must not mutate the dict the
+caller passed in. Before the fix they wrote ``recursion_limit`` (and, further
+down, the PR diff / GitHub App token) straight onto the caller's dict, silently
+overriding a caller's explicit limit and leaking data across runs that share a
+base config.
+
+These exercise the offline early-return path (no ``thread_id``), which is enough
+to prove the top-of-factory defensive copy: the caller's dict is left untouched.
+"""
+
+from __future__ import annotations
+
+import copy
+from typing import Any
+
+from agent.chat import get_chat_agent
+from agent.reviewer import get_reviewer_agent
+
+
+def _config(**overrides: Any) -> dict[str, Any]:
+    base: dict[str, Any] = {"configurable": {"some_key": "value"}}
+    base.update(overrides)
+    return base
+
+
+async def test_get_reviewer_agent_preserves_caller_config() -> None:
+    config = _config(recursion_limit=200)
+    snapshot = copy.deepcopy(config)
+
+    await get_reviewer_agent(config)
+
+    # An explicit recursion_limit must survive (setdefault, not overwrite), and
+    # nothing else may be written back onto the caller's dict.
+    assert config == snapshot
+
+
+async def test_get_chat_agent_preserves_caller_config() -> None:
+    config = _config(recursion_limit=200)
+    snapshot = copy.deepcopy(config)
+
+    await get_chat_agent(config)
+
+    assert config == snapshot
+
+
+async def test_get_reviewer_agent_does_not_inject_default_into_caller_config() -> None:
+    # When the caller sets no limit, the factory applies its default on its own
+    # copy without mutating the caller's dict.
+    config = _config()
+    snapshot = copy.deepcopy(config)
+
+    await get_reviewer_agent(config)
+
+    assert config == snapshot
+    assert "recursion_limit" not in config
+
+
+async def test_get_chat_agent_does_not_inject_default_into_caller_config() -> None:
+    config = _config()
+    snapshot = copy.deepcopy(config)
+
+    await get_chat_agent(config)
+
+    assert config == snapshot
+    assert "recursion_limit" not in config


AnnaSuSu:
rebased on latest main. heads up: upstream moved the reviewer's diff/token handling into PrepareReviewerRunMiddleware, so the parts of this PR that guarded against diff/token leaking through the config are no longer needed. what's left is the still-present bug: both get_reviewer_agent and get_chat_agent write recursion_limit straight onto the caller's config. so this now just shallow-copies the config and uses setdefault for recursion_limit (chat also still stashes a token, so it keeps that copy too). updated the reviewer tests to match upstream's middleware assertions, kept the config-isolation regression tests. green locally.

AnnaSuSu:
@johannes117 any chance of a look at this one? it's rebased on main and the affected tests pass locally (51 across `tests/test_factory_config_isolation.py` and `tests/test_reviewer.py`).

scope is smaller than when it was opened: upstream moved the reviewer's diff/token handling into `PrepareReviewerRunMiddleware`, so those guards are no longer needed. what remains is a bug that's still there, both `get_reviewer_agent` and `get_chat_agent` write `recursion_limit` straight onto the caller's RunnableConfig. they're langgraph entrypoint factories, so a caller reusing a base config sees it leak across runs, and an explicit lower ceiling gets silently overwritten.

the red Vercel check is the preview deploy waiting on team authorization (see the bot comment above), not the tests.

PR #1630, opened 2026-06-28
fix: update Slack trace reply on web handoff

ramon-langchain:
## Description
Updates the original Slack trace/Web message when a Slack-started run moves to Web, both while the agent is already running and when Web starts a new run from an idle Slack thread. The injected handoff notice now uses Codex-style XML-tagged contextual fragments and is kept as a separate content block before the user prompt.

## Release Note
Slack handoff UX now updates the original Open in Web message when follow-ups move to Web.

## Test Plan
- [x] Verify busy Web follow-ups update the stored Slack trace reply timestamp without posting a new Slack message.
- [x] Verify idle Slack-to-Web run starts inject the handoff instruction and update the original Slack trace reply.
- [x] Verify injected handoff context is formatted as a separate XML-style contextual fragment.

Made by [Open SWE](https://openswe.vercel.app/agents/17f5688d-1f0e-5951-e28b-17dace20ec53)

### DIFF
--- a/agent/dashboard/thread_api.py
+++ b/agent/dashboard/thread_api.py
@@ -19,6 +19,7 @@ from pydantic import BaseModel, ConfigDict, Field
 
 from ..utils.langsmith import get_langsmith_trace_url
 from ..utils.sandbox import create_sandbox
+from ..utils.slack import lookup_slack_thread_run_mapping, update_slack_trace_reply_for_web_handoff
 from ..utils.thread_ops import (
     get_thread_active_status,
     langgraph_client,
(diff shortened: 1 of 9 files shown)


johannes117:
Addressed in d6d8680c.

**Fix:** `store_slack_run_mapping` now reads the existing thread mapping before overwriting it. When `trace_message_ts` is not passed (the follow-up Slack mention path), it preserves the previously stored `trace_message_ts` instead of clobbering it. This ensures `_notify_slack_web_handoff` can still find the original trace reply message on a subsequent move to Web.

Added a test (`test_slack_run_mapping_preserves_trace_message_ts_on_followup_mention`) covering the scenario: first mention stores `trace_message_ts="1.1"`, follow-up mention writes a new `run_id` without it, and the stored timestamp is preserved.

PR #1655, opened 2026-06-30
fix: jump + ResizeObserver settle for review scroll-to

carolinedivittorio:
## Description
The PR reviews page scroll-to (from the blocks on the left or comments on the right) used smooth scrolling plus frame-count correction loops, which raced the diff's height reconciliation under virtualization — so targets landed slightly off, drifted as content settled, or overshot. This switches block/file navigation and finding/comment centering to an instant jump that re-asserts its target via a `ResizeObserver` (the real "layout settled" signal) instead of fixed frame budgets, so it lands deterministically as off-screen cards mount, files expand, and annotation cards measure. Holds bail on user wheel/touch input and a new navigation cancels the previous hold.

## Release Note
Scrolling to a block or comment on the PR reviews page now lands precisely instead of drifting or overshooting.

## Test Plan
- [ ] On a large PR, click blocks in the left agenda and comments on the right; target should land flush/centered without a visible bounce or settle-drift.
- [ ] Navigate rapidly between targets; no fighting between successive jumps.
- [ ] Start a manual scroll right after clicking a target; the auto-hold should yield to you immediately.

Made by [Open SWE](https://openswe.vercel.app/agents/019f1ab8-0873-758e-be1c-aa7698833352)

### DIFF
--- a/ui/src/components/agents/ReviewMainBody.tsx
+++ b/ui/src/components/agents/ReviewMainBody.tsx
@@ -217,46 +217,27 @@ function selectedRangeFromDiff(
   }
 }
 
-// Scroll a file card / group flush to the top of the diff scroller. Under
-// virtualization, scrollIntoView computes its target against estimated row
-// heights; scrolling past unmeasured files reconciles their real heights
-// mid-animation and the Virtualizer re-pins its scroll anchor, which leaves the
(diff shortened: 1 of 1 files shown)


johannes117 (approved):
fantastic, tested locally, performant

PR #1764, opened 2026-07-16
fix: add exc_info to swallowed exception in push re-review webhook

SurajsinghBayas:
In process_github_push_event, the except Exception block that guards fetch_pr_review_threads and reconcile_findings_with_review_threads only logs a warning message without the traceback. This makes it impossible to diagnose what actually went wrong from logs alone.

Added exc_info=True to capture the full exception traceback, making debugging significantly easier.


SurajsinghBayas:
@christian-bromann @wbbradley @pkiv any updates

wbbradley (approved):
LGTM

PR #1793, opened 2026-07-21
feat: optional blocking conclusion for the Open SWE Review check

ericlitman:
## Description

Adds an opt-in **merge-on-clean** mode for the `Open SWE Review` check run. When the `REVIEW_CHECK_BLOCKING` env var is set, the check concludes as `failure` while findings are open, so branch protection / auto-merge can gate on review cleanliness. Three behaviors change together, all behind the flag:

- **Surfaced findings fail the check** instead of concluding `success` with a count.
- **Blocking checks count all standing surfaced findings**, not just newly posted ones — a re-review that surfaces nothing new can't flip a dirty PR green.
- **Unfinished reviews fail instead of settling neutral.** Both settle paths — the after-agent middleware and the run-completion handler from #1779 — now share one conclusion helper (`incomplete_review_check_result`), so a crashed or timed-out reviewer can't satisfy a required check through the completion path (`neutral` passes required checks).

**Default off:** without the env var, every conclusion is byte-identical to today's informational behavior (covered by tests).

The toggle is deliberately env-scoped (deployment-level merge policy) rather than a team setting: whoever operates branch protection owns it.

## Why

We run Open SWE with required-check gating (auto-merge blocked until the review is clean). The informational check can't express "this PR has open findings" to branch protection. This is the minimal seam that makes the existing check usable as a merge gate.

## Test Plan

- [x] Full suite passes (1598); `ruff check`, `ruff format --diff`, `basedpyright` clean
- [x] Flag-off conclusions byte-identical (literal-tuple tests in `tests/github/test_github_checks.py`)
- [x] Blocking mode: surfaced findings → `failure`; standing-findings count on re-review (`tests/reviewer/test_reviewer_publish.py`)
- [x] Crash settle: middleware and completion-handler paths both conclude `failure` under the flag (`tests/middleware/test_settle_review_check.py`, `tests/webhooks/test_completion_webhook.py` — the completion test would have settled `neutral` before this change)

### DIFF
--- a/agent/tools/publish_review.py
+++ b/agent/tools/publish_review.py
@@ -48,7 +48,7 @@ from ..review.publish import (
 )
 from ..review.reconcile import reconcile_findings_with_review_threads
 from ..utils.dashboard_links import dashboard_review_url
-from ..utils.github_checks import review_check_conclusion
+from ..utils.github_checks import review_check_blocking_enabled, review_check_conclusion
 from ..utils.github_token import (
     GitHubAuthError,
     get_github_token,
@@ -192,6 +192,20 @@ async def _resolve_review_trace_url(thread_id: str, config_override: object) -> 
 
 def _is_reviewer_eval_mode(configurable: dict[str, Any]) -> bool:
     return configurable.get("reviewer_eval") is True or configurable.get("eval") is True
+
+
+async def _review_check_finding_count(thread_id: str, newly_posted_count: int) -> int:
+    # Blocking checks count all standing findings; informational checks preserve publish-time count.
+    if not review_check_blocking_enabled():
+        return newly_posted_count
+    findings = await list_findings_async(thread_id)
+    return sum(
+        1
+        for finding in findings
+        if finding.get("status", "open") == "open"
+        and isinstance((surface := finding.get("surface")), dict)
+        and surface.get("state") in {"surfaced", "resolve_pending"}
+    )
 
 
 async def _publish_review_eval_dry_run_async(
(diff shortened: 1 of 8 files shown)


jkennedyvz (commented):
Automated review workflow: this review found two concrete ways the proposed blocking check can report success while findings remain open. These comments were generated and validated by an automated code-review workflow.

agent/utils/github_checks.py
```diff
@@ -153,18 +154,36 @@ async def post_autofix_status_check(
     return True
 
 
-def review_check_conclusion(surfaced_count: int) -> tuple[CheckConclusion, str, str]:
-    """Map a publish result to (conclusion, title, summary).
+def review_check_blocking_enabled() -> bool:
```
jkennedyvz, line 157: **Automated review workflow — High: unchanged-diff pushes incorrectly pass**

When blocking mode is enabled, a push that preserves the normalized PR diff—such as an empty commit or a rebase producing the same patch—takes the unchanged-diff branch in `process_github_push_event`. That branch creates a check on the new HEAD and hardcodes `conclusion="success"` in `agent/webhooks/github.py`, without consulting this flag or the standing surfaced findings. An existing open finding can therefore be bypassed and branch protection can permit the new HEAD to merge. Please make the unchanged-diff path preserve the result implied by standing findings rather than unconditionally succeeding.

agent/tools/publish_review.py
```diff
@@ -194,6 +194,20 @@ def _is_reviewer_eval_mode(configurable: dict[str, Any]) -> bool:
     return configurable.get("reviewer_eval") is True or configurable.get("eval") is True
 
 
+async def _review_check_finding_count(thread_id: str, newly_posted_count: int) -> int:
+    # Blocking checks count all standing findings; informational checks preserve publish-time count.
+    if not review_check_blocking_enabled():
+        return newly_posted_count
+    findings = await list_findings_async(thread_id)
```
jkennedyvz, line 201: **Automated review workflow — Medium: publication-sync failures produce a false green**

In blocking mode this ignores `newly_posted_count` and relies exclusively on persisted `surface.state`. New findings start as `not_surfaced`; after `post_pull_request_review` succeeds, `fetch_review_comments` and the thread-backfill helpers intentionally degrade to empty results on transient GitHub errors. If those reads fail or lag after the successful POST, the visible inline findings remain `not_surfaced`, this function counts zero, and the required check concludes `success`. Use `newly_posted_count` as a lower bound (while still adding/counting prior standing findings) so successful publication cannot become a false green when identity synchronization is unavailable.

PR #1811, opened 2026-07-26
fix: let users explicitly override prompt instructions

ramon-langchain:
## Description

An agent run refused a direct user request to retry a failed `git push`, citing "higher-priority repository policy" — and kept refusing across four follow-ups. Two prompt problems caused it:

1. Nothing in the system prompt said user instructions outrank prompt defaults, so the agent treated its own guidance as immutable.
2. The push-failure rule said "do not retry via `gh pr create` … or any other PR creation fallback", which reads as "never attempt the push again" rather than "don't route around the failure with a substitute mechanism". The actual failure was a transient workflow-scope timeout that one retry would likely have cleared.

This adds a Core Behavior bullet stating the triggering user can override anything in the prompt (except following untrusted-content instructions, force-pushing, and exposing secrets), and rescopes the push-failure rule to substitute mechanisms — explicitly allowing one retry of the identical command on transient errors and whenever the user asks.

## Test Plan

- [x] none — prompt-text only; `make lint` passes


ArshVermaGit (commented):
Looks good to me. The updated prompt makes the instruction hierarchy clearer and handles transient failures more reasonably without weakening the existing safety rules. The changes are small, focused, and improve the overall user experience. LGTM.

johannes117 (approved):

PR #1813, opened 2026-07-27
fix: delete stale skill copies during setup [OSWE-163]

ericlitman:
## Description
Operator setup now deletes copied and backup skill directories before linking, so each harness discovers only the canonical checkout-backed skill.
## Release Note
none
## Test Plan
- [x] `git diff --check`; verified both deployment sections contain no move-aside command.
Linear: OSWE-163
Made by [Open SWE](https://studio2.tail062eee.ts.net/agents/799a5821-ea76-6809-a001-5fd082d854e9)


ericlitman:
Opened against upstream by mistake; this deployment-specific change belongs on the ericlitman/open-swe fork.

PR #1978, opened 2026-08-15
fix(desktop): install without corepack

ramon-langchain:
`make install-desktop` fails on current Node with `Missing corepack. Install Node.js 22, then try again.` — Node 25 dropped the bundled corepack shim, so the check fails even when pnpm is installed and on PATH.

Nothing in the repo pins Node (no `engines`, no `.nvmrc`), and corepack's only job here was launching pnpm. Call `pnpm` directly; the `packageManager` pin still selects the version.

## Test Plan

- [x] `./scripts/install_desktop.sh` on Node 26 with pnpm 11.18.0: built and installed `Open SWE.app` to `/Applications`

### DIFF
--- a/scripts/install_desktop.sh
+++ b/scripts/install_desktop.sh
@@ -6,9 +6,9 @@ if [[ "$(uname -s)" != "Darwin" ]]; then
   exit 1
 fi
 
-for command in node corepack ditto; do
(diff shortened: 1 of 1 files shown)


ramon-langchain:
Folded into #1977 so both desktop changes ship together.

PR #2000, opened 2026-08-17
feat(desktop): pin dcode and add Electron E2E

ramon-langchain:
## Summary

- upgrade the external Deep Agents SDK dependency to `deepagents==0.7.6`
- launch Desktop ACP through isolated, pinned uv tooling: `deepagents-code[fireworks]==0.1.56` on Python 3.13, never a `dcode` from `PATH`
- add real Electron Playwright coverage that reuses the existing fake LLM, git remote, and fake GitHub API, and run it in CI under Xvfb

## Test plan

- `make test`
- `make lint`
- `make typecheck`
- `pnpm --dir desktop run check`
- `npm --prefix tests/e2e run test:desktop`
- existing browser full-flow happy path
- Codex autoreview: clean


desktop/test/acp-client.test.cjs
```diff
@@ -11,51 +8,49 @@ const {
...
-  assert.deepEqual(target, {
-    command,
-    args: ["--acp"],
+test("runs ACP from the pinned uv package", () => {
+  const launcherArgs = [
+    "tool",
+    "run",
+    "--isolated",
+    "--python",
+    "3.13",
```
ramon-langchain: why repeat this arg list in tests?

desktop/test/acp-client.test.cjs
```diff
@@ -11,51 +8,49 @@ const {
...
-  const target = dcodeTarget({ env: { HOME: home }, platform: "darwin" })
-  assert.deepEqual(target, {
-    command,
-    args: ["--acp"],
+test("runs ACP from the pinned uv package", () => {
+  const launcherArgs = [
+    "tool",
+    "run",
+    "--isolated",
+    "--python",
```
ramon-langchain: why use old python?

tests/e2e/fake_llm.py
```diff
@@ -64,6 +65,34 @@ def farewell(name):
...
+        "head": FEATURE_BRANCH,
+        "base": BASE_BRANCH,
+        "title": PR_TITLE,
+        "body": "Adds a `greet()` helper as requested.",
+        "draft": True,
+    }
+)
+_DESKTOP_IMPLEMENT_SCRIPT = f"""
+set -e
+git config user.email "dev@example.com"
```
ramon-langchain: whats going on here? do not mess with the users git config please

desktop/src/acp-client.cts
```diff
@@ -1,52 +1,64 @@
...
-  if (installedCommand && fs.existsSync(installedCommand)) {
-    return { command: installedCommand, args }
+  const launcherArgs = [
+    "tool",
+    "run",
+    "--isolated",
+    "--python",
+    "3.13",
+    "--from",
+    "deepagents-code[fireworks]==0.1.56",
```
ramon-langchain, line 32: I do not understand why we need to specify a version here as well as in our version lock for deepagents

desktop/test/acp-client.test.cjs
ramon-langchain: Updated in 8deeb30d: the isolated uv environment now uses Python 3.14. — Claude

tests/e2e/fake_llm.py
ramon-langchain: Updated in 8deeb30d: the E2E script no longer writes git config at any scope. The fixed test identity is passed only to the single commit via `git -c user.email=... -c user.name=... commit`. — Claude

desktop/src/acp-client.cts
ramon-langchain: `0.1.56` is the `deepagents-code` distribution version, while `0.7.6` is the Deep Agents SDK version. They are a compatible pair: dcode 0.1.56 itself pins SDK 0.7.6. In 8deeb30d the isolated dcode environment also explicitly applies `--with deepagents==0.7.6`, matching the root project, so the same SDK version is used across the repo. — Claude

desktop/test/acp-client.test.cjs
ramon-langchain: Updated in 8deeb30d: the test now validates that the final args are composed from `target.launcherArgs` plus `--acp`, and checks only the launcher invariants. The full argument list is no longer repeated in the test. — Claude

PR #2069, opened 2026-08-19
feat: share sandbox files with LangSmith download URLs

ramon-langchain:
## Description
Upgrade LangSmith to 0.11.1 and expose its signed sandbox download URLs through the native async client with bounded defaults. Enable the tool and large-artifact guidance only for LangSmith sandbox deployments.

## Release Note
Agents can now share large sandbox files such as videos and images through expiring download URLs.

## Test Plan
- [x] Verify async URL generation, SDK option forwarding, input validation, and provider-gated agent assembly.

Made by [Open SWE](https://openswe.vercel.app/agents/7424a24b-bc20-516d-9e88-3c4d1c65d4b0)


agent/tools/create_sandbox_file_download_url.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import Any, Literal
+
+from langchain_core.tools import tool
+from langgraph.config import get_config
+
+from ..integrations.langsmith import get_async_sandbox_client
+from ..utils.sandbox_paths import aresolve_sandbox_work_dir
+from ..utils.sandbox_state import get_sandbox_backend, unwrap_sandbox_backend
+
+_DEFAULT_EXPIRY_SECONDS = 24 * 60 * 60
```
ramon-langchain: change this to never expires

agent/tools/create_sandbox_file_download_url.py
```diff
@@ -0,0 +1,73 @@
...
+    if not download.download_url:
+        raise RuntimeError("LangSmith did not return a download URL")
+    return {
+        "url": download.download_url,
+        "file_path": path,
+        "expires_at": download.expires_at,
+    }
+
+
+create_sandbox_file_download_url = tool(
```
ramon-langchain: hmm dont we have a declarative way to do this?

agent/tools/create_sandbox_file_download_url.py
ramon-langchain: Done in 6e7d398e — `expires_in_seconds` now defaults to `None`, so links never expire unless a caller deliberately asks for a window. Dropped `_DEFAULT_EXPIRY_SECONDS` and reworded the description accordingly; the defaults test now asserts `expires_in_seconds: None` reaches the SDK.

— Claude

agent/tools/create_sandbox_file_download_url.py
ramon-langchain: Yes — and it now uses it. Every other tool in `agent/tools/` is just a plain async function whose docstring becomes the description; the imperative `tool(name, description=...)(fn)` form is only needed by `output_iframe`, which has to declare `response_format="content_and_artifact"`. This tool returns a plain dict, so in 6e7d398e it is a bare `async def create_sandbox_file_download_url(...)` with the description as its docstring, and the private `_`-prefixed indirection is gone.

— Claude

agent/server.py
```diff
@@ -1176,6 +1189,7 @@ async def _prepare(self, state: PrepareRunState, runtime: Runtime) -> dict[str,
                 admin_environments=self._admin_environments,
                 source=self._source,
                 slack_context=_slack_tools_enabled(configurable),
+                sandbox_file_downloads=_sandbox_file_downloads_enabled(configurable),
```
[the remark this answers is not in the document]
ramon-langchain: Fixed in 6e7d398e by taking the second option you suggested: `_sandbox_file_downloads_enabled` now also excludes desktop runs (`and not is_desktop_run(resolved)`), so the middleware's prompt rendering and `get_agent`'s toolset compute availability from the same predicate. The now-redundant `and not local_run` at the `get_agent` call site is gone, which removes the chance of the two drifting apart again.

Worth noting the guidance was not actually reaching desktop prompts even before this: `PrepareAgentRunMiddleware._prepare` returns early for `is_desktop_run` and calls `construct_system_prompt(..., source="desktop")` without `sandbox_file_downloads`, which defaults to `False`. So this is a latent-trap fix rather than a live bug.

— Claude

agent/tools/create_sandbox_file_download_url.py
```diff
@@ -0,0 +1,65 @@
...
+    config = get_config()
+    configurable = config.get("configurable", {}) if isinstance(config, dict) else {}
+    thread_id = configurable.get("thread_id") if isinstance(configurable, dict) else None
+    if not isinstance(thread_id, str) or not thread_id:
+        raise ValueError("no thread_id in run config")
+
+    backend_proxy = await get_sandbox_backend(thread_id)
+    work_dir = await aresolve_sandbox_work_dir(backend_proxy)
+    path = posixpath.normpath(
+        file_path.strip()
```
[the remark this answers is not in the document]
johannes117: Valid finding, addressed in a76d2989. Download paths are now restricted to the resolved sandbox work directory before URL generation. The tool also resolves the existing file with `realpath` and repeats containment validation, preventing both `..`/absolute-path escapes and symlink escapes. Focused tests cover traversal, sibling-prefix paths, absolute paths, and an out-of-workdir symlink.

johannes117 (approved):

PR #2094, opened 2026-08-20
feat: use C for new threads

langchain-infra:
## Description
Match Linear by using bare `C` to open a new thread in both the web and desktop apps, while preserving typing-context guards. Remove the desktop-only `Cmd/Ctrl+N` accelerator so the shortcut is consistent across surfaces.
## Release Note
The New Thread shortcut is now `C` in web and desktop clients.
## Test Plan
- [x] UI shortcut/hotkey tests, typecheck, and production build
- [x] Desktop build, tests, and typecheck
- [ ] UI lint unavailable because the repo does not install the configured `eslint` executable
Made by [Open SWE](https://openswe.vercel.app/agents/82cc3971-f1c0-5061-877c-6a21ef5af75a)


langchain-infra:
@open-swe fix ci

ramon-langchain (approved):

PR #2161, opened 2026-08-21
feat(ui): replace desktop sidebar heading with project selector

johannes117:
## Description
The desktop sidebar's static "Projects and threads" label is replaced with a project selector menu (All projects or a single project) that filters the local thread list, with the add-project button on the same line.

## Release Note
Desktop sidebar now lets you scope the thread list to a single local project.

## Test Plan
- [ ] In the desktop app, open the sidebar's This Mac view, pick a project from the selector and confirm only that project's threads show; switch back to All projects.
- [ ] Remove a project from the selector menu while it is selected and confirm the list falls back to All projects.


johannes117:
Superseded: change moved onto open-swe/source-toggle-polish (#2145).

PR #2222, opened 2026-08-25
fix: wait out a resuming sandbox instead of calling it unreachable

ramon-langchain:
## Description
Wait through transient 409s while a sandbox resumes instead of terminating the run. Provisioning retries use a 30-second monotonic deadline and cap `Retry-After` sleeps to the remaining budget.

## Release Note
Sandbox-backed runs now recover when proxy configuration races sandbox provisioning.

## Test Plan
- [x] Focused provisioning retry tests

Made by [Open SWE](https://openswe.vercel.app/agents/ca1d3c3d-6053-5df2-a7d5-39b95edcd6d9)

### DIFF
--- a/agent/integrations/langsmith.py
+++ b/agent/integrations/langsmith.py
@@ -362,7 +362,9 @@ async def _patch_proxy_config(
     sandbox_name: str,
 ) -> None:
     attempt = 0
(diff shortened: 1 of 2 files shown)


ramon-langchain:
Fixed in sandboxes

PR #2244, opened 2026-08-26
refactor(repo-snapshots): migrate to TypedStore

ramon-langchain:
Stacked on #2243 — review that first, and merge it first. This PR targets that branch.

Fourth namespace onto `TypedStore`. `environments` is now the only `KeyedRecordStore` user left; once it migrates, `KeyedRecordStore` deletes.

## The untyped patch API here was `_set_status`

```python
await _set_status(full_name, "ready", status_message=None, extra={
    "snapshot_id": snapshot_id, "snapshot_name": snapshot_name,
    "build_log": log_tail, "last_built_at": now_iso(),
})
```

Nothing checked those keys. It's replaced by `mark_building` / `mark_ready` / `mark_failed` on `RepoSnapshotStore`, so each transition names the fields it touches in its signature.

## Dead defensive code

`_build_snapshot_sync` had six reads shaped like `int(record.get("vcpus") or DEFAULT_BUILD_VCPUS)`. The model carries those defaults, and `RepoSnapshotUpdate`'s range validators (`_MIN_VCPUS <= v <= _MAX_VCPUS`, etc.) are the only write path, so a zero or missing value can't reach it.

Also gone: `is_repo_snapshot_build_stale(record)` becomes the `build_is_stale` property, and `delete_repo_snapshot`'s `bool` return — the one caller did its own existence check and discarded the result.

## Tests

Moved onto the shared `fake_store` fixture from #2243, so they run the real `model_dump`/`model_validate` path instead of stubbing store functions out. That reached five cases the previous mocks couldn't:

- `create` is idempotent
- sizing fields survive an update that omits them (the `if value is not None` branches had no coverage)
- `mark_building` on a missing record raises
- the empty-dockerfile 400
- both build outcomes asserted against the **stored record** rather than captured `_set_status` calls — the old tests asserted on the arguments to a mock, so they'd have passed even if the write never happened

The two `create_langsmith_sandbox` tests at the bottom of the file are unchanged.

## Test Plan

- [x] `make typecheck` — 0 errors
- [x] `make lint` — clean
- [x] `make test` — 1913 passed, 0 failed

## Note on the flaky Slack webhook tests

#2243's description called the 7 `tests/github/test_github_issue_webhook.py` failures pre-existing. That was right about them not being ours, but imprecise: they're **order-dependent, not deterministic** — two full runs on this branch passed all 1913.

Cause looks like `agent/utils/slack_events.py`, which keeps `_claimed_keys` as a process-global `OrderedDict` with no TTL and 2048-entry LRU eviction. It ships a `reset_slack_event_claims()` helper that nothing calls, and `tests/conftest.py` only resets `ttl_cache`. So whether a webhook test sees a stale claim depends on what ran before it. Not fixed here — out of scope — but it's a small autouse-fixture fix for whoever wants it.

### DIFF
--- a/agent/baby_sit.py
+++ b/agent/baby_sit.py
@@ -5,12 +5,14 @@ import logging
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from datetime import UTC, datetime, timedelta
-from typing import Any, TypedDict, cast
+from typing import Any
 
 from langgraph_sdk import get_client
 from langgraph_sdk.errors import ConflictError
+from pydantic import BaseModel, ConfigDict, Field
 
-from agent.store import delete_value, get_value, now_iso, put_value, search_all_values
+from agent.source_context import SourceContext
+from agent.store import TypedStore, now_iso
 from agent.thread_ids import baby_sit_lock_thread_id
 
 from .dispatch import dispatch_agent_run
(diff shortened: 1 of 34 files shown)


ramon-langchain:
Superseded by #2248, which removed per-repo snapshots entirely. Nothing here to migrate any more.

The `TypedStore` work continues in #2243. `KeyedRecordStore` now has one user left (`environments.py`) instead of two, so it's one PR from deleting.

PR #2246, opened 2026-08-26
feat: mirror every agent thread into a Slack firehose channel

ramon-langchain:
Set a channel under **Admin → Slack Firehose Channel** (experimental, blank = off) and every agent thread is duplicated into it as a read-only thread.

Each thread gets a root message — title linking to the web UI, source, repo, requester — and under it the inbound request plus every block of prose the agent produces, so the channel reads like the web transcript rather than just the messages the agent chose to send.

Tool calls do not get a message each. They collapse into one rolling `task_card` that updates in place and settles to `complete` when the agent next speaks or the run ends. [`task_card`](https://docs.slack.dev/reference/block-kit/blocks/task-card-block/) is Slack's native block for showing an agent's thinking steps; a workspace that rejects it falls back permanently to a Markdown card.

### How it hooks in

A middleware in the agent graph, so Slack, Linear, GitHub, web and scheduled threads are all covered without touching any of their existing paths. Nothing about current behaviour changes when no channel is configured.

- Desktop ("This Mac") and internal stop-summary runs are excluded.
- Mirrored writes are serialized per thread and fired off the critical path — a Slack failure can never fail a run.
- The root message ts lives in thread metadata, so a restarted process reuses the same Slack thread instead of opening a second one.
- `input_message_kind()` is new; the firehose mirrors only envelopes with `kind="human"`, which keeps Open SWE's own prompt scaffolding (framing blocks, sender metadata, dynamic context) out of Slack.

### Operator notes

Invite the bot to the channel first, or every post fails with `not_in_channel`. Replies in a firehose thread go nowhere, but an `@Open SWE` mention there starts a brand new thread — the firehose thread is not bound to the original.

## Test Plan

- [x] `make lint`, `make typecheck`, `pnpm typecheck`
- [x] 1896 unit tests pass, including 17 new ones across `tests/slack/test_slack_firehose.py`, `tests/middleware/test_firehose_middleware.py`, `tests/dashboard/test_team_firehose_setting.py`
- [x] New E2E `tests/e2e/tests/slack_firehose.spec.ts` (2/2): drives a real Slack mention through the real webhook and agent, asserts the firehose channel gets the root message, the request, the agent's prose and settled `task_card` replies — and asserts zero Slack traffic when no channel is configured
- [x] Full E2E suite: 53 passed / 3 failed. The same suite with this middleware unwired fails 3 too, two of them identical (`dashboard.spec.ts:690`, `threads_workspace.spec.ts:735`); the third differs between runs, so the failing set is pre-existing flakiness rather than a regression


johannes117:
I think this will just be superceded by the new code channels PR: https://github.com/langchain-ai/open-swe/pull/2252

johannes117:
superceded by slack code channels

PR #2294, opened 2026-08-28
feat(model): allow OPENAI_USE_RESPONSES_API escape hatch for Chat Completions hosts (#2292)

### LINKED ISSUE #2292: docs: openai: models force Responses; Chat Completions OPENAI_BASE_URL hosts have no escape hatch
`make_model` already reads `OPENAI_BASE_URL` / `OPENAI_API_BASE` for `openai:` model ids ([INSTALLATION.md](https://github.com/langchain-ai/open-swe/blob/main/docs/INSTALLATION.md), `agent/utils/model.py`). Two details are easy to get wrong on a Chat Completions-first host:

1. For every `openai:` id, `make_model` sets `use_responses_api = True`. If `OPENAI_BASE_URL` is unset it falls back to `OPENAI_RESPONSES_WS_BASE_URL` (`wss://api.openai.com/v1`). Gateway overrides can flip `use_responses_api`; the direct env path cannot.
2. `_coerce_openai_chat_completions_kwargs` already exists when `use_responses_api is False`, so Chat Completions is a supported wire, just not reachable from `OPENAI_BASE_URL` alone.

PZERO is a live OpenAI-compatible host. Catalog: `GET https://api.pzero.studio/v1/models`. Default text id: `deepseek-v4-flash`. Prepaid Bearer key (`pzero_…`). `POST /v1/responses` exists (401 without a key, not 404) but Chat Completions is the working path for agent tools.

```bash
export OPENAI_BASE_URL="https://api.pzero.studio/v1"
export OPENAI_API_KEY="pzero_…"
# LLM_MODEL_ID / dashboard model: openai:deepseek-v4-flash
```

That env pair still forces Responses today. Do not treat the `wss://` fallback as an HTTP base. Do not file a named first-party provider (open OrcaRouter/MiniMax threads already cover that).

If a named vendor subsection is too much, the escape hatch is the docs/code gap: let `OPENAI_BASE_URL` hosts set `use_responses_api=false` (env or profile) instead of always True.

teddiesloco:
### Summary

Fixes #2292.

For `openai:` model identifiers, `make_model()` previously hardcoded `use_responses_api = True`. When pointing `OPENAI_BASE_URL` to OpenAI-compatible hosts that serve standard Chat Completions (`POST /v1/chat/completions`) rather than WebSocket/Responses API, callers were forced onto Responses wire unless routing through LangSmith Gateway.

### Changes
- Check `OPENAI_USE_RESPONSES_API` environment variable (`0`, `false`, `off`, `no`).
- Use `model_kwargs.setdefault("use_responses_api", True)` so caller-provided kwargs or environment flags can toggle Chat Completions mode smoothly.

### DIFF
--- a/agent/utils/model.py
+++ b/agent/utils/model.py
@@ -147,7 +147,12 @@ def make_model(model_id: str, *, use_gateway: bool | None = None, **kwargs: Unpa
             or os.environ.get("OPENAI_API_BASE")
             or OPENAI_RESPONSES_WS_BASE_URL
         )
-        model_kwargs["use_responses_api"] = True
+        # Allow OPENAI_USE_RESPONSES_API=false or caller kwargs to override default Responses wire
+        env_use_responses = os.environ.get("OPENAI_USE_RESPONSES_API")
+        if env_use_responses is not None:
+            model_kwargs["use_responses_api"] = env_use_responses.lower() not in ("0", "false", "no", "off")
+        else:
+            model_kwargs.setdefault("use_responses_api", True)
 
     enabled = gateway_env_default() if use_gateway is None else use_gateway
     gateway_applied = False


ArshVermaGit (commented):
Looks good to me. The change is focused and correctly handles the environment variable override. Approved.

PR #2330, opened 2026-08-30
fix(ui): deduplicate sidebar show more action

johannes117:
Render one Recents “Show more” action and load every paginated thread source that still has results.

Made by [Open SWE](https://openswe.vercel.app/agents/01a053ab-f3d1-76b6-8582-37a7d78757e6)


johannes117:
Consolidated into #2334.

PR #2331, opened 2026-08-30
fix: prefer new PRs for local threads

johannes117:
Make local threads default to their own task branch and new PR, rather than attaching work to a similar PR from another thread. Existing PRs remain valid when explicitly targeted or already tied to the thread's branch.

No tests added; this is prompt-only guidance and static prompt assertions are intentionally avoided.

Made by [Open SWE](https://openswe.vercel.app/agents/01a053bc-96ff-74a3-99d2-f7ddcc15dac2)


johannes117:
Consolidated into #2334.

PR #2332, opened 2026-08-30
fix: share sidebar state across threads

johannes117:
Share one sidebar query across thread routes, fetching the selected thread separately only when it falls outside the shared page. Refresh both sources on mount and window focus.

Validation:
- `pnpm --dir ui exec vitest run src/features/agents/lib/queries.test.tsx --no-color`
- focused `oxfmt --check` and `oxlint --deny-warnings`
- `pnpm --dir ui run typecheck`

Made by [Open SWE](https://openswe.vercel.app/agents/01a053c0-ba8b-71da-8861-1118de64d59a)


johannes117:
Consolidated into #2334.

PR #2333, opened 2026-08-30
fix: preserve new thread stream during navigation

johannes117:
Preserve the layout stream when opening a newly created thread, so its messages continue rendering after navigation. Stabilize the affected onboarding and desktop E2E selectors.

Checks:
- UI unit tests
- UI typecheck
- Focused format and lint

Made by [Open SWE](https://openswe.vercel.app/agents/01a053b7-c5be-743d-b83d-15958250017f)


johannes117:
Consolidated into #2334.

PR #2343, opened 2026-08-31
perf(ui): resume thread streams instead of replaying them

johannes117:
Switching between threads was slow, flickered, and replayed the whole run — re-animating tool calls the user had already watched.

## The problem

Four separate causes, all on the thread lifecycle:

1. **Replay.** The SDK's `joinStream` defaults `lastEventId` to `"-1"`, and its internal last-event tracking lives only inside a single `joinStream` call. Every rejoin — remount, switching back, a dropped stream — therefore re-delivered the run from event 0.
2. **Blank pane.** The detail endpoint set `summary["messages"] = []` and left the transcript entirely to the SDK's client-side hydrate, so an opened thread showed nothing until `getState` returned. `RecentAgentThreads` worked around this by keeping three thread trees mounted, at the cost of a live SSE connection and a hidden DOM tree each.
3. **Sidebar churn.** `useSidebarThreads` refetched the whole list every 2s while anything ran, replacing every row's identity and re-rendering the list.
4. **Reordering.** The sidebar sorted on `updated_at`, which a running agent bumps continuously, so rows shuffled under the pointer for as long as the agent worked.

## The changes

**Resume instead of replay** (`ui/src/features/agents/lib/streamResume.ts`) — wraps the client so the initial run stream and every rejoin record the ids they see, and a rejoin resumes from the last one. The SDK sends it as a `Last-Event-ID` header; a server that ignores it behaves exactly as before, so this can only narrow the payload.

**Transcript snapshot** — the detail endpoint returns the last 10 human-anchored turns, so an opened thread paints immediately. Cutting on human anchors means no tool result outlives its call. It's rendered through the existing `streamMessagesToUi` pipeline rather than converted server-side, so there's one definition of how a message becomes a chunk and the snapshot inherits every fix made for the live path.

The client asks for it once per open and opts out afterwards (`include_transcript=false`), because this endpoint doubles as the run-status heartbeat while a run is live — a snapshot on every beat would be a `get_state` every three seconds.

**Cache and mount less** — thread details are retained (`gcTime: Infinity`) and an opening thread is seeded from its sidebar row, so returning to a thread paints from cache. Only the active thread is mounted now. Each thread still gets its own stream provider rather than re-pointing a shared one: the SDK's thread switch clears the transcript *and cancels whatever runs the user had queued on the thread being left*, so a follow-up typed during a run would vanish on navigation.

**Targeted status polling** — a batch `POST /threads/statuses` returns only the fields that move while a run is in flight. The sidebar polls just the running rows and patches them into the cached lists, building a new object only where something actually changed so untouched rows keep their identity. Since the ids come from the caller, each thread's readability is checked individually, and the id list is capped. A run *ending* can add or remove rows, which a status patch can't express, so that edge reconciles once with a full invalidate.

**Stable sort anchor** — the sidebar now orders on `max(created_at, last_user_message_at)`. A row moves for what the user did, not for every write the run makes afterwards.

## Testing

- 311 UI tests and the Python suite pass. `ruff`, `basedpyright`, `oxlint` and `oxfmt` clean.
- New coverage: the resume layer (including the no-options rejoin path, which is the one `tryReconnect` actually uses), transcript windowing and conversion, the stable anchor, the transcript opt-out, and per-thread authz on the status endpoint.
- Pre-existing on `main`, unrelated to this change: `tests/dashboard/test_voice.py::test_transcribe_audio_validates_and_forwards`, two `tests/sandbox/test_proxy_auth.py` cases, and some react-icons `className` type errors.

Not yet verified in a running app — worth exercising thread switching during a live run, and confirming a rejoin no longer replays.

### DIFF
--- a/agent/dashboard/routes.py
+++ b/agent/dashboard/routes.py
@@ -206,6 +206,7 @@ from .thread_api import (
     get_dashboard_thread_pull_request_status,
     get_dashboard_thread_recovery_patch,
     get_dashboard_thread_state,
+    get_dashboard_thread_statuses,
     get_dashboard_thread_working_tree_diff,
     list_dashboard_threads,
     list_dashboard_threads_page,
@@ -1944,6 +1945,18 @@ async def api_get_pull_request_checks(
     )
 
 
+class ThreadStatusesRequest(BaseModel):
+    threadIds: list[str] = Field(default_factory=list, max_length=200)
+
+
+@router.post("/threads/statuses")
+async def api_get_thread_statuses(
+    payload: ThreadStatusesRequest,
+    session: dict[str, Any] = _SESSION_DEP,
+) -> dict[str, Any]:
+    return await get_dashboard_thread_statuses(payload.threadIds)
+
+
 @router.get("/threads/{thread_id}/pull-request-status")
 async def api_get_thread_pull_request_status(
     thread_id: str,
@@ -1976,6 +1989,7 @@ async def api_get_thread_pull_request_context(
 async def api_get_thread(
     thread_id: str,
     mark_viewed: bool = True,
+    include_transcript: bool = True,
     session: dict[str, Any] = _SESSION_DEP,
 ) -> dict[str, Any]:
     return await get_dashboard_thread(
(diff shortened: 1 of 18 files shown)


johannes117:
Superseded by #2345, which rebuilds the thread runtime lifecycle from current main around keyed LangGraph v2 controllers instead of extending the legacy resume and hidden-mount wiring.

PR #2402, opened 2026-09-03
chore: require Python 3.14 as the minimum version

ramon-langchain:
Raises every declared Python floor in the repo to 3.14.

- `pyproject.toml`: `requires-python = ">=3.14"`, ruff `target-version = "py314"`, basedpyright `pythonVersion = "3.14"`
- `langgraph.json`, `langgraph.desktop.json`, `tests/e2e/langgraph.{desktop,e2e}.json`: `python_version` 3.12 → 3.14
- `Dockerfile`: base image → `langchain/langgraph-api:0.13.2-py3.14`
- `desktop/scripts/build-local-backend.cjs`: default bundled runtime → 3.14
- `docs/INSTALLATION.md`: the prerequisite said 3.14 was unsupported due to dependency constraints — it resolves cleanly, so it now reads `Python 3.14+`
- `agent/github/pull_request_checks.py`: the `typing_extensions.TypedDict` shim existed only because pydantic rejected a stdlib `TypedDict` in a response model below 3.12; it is now the stdlib import

### Lint churn

The py314 ruff target activates pyupgrade rules that were inert at py311, so most of the file count here is their autofix: unquoted forward references (UP037), PEP 695 `type` aliases and generic parameters (UP040/UP046/UP047), and `ruff format` rewriting `except (A, B):` as PEP 758 `except A, B:` in 15 files. No behavior change.

### langgraph-api 0.12.6 → 0.13.3

The first cloud build failed pulling `langgraph-api-unlicensed:0.12.6-py3.14` — 0.12.6 only publishes py3.11 through py3.13, and py3.14 images start at 0.13.0. So the three `langgraph*.json` manifests, the root `Dockerfile` base tag, and the `constraint-dependencies` floor in `pyproject.toml` all move to 0.13.3 (latest) together, keeping local dev on the same runtime the deployment serves. That pulls `langgraph-runtime-inmem` 0.32.6 → 0.33.3 and the opentelemetry/grpcio stacks forward.

### Dependency movement

Relocking under 3.14 moves **daytona 0.26.0 → 0.140.0** (0.26 publishes no 3.14 wheels) and websockets 16.1.1 → 15.0.1. The daytona jump is the one thing in this PR worth a careful look — it is a large API-surface change behind `agent/sandboxes/providers/daytona.py`.

## Test Plan

- [x] `make lint` — ruff check and `ruff format --diff` both clean across 500 files
- [x] `make typecheck` — basedpyright, 0 errors
- [x] `uv lock` resolves on CPython 3.14.6; `uv sync --extra dev` installs
- [x] All three graph modules plus `agent.webapp` import under 3.14, and `SANDBOX_FACTORIES` registers every provider
- [x] `0.13.3-py3.14` confirmed to exist on Docker Hub before pinning it
- [x] All four `langgraph*.json` manifests pass `langgraph_cli.config.validate_config`


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "open-swe-agent"
 version = "0.1.0"
 description = "Open SWE Agent - Python agent for automating software engineering tasks"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.14"
 license = { text = "MIT" }
 dependencies = [
     "deepagents==0.7.6",
```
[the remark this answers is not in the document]
ramon-langchain: Not a change in this PR — the comment is anchored on an unchanged context line.

`deepagents` went 0.7.11 → 0.7.6 on `main` in ad29b0f08 ("chore: update Python and JavaScript dependencies", #2156), which is this PR's merge base. Verified:

```
$ git diff $(git merge-base origin/main HEAD) HEAD -- pyproject.toml | grep deepagents
(no output)
```

The only pyproject line this PR touches in that hunk is `requires-python`. The `0.7.6` you see at line 9 arrived via the `origin/main` merge in b252be0, so re-pinning 0.7.11 here would revert a main-side dependency decision from inside a Python-version bump — wrong PR for it.

The underlying point about `TracePolicy(process_inputs=omit_payload)` on `FilesystemMiddleware` / `SubAgentMiddleware` / `SkillsMiddleware` looks worth chasing, just against `main` rather than here. Flagged to the author.

— Claude

AGENTS.md
```diff
@@ -10,7 +10,7 @@ A separate **reviewer** graph runs read-only code reviews on PRs, and a **review
 
 ## Commands
 
-Dependencies are managed with **uv**. Tests use pytest (`asyncio_mode = "auto"`). Lint/format is **ruff** (line-length 100, target py311). Type checking is **basedpyright** (`typeCheckingMode = "standard"`). `requires-python = ">=3.11"`; `langgraph.json` pins the runtime to 3.12.
+Dependencies are managed with **uv**. Tests use pytest (`asyncio_mode = "auto"`). Lint/format is **ruff** (line-length 100, target py314). Type checking is **basedpyright** (`typeCheckingMode = "standard"`). `requires-python = ">=3.14"`; `langgraph.json` pins the runtime to 3.14.
```
mdrxy, line 13: feel like this file is prob way more verbose than needed

pyproject.toml
```diff
@@ -68,7 +68,7 @@ packages = ["agent"]
 
 [tool.ruff]
 line-length = 100
-target-version = "py311"
+target-version = "py314"
```
mdrxy, line 71: nit: this line can be dropped (ruff infers from `requires-python` min)

mdrxy (approved):

PR #2409, opened 2026-09-03
fix(ui): forward backend responses instead of stringifying them

ramon-langchain:
Dev is down: every `/dashboard/api/*` call answers `200 text/plain` with the body `[object Object]`, so all dashboard queries fail to parse and the sidebar shows "Cloud threads unavailable".

## Cause

`ui/server/backend-proxy.ts` returned h3's `proxyRequest` result. The server runtime bundles its own copy of h3, and a proxy result from a different copy is not a shape it recognises — it coerced it with `String()`, replacing the backend's `application/json` with `text/plain` and flattening every non-200 to 200.

`ad29b0f08` (#2156, merged today) moved the direct `h3` dep from `2.0.1-rc.22` to `2.0.1-rc.29`; nitro still pins rc.22. Before that both were rc.22, so the shapes matched.

The handler now builds a plain web `Response` from `fetch` — status, headers, streamed body, `set-cookie` lines, and `redirect: "manual"` for the OAuth hops all preserved. `h3` is dropped from `ui/package.json`, so no direct copy remains to skew against.

## Why CI was green

The E2E suite ran a topology no environment runs. `global-setup.ts` baked the harness in as the client's API base, so the browser called the backend directly and `/dashboard/api/*` never touched the app server — the proxy had zero coverage. `dev-mock.sh` already had the correct same-origin shape.

The specs now drive the app server origin with an empty client API base, so every dashboard call crosses the real proxy the way a deployed browser does. With the proxy broken, 5 of 5 completed `dashboard.spec.ts` specs failed on `SyntaxError: "[object Object]" is not valid JSON` — the exact error from dev. No spec changes were needed. The harness routes that existed only to serve the app from the wrong origin are deleted (~110 lines).

## Test Plan

- [x] `dashboard.spec.ts` + `ssr.spec.ts`: 13 passed (41.2s), red before the fix and green after
- [x] Verified against the real production build with a stub backend: `content-type: application/json`, was `text/plain` + `[object Object]`
- [x] `make lint`

[no screenshot]


ramon-langchain:
Re: the Open SWE review (no issues found) — acknowledged, nothing to act on there.

Since that review ran, `d157636` landed one more change: `/threads` and `/store` are now fronted by the app server in E2E runs. They are LangGraph's own API, served by the mock harness alongside the fake-SaaS routes; once the specs moved onto the app-server origin those calls were hitting the app router and coming back as its 404 page. The app router has no route under either prefix, and a deployed build still fronts only `/dashboard/api` and `/webhooks`.

Full browser suite is green locally (57 passed, 1 skipped) and all three CI shards pass.

— Claude

langchain-infra (approved):

PR #2410, opened 2026-09-03
feat: simplify installation to a GitHub + Slack minimum with runtime discovery

langchain-infra:
## Summary

Open SWE now installs with seven variables: `LANGSMITH_API_KEY`, one model provider key (or `LANGSMITH_GATEWAY_ENABLED=true`), `GITHUB_APP_CLIENT_ID`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_WEBHOOK_SECRET`, `SLACK_BOT_TOKEN`, and `SLACK_SIGNING_SECRET`. Everything else is discovered at runtime, has a safe default, or is an optional add-on.

**Runtime discovery** (an explicit env value always wins):

- **GitHub App.** `GITHUB_APP_CLIENT_ID` is the JWT `iss` (GitHub's documented recommendation); `GITHUB_APP_ID` is a deprecated fallback. The installation is resolved from the repository or organization of the run, else the app's only installation. Several installations with no context skip token minting and log the account names. Lookups are cached in process.
- **Slack.** The bot user id and handle come from `auth.test` + `users.info`, fetched at startup and lazily on the first webhook, throttled to once a minute. Setting `SLACK_BOT_USER_ID` disables discovery.
- **LangSmith.** Credentials are read by their standard names (`LANGSMITH_API_KEY`, `LANGSMITH_ENDPOINT`), so a deployment uses its own workspace and local `langgraph dev` traces with the same variable. The tenant id is read from tracing-project metadata (or the first project in the workspace); the web host for trace links is derived from the endpoint; feedback is recorded in that one workspace.
- **Dashboard.** `DASHBOARD_BASE_URL` is always allowed for credentialed CORS; `DASHBOARD_ALLOWED_ORIGINS` only adds further origins.

**Deprecated, still honored as fallbacks with a startup warning:** `GITHUB_APP_ID`, `DEFAULT_REPO_OWNER`/`DEFAULT_REPO_NAME`, `SLACK_REPO_OWNER`/`SLACK_REPO_NAME`, and the second-workspace aliases `LANGSMITH_API_KEY_PROD`, `LANGSMITH_ENDPOINT_PROD`, `LANGSMITH_URL_PROD`, `LANGSMITH_TENANT_ID_PROD`, `LANGSMITH_TRACING_PROJECT_ID_PROD`, `LANGGRAPH_URL_PROD`. The `LANGCHAIN_*` SDK aliases (`LANGCHAIN_API_KEY`, `LANGCHAIN_TRACING_V2`, `LANGCHAIN_PROJECT`) are no longer read and only warn.

**Also in this PR**

- The team setting `default_repo` is the canonical default repository; `repo:name` shorthand resolves against its owner. The hardcoded `langchain-ai` owner default is removed.
- Startup logs which surfaces are enabled, which variable a surface is still missing, explicit overrides, and any deprecated variable found. Names only, never values.
- `make setup` (`scripts/setup_env.py`) writes `.env` interactively without echoing secrets, reads the GitHub App `.pem`, generates `DASHBOARD_JWT_SECRET` and `TOKEN_ENCRYPTION_KEY` as independent random keys, generates the webhook secret when left blank (and tells you how to copy it from `.env` rather than printing it), and writes the file with mode 600. Existing values are preserved unless `--force`.
- `docs/INSTALLATION.md` is rewritten around the seven-variable happy path; the dashboard (which also provides per-user GitHub identity), Linear, Slack account linking, custom snapshots, other sandbox providers, and production deployment are collapsed optional sections, followed by an overrides/deprecations table.

**Behavior changes to note**

- Deployments that set both `LANGSMITH_API_KEY` and `LANGSMITH_API_KEY_PROD` to different workspaces now use `LANGSMITH_API_KEY`; feedback is no longer written to both.
- Deployments that relied on the implicit `langchain-ai` owner for `repo:name` shorthand now need a team default repository (or `DEFAULT_REPO_OWNER`).
- CORS middleware is installed whenever `DASHBOARD_BASE_URL` is set, restricted to that origin.

Not included: persisting the env-seeded `default_repo` into the Store so the env vars can be deleted (needs the Store reachable during lifespan), and removing the deprecated reads after the transition period.

## Test Plan

- [x] New unit tests: installation discovery (env override, repo/org context, single, multiple, zero, throttling, pagination), issuer selection, Slack identity discovery and throttling, tenant and host derivation, LangSmith credential precedence (standard name over `_PROD`, single feedback workspace), CORS derivation, default-owner precedence, startup report (surfaces, aliases, deprecations), installer (secret generation, dotenv merge, private-key round trip, file mode, secret never printed).
- [x] Full pytest suite: 2118 passed. The 8 remaining failures reproduce identically on `main` (Slack webhook test ordering) and are unrelated.
- [x] `make lint`, `ruff format --check`, `ty check` (two pre-existing warnings), and `pnpm --filter open-swe-dashboard run typecheck` clean.
- [x] Verified locally against real credentials, values never displayed: GitHub App installation discovery (one installation, matching the explicit override) and token mint; Slack `auth.test` identity; LangSmith tenant/project resolution; a backend booted with only the seven minimum variables; signed Slack/GitHub webhooks accepted and ignored by design, bad signatures rejected; CORS/CSRF behavior; a sandbox booted from the LangSmith default snapshot (gh, git, Python, Node, uv, Docker present) and deleted; the interactive installer via `expect`.
- [ ] Preview deployment: labeled `preview`; currently blocked by a `docs/INSTALLATION.md` conflict with #2377 in the preview tree.


langchain-infra:
Split into a stack so the pieces can be reviewed and previewed independently: #2463 (unified config + standard LangSmith env) → #2464 (runtime discovery) → #2465 (team default repo, startup report, make setup) → #2466 (docs). Same end state as this branch, with every env read going through the new `agent/config.py` registry.

PR #2443, opened 2026-09-03
fix: verify Slack signatures from raw request bytes

### LINKED ISSUE #2310: Slack signature is computed over a lossily re-decoded body instead of raw bytes
**Severity:** Low
**Type:** Correctness (signature verification)
**File:** `agent/utils/slack.py:180-185`

## Description

```python
base_string = f"v0:{timestamp}:{body.decode('utf-8', errors='replace')}"
expected = (
    "v0="
    + hmac.new(secret.encode("utf-8"), base_string.encode("utf-8"), hashlib.sha256).hexdigest()
)
return hmac.compare_digest(expected, signature)
```

Slack computes its signature over the raw request body. This implementation decodes those bytes
with `errors="replace"` and re-encodes them. For any byte sequence that is not valid UTF-8, the
round trip is lossy — each invalid byte becomes U+FFFD (`EF BF BD`) — so the HMAC is taken over
different bytes than Slack signed.

The sibling verifiers in this repo both hash the raw body directly and do not have this problem:

```python
# agent/utils/github_comments.py:103
expected = "sha256=" + hmac.new(secret.encode(), body, hashlib.sha256).hexdigest()

# agent/webhooks/common.py:1033
expected = hmac.new(secret.encode("utf-8"), body, hashlib.sha256).hexdigest()
```

## Impact

Low in practice. Slack sends UTF-8 (`application/json` or `application/x-www-form-urlencoded`), so
the decode is lossless and the computed digest matches. The failure mode is a **valid** request
being rejected with a 401 if a non-UTF-8 byte ever reaches the endpoint — a fail-closed direction,
not an authentication bypass. It is a robustness and consistency defect rather than a
vulnerability.

Worth fixing because the correct form is shorter than the current one and matches the two
verifiers beside it, and because a lossy transform inside a signature check is the kind of detail
that is hard to reason about later.

## Suggested fix

Build the base string as bytes and skip the round trip:

```python
base_string = b"v0:" + timestamp.encode("utf-8") + b":" + body
expected = "v0=" + hmac.new(secret.encode("utf-8"), base_string, hashlib.sha256).hexdigest()
```

---

abhinavkr26104:
Fixes #2310.

## Summary
- Build Slack's signed payload from the original request bytes.
- Add a regression test with an invalid UTF-8 byte, signed as Slack signs it.

## Testing
- 
uff format agent/slack/client.py tests/slack/test_slack_context.py`n- 
uff check agent/slack/client.py tests/slack/test_slack_context.py`n- Full pytest is blocked locally: the Windows host lacks MSVC link.exe, required to build the transitive obstore dependency.

### DIFF
--- a/agent/slack/client.py
+++ b/agent/slack/client.py
@@ -176,10 +176,10 @@ def verify_slack_signature(
     if abs(int(time.time()) - request_timestamp) > max_age_seconds:
         return False
 
-    base_string = f"v0:{timestamp}:{body.decode('utf-8', errors='replace')}"
+    base_string = b"v0:" + timestamp.encode("utf-8") + b":" + body
     expected = (
         "v0="
-        + hmac.new(secret.encode("utf-8"), base_string.encode("utf-8"), hashlib.sha256).hexdigest()
+        + hmac.new(secret.encode("utf-8"), base_string, hashlib.sha256).hexdigest()
     )
     return hmac.compare_digest(expected, signature)
 
--- a/agent/slack/client.py
+++ b/agent/slack/client.py
@@ -177,10 +177,7 @@ def verify_slack_signature(
         return False
 
     base_string = b"v0:" + timestamp.encode("utf-8") + b":" + body
-    expected = (
-        "v0="
-        + hmac.new(secret.encode("utf-8"), base_string, hashlib.sha256).hexdigest()
-    )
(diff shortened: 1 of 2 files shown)


johannes117:
Closing as a duplicate of #2319, which already fixes Slack signature verification over the raw request body.

PR #2444, opened 2026-09-03
fix: retain hunks for left-side review findings

### LINKED ISSUE #2305: Reviewer: `extract_diff_hunk` ignores `side`, so LEFT-side findings lose their diff hunk
**Severity:** Medium
**Type:** Correctness (reviewer output quality)
**Files:** `agent/review/diff.py:157-179`, `agent/tools/add_finding.py:136-138`

## Description

The reviewer supports findings anchored to either side of a diff. `compute_diff_line_set` builds
`{file: {"RIGHT": {new_lines}, "LEFT": {old_lines}}}`, and `is_range_in_diff` takes an explicit
`side` argument to pick the right set:

```python
def is_range_in_diff(line_set, file, start_line, end_line, side: DiffSide = "RIGHT") -> bool:
    """... ``side`` selects new-side lines (``RIGHT``) or old-side lines (``LEFT`` — deletions)."""
```

`extract_diff_hunk` has no `side` parameter and unconditionally compares the caller's line
numbers against the **new-side** hunk range:

```python
for hunk in hunks:
    if hunk.new_start <= end_line and start_line <= hunk.new_end:
        return hunk.body
return None
```

`add_finding` validates `side` (`{"LEFT", "RIGHT"}`), correctly forwards it to `is_range_in_diff`,
then two statements later calls `extract_diff_hunk` without it:

```python
in_diff = not isinstance(diff_line_set, dict) or is_range_in_diff(
    diff_line_set, file, start_line, end_line, side=_cast_side(side)     # side honoured
)
...
diff_hunk = extract_diff_hunk(diff_text, file, start_line, end_line)     # side dropped
```

The same pair of integers is interpreted as old-side by the first call and new-side by the second.

## Reproduction

```python
from agent.review.diff import compute_diff_line_set, extract_diff_hunk, is_range_in_diff

DIFF = """diff --git a/x.py b/x.py
--- a/x.py
+++ b/x.py
@@ -100,3 +100,1 @@ def f():
-    old_a()
-    old_b()
-    old_c()
+    new()
"""

ls = compute_diff_line_set(DIFF)
print(sorted(ls["x.py"]["LEFT"]))                                    # [100, 101, 102]
print(sorted(ls["x.py"]["RIGHT"]))                                   # [100]
print(is_range_in_diff(ls, "x.py", 102, 102, side="LEFT"))           # True
print(extract_diff_hunk(DIFF, "x.py", 102, 102))                     # None   <-- bug
```

Actual output:

```
LEFT lines : [100, 101, 102]
RIGHT lines: [100]
is_range_in_diff(LEFT, 102) -> True
extract_diff_hunk(102)      -> None
```

## Impact

A LEFT-side finding is accepted as in-diff, then stored with `diff_hunk = None`. Whenever the
deleted region is larger than the added region (the common shape for a deletion), the old-side
line number falls outside the new-side range and the lookup misses. Findings about deleted code
are published without the surrounding hunk context that the review UI and downstream prompts rely
on. When a file has multiple hunks the failure is worse than `None`: the old-side number can
overlap a *different* hunk's new-side range, attaching the wrong code to the finding.

The existing tests (`tests/reviewer/test_reviewer_diff.py:81-97`) only cover RIGHT-side lookups,
which is why this is not caught.

## Suggested fix

Give `extract_diff_hunk` a `side: DiffSide = "RIGHT"` parameter and select the hunk range to
compare against accordingly (`hunk.old_start`/`hunk.old_end` for `LEFT`), then pass the finding's
`side` from `add_finding`. Add a LEFT-side case to the test file.

## Secondary note

`extract_diff_hunk` also collects every file diff matching the path but only inspects the first:

```python
file_diffs = [fd for fd in parse_unified_diff(diff_text) if fd.file == file]
hunks = file_diffs[0].hunks
```

`compute_diff_line_set` handles the duplicate-path case correctly (it merges via `setdefault`).
The two functions disagree; worth aligning while fixing the above.

---

abhinavkr26104:
Fixes #2305.

## Summary
- Match hunk ranges against the finding side.
- Preserve hunks when the same file appears in multiple diff sections.
- Add a deleted-line regression test.

## Testing
- 
uff format agent/review/diff.py agent/tools/add_finding.py tests/reviewer/test_reviewer_diff.py`n- 
uff check agent/review/diff.py agent/tools/add_finding.py tests/reviewer/test_reviewer_diff.py`n- Full pytest is blocked locally: the Windows host lacks MSVC link.exe, required to build the transitive obstore dependency.

### DIFF
--- a/agent/review/diff.py
+++ b/agent/review/diff.py
@@ -159,6 +159,7 @@ def extract_diff_hunk(
     file: str,
     start_line: int | None,
     end_line: int | None,
+    side: DiffSide = "RIGHT",
 ) -> str | None:
     """Extract the hunk body covering ``file:start_line..end_line``.
 
@@ -168,13 +169,16 @@ def extract_diff_hunk(
     file_diffs = [fd for fd in parse_unified_diff(diff_text) if fd.file == file]
     if not file_diffs:
         return None
-    hunks = file_diffs[0].hunks
+    hunks = tuple(hunk for file_diff in file_diffs for hunk in file_diff.hunks)
     if not hunks:
         return None
     if start_line is None or end_line is None:
         return hunks[0].body
     for hunk in hunks:
-        if hunk.new_start <= end_line and start_line <= hunk.new_end:
+        hunk_start, hunk_end = (
+            (hunk.old_start, hunk.old_end) if side == "LEFT" else (hunk.new_start, hunk.new_end)
+        )
+        if hunk_start <= end_line and start_line <= hunk_end:
             return hunk.body
     return None
 
--- a/agent/review/diff.py
+++ b/agent/review/diff.py
@@ -169,7 +169,7 @@ def extract_diff_hunk(
     file_diffs = [fd for fd in parse_unified_diff(diff_text) if fd.file == file]
     if not file_diffs:
         return None
-    hunks = tuple(hunk for file_diff in file_diffs for hunk in file_diff.hunks)
+    hunks = file_diffs[0].hunks
     if not hunks:
         return None
     if start_line is None or end_line is None:
--- a/agent/tools/add_finding.py
+++ b/agent/tools/add_finding.py
@@ -134,7 +134,7 @@ async def add_finding(
     if isinstance(diff_text, str) and diff_text:
         from agent.review.diff import extract_diff_hunk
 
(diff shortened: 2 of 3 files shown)


johannes117:
Closing as a duplicate of #2314, which already handles LEFT-side diff hunk lookup and repeated file entries.

PR #2446, opened 2026-09-03
fix: support desktop artifact paths on Windows

### LINKED ISSUE #2302: `os.getuid()` in agent/desktop.py breaks the desktop backend on Windows
**Severity:** Medium
**Type:** Portability / Crash
**File:** `agent/desktop.py:51`

## Description

`_artifacts_root()` namespaces the artifacts temp directory by POSIX uid:

```python
def _artifacts_root() -> Path:
    configured = os.environ.get("OPEN_SWE_LOCAL_ARTIFACTS_DIR")
    if configured:
        return Path(configured)
    return Path(tempfile.gettempdir()) / f"open-swe-artifacts-{os.getuid()}"
```

`os.getuid` does not exist on Windows. There is no `sys.platform` guard anywhere in
`agent/desktop.py`.

## Impact

On Windows the desktop/local backend raises immediately:

```
AttributeError: module 'os' has no attribute 'getuid'. Did you mean: 'getpid'?
agent\desktop.py:51: AttributeError
```

This is reachable through `desktop_artifact_routes`, which sets up the routes for offloaded tool
results and evicted conversation history — so the failure is on the agent-assembly path, not an
optional feature.

Two tests fail on Windows for this reason:

- `tests/agent/test_agent_assembly_context.py::test_desktop_agent_loads_snapshotted_and_bundled_skills`
- `tests/agent/test_agent_assembly_context.py::test_desktop_agent_honors_gateway_environment`

Open SWE Desktop is documented as macOS-only (`Makefile: install-desktop`, `desktop/README.md`),
so this may be intentional today. It is still worth fixing: the value is only used to keep two
users' temp directories apart, the fix is one line, and the current code turns "Windows is not a
release target" into "the module cannot be imported and exercised on Windows at all", which is
what makes the test suite unrunnable there.

## Suggested fix

```python
_uid = getattr(os, "getuid", None)
suffix = _uid() if _uid else getpass.getuser()
return Path(tempfile.gettempdir()) / f"open-swe-artifacts-{suffix}"
```

Or gate the whole desktop path behind an explicit platform check and skip the tests accordingly.

---

abhinavkr26104:
Fixes #2302.

## Summary
- Use the POSIX UID when available and the current username otherwise.
- Add a no-os.getuid regression test.

## Testing
- 
uff format agent/desktop.py tests/agent/test_desktop.py`n- 
uff check agent/desktop.py tests/agent/test_desktop.py`n- Full pytest is blocked locally: the Windows host lacks MSVC link.exe, required to build the transitive obstore dependency.

### DIFF
--- a/agent/desktop.py
+++ b/agent/desktop.py
@@ -1,4 +1,5 @@
 import asyncio
+import getpass
 import json
 import os
 import re
@@ -71,7 +72,9 @@ def _artifacts_root() -> Path:
     configured = os.environ.get("OPEN_SWE_LOCAL_ARTIFACTS_DIR")
     if configured:
         return Path(configured)
-    return Path(tempfile.gettempdir()) / f"open-swe-artifacts-{os.getuid()}"
+    uid = getattr(os, "getuid", None)
+    identity = str(uid()) if uid else getpass.getuser()
+    return Path(tempfile.gettempdir()) / f"open-swe-artifacts-{identity}"
 
 
 async def desktop_artifact_routes(thread_id: str) -> dict[str, FilesystemBackend]:
--- a/tests/agent/test_desktop.py
+++ b/tests/agent/test_desktop.py
@@ -7,6 +7,7 @@ import pytest
 from blockbuster import BlockBuster
 
 from agent.desktop import (
+    _artifacts_root,
     create_desktop_backend,
     desktop_artifact_routes,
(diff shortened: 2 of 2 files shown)


johannes117:
Closing as a duplicate of #2311, which already makes the desktop artifacts root portable when os.getuid is unavailable.

PR #2448, opened 2026-09-03
fix: preserve concurrent queued follow-ups

### LINKED ISSUE #2370: Queued follow-ups can be lost or replayed by concurrent store writes
## Summary

`queue_message_for_thread` implements append as a non-atomic read-modify-write of one `pending_messages` store record. Concurrent webhook or dashboard follow-ups can overwrite one another. It also races the consumer, which reads and deletes the same record before processing it.

## Code references

- `agent/utils/thread_ops.py:43-75` reads the full queue with `get_item`, appends, then writes it with `put_item`.
- `agent/middleware/check_message_queue.py:220-234` reads then deletes that same key before consuming its messages.

## Reproduction

1. Start two `queue_message_for_thread(thread_id, ...)` calls concurrently.
2. Have both `get_item` calls observe the same current `messages` list.
3. Let each append a different message and complete both `put_item` calls.

The later write replaces the earlier one, losing a follow-up. If a consumer reads and deletes between a writer read and write, the stale writer can also restore an already-consumed message, which is then replayed at the next model call.

## Expected behavior

Every accepted follow-up is delivered exactly once and FIFO ordering is preserved, including across concurrent workers.

## Actual behavior

Messages can be silently lost or duplicated under normal concurrent Slack edits, dashboard follow-ups, or automation deliveries.

## Why it matters

Thread follow-ups are user instructions. Dropping or replaying one can cause the agent to perform obsolete work, miss a correction, or repeat an action.

## Suggested direction

Use an append-only per-message representation, or a durable atomic/CAS queue operation. A process-local lock would not protect multi-worker deployments. Add deterministic concurrent-producer and producer-vs-consumer regression tests.

## Duplicate check

Closed #888 and merged queue-related PRs (#1896, #2055, #2298, #2367) address queue capability and UI/dispatch behavior, but do not make the shared read-modify-write queue atomic.

abhinavkr26104:
Fixes #2370.

## Summary
- Store each queued follow-up in its own append-only record.
- Consume records in creation order while retaining legacy queue compatibility.
- Add concurrent producer and ordered consumer regression coverage.

## Testing
- Ruff format: agent/utils/thread_ops.py, agent/middleware/check_message_queue.py, tests/utils/test_thread_ops.py, tests/middleware/test_check_message_queue.py
- Ruff check: the same files
- Full pytest is blocked locally because the Windows host lacks MSVC link.exe, required to build the transitive obstore dependency.


abhinavkr26104:
@christian-bromann can you review the pr

PR #2458, opened 2026-09-04
fix(integrations): retry transient tool loader failures

### LINKED ISSUE #2453: Transient integration-loader failures disable tools until process restart
## Summary

`DynamicToolMiddleware` permanently caches an empty tool set when an integration loader fails once.

## Code reference

`agent/middleware/dynamic_tools.py:165-179`

`_resolve()` catches every loader exception, stores `tools = []`, and sets `resolved.done = True`. All later calls return the cached empty result without attempting another handshake.

## Reproduction

Use an `IntegrationGroup` whose loader raises on its first call and returns a valid tool on its second call. Call `_build(["working_tool"])` twice. The second call still reports the tool unavailable and the loader invocation count remains one.

## Expected behavior

Transient MCP, credential, or network failures should be retried on a later request, ideally with bounded backoff/TTL to avoid a thundering herd.

## Actual behavior

The first failure is treated as a permanent process-lifetime result. Recovery requires restarting the application.

## Why it matters

A short-lived integration outage can silently disable connected tools for every thread handled by that process.

## Suggested fix

Do not mark failed resolutions as complete, or cache failures with an expiry and retry policy. Add a regression test where the first load fails and the next load succeeds.

jackjin1997:
Fixes #2453

Transient integration loader failures no longer become process-lifetime empty tool caches. A later request retries the loader and can restore the integration without restarting the worker.

---

The failure path now leaves the group unresolved while successful empty results remain cacheable. The regression test exercises a loader that fails once and succeeds on the next request.

Local verification: focused middleware tests (10 passed), Ruff check, Ruff format check, and ty check. The repository-wide ty command completed with two pre-existing warnings outside this change.

## AI Disclosure

This change was implemented and tested with AI assistance.

### DIFF
--- a/agent/middleware/dynamic_tools.py
+++ b/agent/middleware/dynamic_tools.py
@@ -174,7 +174,7 @@ class DynamicToolMiddleware(AgentMiddleware[DynamicToolState]):
                 tools = await self._groups[group].load()
             except Exception:
                 logger.warning("Failed to load %s integration tools", group, exc_info=True)
-                tools = []
+                return resolved.tools
             resolved.tools = {tool.name: tool for tool in tools}
             resolved.done = True
         return resolved.tools
--- a/tests/middleware/test_dynamic_tools.py
+++ b/tests/middleware/test_dynamic_tools.py
@@ -217,6 +217,32 @@ async def test_a_group_that_fails_to_build_is_reported_not_raised() -> None:
     assert "unavailable right now" in message.content
 
 
+async def test_a_group_that_fails_to_build_is_retried() -> None:
+    attempts = 0
+
+    async def load() -> list[BaseTool]:
+        nonlocal attempts
+        attempts += 1
+        if attempts == 1:
+            raise RuntimeError("mcp temporarily unreachable")
+        return [_tool("analyzePlan")]
+
+    middleware = DynamicToolMiddleware(
+        {"Corridor": IntegrationGroup(tool_names=("analyzePlan",), load=load)}
+    )
+    coroutine = cast(Any, cast(StructuredTool, middleware.tools[0]).coroutine)
+
+    first = await coroutine(tool_names=["analyzePlan"], state={}, tool_call_id="load-1")
+    second = await coroutine(tool_names=["analyzePlan"], state={}, tool_call_id="load-2")
+
+    first_message = cast(dict[str, Any], first.update)["messages"][0]
+    assert first_message.status == "error"
+    second_update = cast(dict[str, Any], second.update)
+    assert second_update["loaded_integration_tools"] == ["analyzePlan"]
+    assert second_update["messages"][0].status == "success"
+    assert attempts == 2
+
+
 async def test_a_group_whose_catalog_is_empty_is_not_offered() -> None:
     async def load() -> list[BaseTool]:
         return []


ArshVermaGit (commented):
this cleanly preserves cached tools on transient loader failures, with a focused regression test covering the retry behavior. Approvable

jackjin1997:
@ramon-langchain, this fix is still active: #2459 was closed as a duplicate in favor of this PR, and the current #2619 refactor moves the loader but does not replace the retry-on-transient-failure behavior covered here. Could you add `!keep-open` or route this to the current middleware reviewer before the automatic cleanup closes it? The focused regression remains ready to review.

jackjin1997:
@ramon-langchain since you wrote most of the integration loader path, could you take a look before the 14-day auto-close lands next week? The change is small: a loader exception no longer marks the group resolved with an empty cache, so the next request retries instead of the worker staying tool-less for its lifetime. The duplicate #2459 was already closed in favor of this one, checks are green, and the fail-once/succeed-next regression is included. Happy to adjust if you would rather handle the retry differently.

agent/middleware/dynamic_tools.py
```diff
@@ -174,7 +174,7 @@ async def _resolve(self, group: str) -> dict[str, BaseTool]:
                 tools = await self._groups[group].load()
             except Exception:
                 logger.warning("Failed to load %s integration tools", group, exc_info=True)
-                tools = []
+                return resolved.tools
```
ramon-langchain: should this be just `return {}` since we know that `resolved.done` is false?

agent/middleware/dynamic_tools.py
```diff
@@ -174,7 +174,7 @@ async def _resolve(self, group: str) -> dict[str, BaseTool]:
                 tools = await self._groups[group].load()
             except Exception:
                 logger.warning("Failed to load %s integration tools", group, exc_info=True)
```
ramon-langchain: I know this is existing code but can we switch this to structured logging ?

ramon-langchain (approved):
nits

agent/middleware/dynamic_tools.py
jackjin1997: Agreed. Since `resolved.done` is false on this path, there cannot be a completed cached result to preserve. I changed this to `return {}`.

agent/middleware/dynamic_tools.py
jackjin1997: Updated. The warning now uses a static message with `integration_group` in `extra`, while retaining `exc_info=True`.

jackjin1997:
@ramon-langchain thanks for the review. Both nits are addressed in c0b67d7: the unresolved path now returns `{}` directly, and the warning uses structured logging with `integration_group` in `extra`. Checks are green and the PR is mergeable. One thing to flag: the `pending-deletion` label from the auto-close bot is still on this PR, so it will be closed on the 18th unless it is merged or a maintainer applies `do-not-close` first.

jackjin1997:
Timing note on the auto-close, since this one is already approved. `.github/scripts/labeling/close-old-prs.js` closes a PR once its age is at least 14 days **and** the warning comment is at least 7 days old (`closeDays - warningDays`), with no exemption for review state. The warning here was posted 2026-09-11 14:28 UTC, so today's 14:17 UTC run still reads the warning as 6 days old and skips; tomorrow's run on 19 Sep is the first where both conditions hold, and this PR gets closed then.

A `do-not-close` label, or a `!keep-open` comment from anyone with MEMBER/OWNER/COLLABORATOR association, clears it via `keep_open_on_comment.yml`.

@ramon-langchain you approved this on 12 Sep and both nits are addressed in c0b67d7 (unresolved path returns `{}` directly, warning moved to structured logging). Merging ahead of tomorrow's run would be the cleanest outcome. If it does get closed I will reopen it, but that loses the approval state.

jackjin1997:
The auto-close landed on 19 Sep at 14:24 UTC as expected, so this one went out at `c0b67d7` with ramon's approval on it and both nits addressed. Reopening from my side is refused by GitHub even though the head branch is intact and there is no competing PR, so it needs someone with write access.

Leaving it here rather than opening a replacement, since a fresh PR would only reset the same 14-day clock and would drop the review history. If the fix is still wanted, a reopen plus `!keep-open` (or the `do-not-close` label) is all it takes, and I will rebase on request. If the team would rather carry it internally, the change is two small pieces: treat an unresolved integration as `{}` instead of caching the empty tool set, and log the failure with `integration_group` in `extra`.

PR #2465, opened 2026-09-04
feat: team default repository, startup configuration report, and guided make setup

langchain-infra:
## Summary

Third PR of the install-simplification stack.

- **Team default repository is canonical.** `repo:name` shorthand resolves against the team default's owner; the hardcoded `langchain-ai` owner default is gone. `DEFAULT_REPO_*` / `SLACK_REPO_*` keep working as deprecated fallbacks.
- **Startup configuration report.** At startup the API logs which surfaces are enabled (LangSmith, GitHub, Slack, Linear, Dashboard) or which variable each is missing, which discovery-backed values are explicitly overridden, and every deprecated variable present. Names only; values never reach the log. Deprecation hints come from the registry in `agent/config.py`, so a variable is described in exactly one place.
- **`make setup`.** Interactive `.env` writer for the GitHub + Slack minimum. Secrets are read without echo, `DASHBOARD_JWT_SECRET` and `TOKEN_ENCRYPTION_KEY` are generated independently when the dashboard is enabled, and a generated webhook secret is pointed to in `.env` rather than printed.

## Testing

- New: `tests/utils/test_startup_config.py`, `tests/scripts/test_setup_env.py`, team-default cases in `tests/slack/test_slack_context.py`.
- `uv run pytest tests/`: all green except the 8 Slack webhook cases that already fail on a clean `main`.
- `make setup` driven end to end with `expect` and fake values; a minimum-only `.env` boots with the expected report.

## Stack

1. #2463 unified config + standard LangSmith env
2. #2464 runtime discovery
3. **this PR** config cleanup + guided `make setup`
4. #2466 installation docs rewrite

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all format format-check lint typecheck test tests integration_tests help run dev web desktop install-desktop install-checkout
+.PHONY: all format format-check lint typecheck test tests integration_tests help run dev web desktop install-desktop install-checkout setup
 
 # Default target executed when no arguments are given to make.
 all: help
@@ -30,6 +30,9 @@ install-checkout:
 
 install:
 	uv sync --extra dev
+
+setup:
+	uv run python scripts/setup_env.py
 
 ######################
 # TESTING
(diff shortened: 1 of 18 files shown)


langchain-infra:
Folded into #2464 to shorten the stack; the same five commits now sit on `mukil/runtime-discovery`.

PR #2484, opened 2026-09-05
feat(tracing): trace every run into LANGSMITH_PROJECT

langchain-infra:
## Summary

Sixth PR of the install-simplification stack, on top of #2482. Every run now traces into the project named by `LANGSMITH_PROJECT`, the LangSmith SDK's own variable, instead of the hardcoded `open-swe-agent` / `open-swe-review` projects.

- The graph entrypoints no longer wrap graphs in `tracing_context(project_name=…)`. The SDK reads `LANGSMITH_PROJECT`, and LangGraph Platform sets it to the deployment name, so each deployment's traces stay in its own project. Until now prod, preview, and staging agent runs all landed in one shared `open-swe-agent` project while the FastAPI side (thread titles, feedback) traced into the deployment's project, which is what made staging's traces look mixed up.
- Trace links, thread cost lookups, and thread feedback resolve that same project by name. `LANGSMITH_TRACING_PROJECT_ID` and its `_PROD` alias are removed; `LANGSMITH_PROJECT` is declared in the registry with the SDK's `default`.
- The review trace context keeps its team-configured `review_tracing_project`: that is the project it searches for the reviewed change's own coding-agent traces, not where the reviewer traces itself. `get_langsmith_trace_url` therefore still accepts an explicit project for that caller.
- The `traced_*` entrypoint names stay because `langgraph.json` and the e2e configs import them; they are now plain aliases of the factories.

## Test Plan

- [x] `tests/agent/test_langsmith_trace_url.py`: links follow `LANGSMITH_PROJECT`, fall back to the SDK's `default` project when unset, and feedback resolves the same project.
- [x] Full suite green apart from the 8 Slack cases that fail on a clean `main`.
- [ ] Staging: run from the dashboard and confirm the agent trace and the thread-title trace both appear in the `open-swe-staging` project, and that the "View trace" link opens it.


langchain-infra:
Folded into #2464 to shorten the stack (the code change) and #2466 (the installation-guide text). The tree at the top of the stack is unchanged.

PR #2488, opened 2026-09-05
feat: startup configuration report and guided make setup

langchain-infra:
## Summary

Fourth PR of the install-simplification stack, on the Slack discovery PR. Operator-facing pieces that belong to no single integration.

- **Startup configuration report.** At startup the API logs which surfaces are enabled (LangSmith, GitHub, Slack, Slack sign-in, Linear, Dashboard) or which variable each is missing, which discovery-backed values are explicitly overridden, and every deprecated variable present. Names only; values never reach the log. Deprecation hints come from the registry in `agent/config.py`.
- **`make setup`.** Interactive `.env` writer for the GitHub + Slack minimum. Secrets are read without echo, `DASHBOARD_JWT_SECRET` and `TOKEN_ENCRYPTION_KEY` are generated, and a generated webhook secret is pointed to in `.env` rather than printed. `--dashboard` also collects the dashboard settings and offers the Sign in with Slack client id and secret.
- **Dashboard CORS.** `DASHBOARD_BASE_URL` is always an allowed origin; `DASHBOARD_ALLOWED_ORIGINS` only adds extra ones.

## Testing

- New: `tests/utils/test_startup_config.py`, `tests/scripts/test_setup_env.py`, `tests/dashboard/test_app_cors.py`.
- `make setup` driven end to end with `expect` and fake values; a minimum-only `.env` boots with the expected report.

## Stack

1. #2463 unified config + standard LangSmith env (merged)
2. #2464 GitHub App discovery
3. #2487 Slack bot discovery and repository defaults
4. **this PR** startup report and `make setup`
5. #2489 LangSmith tenant discovery and `LANGSMITH_PROJECT` tracing
6. #2486 single-origin dashboard
7. #2466 installation docs rewrite

PR #2487, opened 2026-09-05
feat(slack): discover the bot identity and make the team default repository canonical

langchain-infra:
## Summary

Third PR of the install-simplification stack, on the GitHub discovery PR. The Slack side needs only `SLACK_BOT_TOKEN` and `SLACK_SIGNING_SECRET`.

- **Bot identity.** The bot user id and handle come from the token via `auth.test` and `users.info` at startup, retried on the first webhook (60s throttle) if Slack was unreachable. Setting `SLACK_BOT_USER_ID` disables discovery entirely, so scripted-fake tests are untouched.
- **Team default repository is canonical.** `repo:name` shorthand resolves against the team default's owner; the hardcoded `langchain-ai` owner default is gone. `DEFAULT_REPO_*` / `SLACK_REPO_*` keep working as deprecated fallbacks, and only fill in while no team default exists.
- **Repository resolution replies.** Thread → channel topic `repo:` → the mentioning user's profile default (for users linked through Sign in with Slack or an admin mapping) → team default. When nothing resolves, the bot replies in the thread instead of failing the webhook silently, and tells an unlinked person how to link their account when Sign in with Slack is configured.

## Testing

- New: `tests/slack/test_slack_bot_identity.py`, `tests/slack/test_slack_no_repository.py`; team-default and precedence cases in `tests/slack/test_slack_context.py` and `tests/github/test_repo_extraction.py`.
- `uv run pytest tests/`: green apart from the 8 Slack webhook cases that already fail on a clean `main`.
- Verified against the real API from a local checkout: Slack identity via `auth.test`/`users.info`, signed webhooks accepted, bad signatures 401. Running on the `open-swe-staging` deployment: mentions resolve the repository through the profile default and the reply appears when nothing resolves.

## Stack

1. #2463 unified config + standard LangSmith env (merged)
2. #2464 GitHub App discovery
3. **this PR** Slack bot discovery and repository defaults
4. #2488 startup report and `make setup`
5. #2489 LangSmith tenant discovery and `LANGSMITH_PROJECT` tracing
6. #2486 single-origin dashboard
7. #2466 installation docs rewrite

PR #2490, opened 2026-09-06
docs(oep): OEP-0001 first-boot setup flow instead of pre-boot configuration

langchain-infra:
Draft OEP recording the direction discussed for the install simplification: a token-protected first-boot setup page that creates the GitHub App through GitHub's App Manifest flow, creates and installs the Slack app from a manifest with an OAuth install, and bootstraps the first admin from the first GitHub login. Pre-boot configuration shrinks to the LangSmith key and two generated secrets.

Depends on the bundled dashboard (#2486). The GitHub/Slack discovery and `make setup` PRs (#2464, #2487, #2488) and the docs rewrite (#2466) are parked as drafts pending this decision.

Per OEP-0000, merging this draft publishes it for discussion and does not accept it.

### DIFF
--- a/oeps/0001-first-boot-setup.md
+++ b/oeps/0001-first-boot-setup.md
@@ -1,0 +1,110 @@
+# OEP-0001: First-boot setup flow instead of pre-boot configuration
+
+- **Authors:** Mukil Loganathan (`@mukil`)
+- **Status:** Draft
+- **Created:** 2026-09-05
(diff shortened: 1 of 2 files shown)


langchain-infra:
Closing: a full OEP is more than this needs. The direction continues as a local prototype of the setup flow; the text here stays as the reference for it.

PR #2498, opened 2026-09-07
docs: installation guide for deploying a single-origin Open SWE

langchain-infra:
## Summary

Rewrites `docs/INSTALLATION.md` around the single-origin deployment that landed in #2486: one deployment serves the API, webhooks and dashboard from one URL, so the separate frontend deploy, CORS and `DASHBOARD_*` URL steps leave the main path.

Order of the guide: create the deployment first (LangGraph Platform or standalone Docker) so you have its URL, get a LangSmith key, create the GitHub App by hand with the full permission and event list, choose model providers and API keys (its own step: provider keys, the LLM Gateway, how the running model is chosen, other optional keys), set the environment variables, verify. Slack, Linear, a separately deployed dashboard, admin API credentials, allowlists and key rotation are collapsed add-on sections, followed by deployment troubleshooting. Sandbox snapshot configuration is gone from the guide: deployments boot from LangSmith's root snapshot, and admins capture Environments in the dashboard when repositories need more. The `#3-create-a-github-app` anchor that `agent/prompt.py` links to is preserved.

Local development is a short section at the end for now; the full local guide with the tunnel setup follows in a stacked PR.

Also: the README's Installation Guide bullet describes the new scope, and a comment in `langgraph.json` drops a stale section reference.

## Test plan

- [x] Internal anchors resolve; `docs/INSTALLATION.md#3-create-a-github-app` still exists
- [x] `langgraph.json` still valid JSON
- [x] The env-var behaviour described (dashboard URLs defaulting to `LANGGRAPH_URL`, no CORS setup on one origin) is what `agent/config.py` and `agent/utils/dashboard_links.py` implement on main, including #2495

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Sandboxes can have network access and powerful tools. Deployments should use lea
 
 Open SWE includes a LangGraph backend, a web dashboard, and an experimental desktop client.
 
-- **[Installation Guide](docs/INSTALLATION.md)** — Set up local development, the GitHub App, LangSmith, integrations, and production deployment
+- **[Installation Guide](docs/INSTALLATION.md)** — Deploy Open SWE for a team: LangGraph Platform or Docker, the GitHub App, environment variables, and the optional Slack and Linear triggers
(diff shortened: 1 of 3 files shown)


langchain-infra:
Folded into #2499, which now targets main and carries both guides.

PR #2501, opened 2026-09-08
feat(auth): require Open SWE's own credentials on the LangGraph API

langchain-infra:
## Summary

The LangGraph API (`/threads`, `/runs`, `/assistants`, `/store`, …) was only protected where the server happened to authenticate: LangSmith key auth on LangGraph Platform, nothing under `langgraph dev`, and nothing on a standalone image left at its default `LANGGRAPH_AUTH_TYPE=noop`. Open SWE already had a custom auth handler, but only the desktop configuration used it.

`langgraph.json` now names `agent.langgraph_auth:auth` (the desktop handler, moved and extended), so every LangGraph API request must carry one of:

- `X-Api-Key` equal to `LANGSMITH_API_KEY`: what the dashboard's proxy already sends, and what the SDK clients used for background work already read from the environment;
- `Authorization: Bearer OPEN_SWE_LOCAL_AUTH_TOKEN`: the desktop app's embedded backend, as before.

Everything else is rejected with 401, in every mode. Custom routes are not covered by the handler and keep their own checks: `/dashboard/api/*` its session cookie, `/webhooks/*` their signatures. `/ok` and `/info` stay public.

`disable_studio_auth` is set deliberately. Under `langgraph dev`, a custom auth handler otherwise lets any request carrying `x-auth-scheme: langsmith` through as a Studio user with no verification, which would make the handler pointless exactly where it matters most. The cost is that LangSmith Studio cannot open a deployment of this app; the dashboard is the UI. Flip the flag if that trade-off changes.

No caller changed: nothing in the codebase talks to the LangGraph API without one of the two credentials. `agent/local_auth.py` is removed and both desktop configs point at the new module.

Docs follow in #2499, which will be stacked on this.

## Test plan

- [x] `tests/auth/test_langgraph_auth.py`: API key accepted (bytes and str header names), bearer accepted, wrong or missing credentials 401, the Studio header alone 401, nothing configured rejects everything, all three configs point at the handler
- [x] `tests/auth`, `tests/dashboard`, `tests/agent`: 810 passed; ruff and ty clean
- [x] Live against `langgraph dev` from this branch with throwaway values: `POST /threads/search` 401 with no credentials, 200 with the API key, 401 with a wrong key, 200 with the bearer token, 401 with a wrong token, 401 with only `x-auth-scheme: langsmith`; `GET /assistants/agent` 401 unauthenticated; `/ok` and `/info` 200; `/dashboard/api/me` still the app's own 401; `GET /webhooks/slack` still 200. The SDK with `LANGSMITH_API_KEY` in the environment listed assistants; with a wrong `api_key` it raised `AuthenticationError`. Server log confirmed `Using auth of type=custom`.


langchain-infra:
Parked for now per discussion; #2499 no longer depends on it.

PR #2526, opened 2026-09-08
fix(ui): strip request framing headers in the backend proxy

ramon-langchain:
`ui/server/backend-proxy.ts` forwarded the inbound request headers to `fetch` verbatim. When Vercel handed the handler a chunked POST, `transfer-encoding` travelled with it and undici rejected the request outright:

```
[TypeError: fetch failed]
  cause: InvalidArgumentError: invalid transfer-encoding header
    at processHeader (node:internal/deps/undici/undici:3204:15)
    at new Request (node:internal/deps/undici/undici:3018:15)
```

The throw escaped the handler, so the dashboard got a 500 on requests that never reached the backend — nothing in the backend's prod logs, three confirmed hits on `POST /dashboard/api/threads/pull-request-checks` in the Vercel runtime log.

The `HOP_BY_HOP` set already existed but was only applied to response headers. Request headers now go through the same filter, plus `content-length`, since the body is re-framed as a stream.

## Test Plan

- [x] `npx vitest run server/backend-proxy.test.ts` — new test asserts `transfer-encoding`, `content-length` and `connection` are dropped while `content-type` and `cookie` are forwarded
- [x] `npx tsc --noEmit`


langchain-infra (approved):

ramon-langchain:
Acknowledged the Open SWE review (no findings). Now fixing the JS format check and looking at the two E2E shards.

— Claude

PR #2537, opened 2026-09-09
refactor(ui): run agent threads on the SDK's useStream

ramon-langchain:
## Summary

Replaces the hand-rolled LangGraph `StreamController` registry in the dashboard with the SDK's `useStream`, following the `@langchain/react` patterns.

- **`ui/src/features/agents/lib/stream/`** is the new stream layer. `AgentStreamProvider` mounts one `useStream` per retained thread and serves the one the route asks for through `useAgentStream()`. `streamPool` (zustand) owns the only app-level lifecycle logic left: which threads stay mounted (active, running, or left within the last minute, capped at 8 idle), and re-keying a lazily created thread to its server id. Switching back to a recently viewed thread is instant and runs never get orphaned.
- SDK bumps: `@langchain/react` 1.0.33 → 1.0.35, `@langchain/langgraph-sdk` 1.10.0 → 1.10.2.
- `OptimisticThreadHydrationRecovery` and the `thread.messages` snapshot fallback are gone. Both worked around a per-thread provider that 404'd on hydrate; the SDK's optimistic echo and the persistent instance cover that flow now.
- One `promptMessage` builder replaces three copies (home, thread follow-up, local view). The desktop mid-run queue protocol moves out of the 650-line local view into `localMessageQueue.ts`.
- Shared LangGraph client factories in `ui/src/lib/langgraph-client.ts`; the global `overrideFetchImplementation` is removed from both agents and review chat in favor of `onRequest` + `fetch`.
- `routes/agents.tsx` derives the active thread from router matches instead of splitting the pathname against a hard-coded list of reserved segments.
- Each transcript remembers its scroll position across navigation. Leaving while pinned to the tail means following new content on return.

## Test Plan

- [x] `pnpm typecheck`, `oxlint`, `oxfmt --check` pass on the changed files
- [x] New unit tests for the stream pool (reuse, cloud/local isolation, lazy re-key, TTL and cap eviction, retention restart on leave) and the provider (route binding, lazy-thread announce only on first accepted run while active, no announce for local or follow-up runs)


langchain-infra (approved):

ui/src/features/agents/lib/stream/useLocalPromptQueue.ts
hntrl, line 1: we have our own queueing semantics inside of `useStream` -- if that doesn't work for open-swe thats good insight for us

hntrl (commented):
StreamPool is very creative lol. At some point I would like to tie `useStream` direct to the components that consume it instead of using this "react singleton" pattern

ui/src/features/agents/lib/stream/useLocalPromptQueue.ts
```diff
@@ -0,0 +1,65 @@
...
+    handoffRef.current = true
+    // oxlint-disable-next-line react/set-state-in-effect
+    setState({ sessionId, items: [] })
+    void takeLocalPromptQueue(client, sessionId)
+      .then((pending) => pending && submit(pending.text, pending.images))
+      .catch(setError)
+      .finally(() => {
+        handoffRef.current = false
+      })
+  }, [client, isRunning, queued.length, sessionId, submit])
```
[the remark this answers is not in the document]
ramon-langchain: Good catch. `submit` resolves `false` on a credential or stream failure, and the store item was already gone by then.

Fixed: the hook now reads the queue, submits, and only clears the store copy once the run is accepted. A failed handoff leaves the follow-up in the store for the next run (the in-memory card clears and the error surfaces in the view). Covered by a new test, "keeps the follow-up in the store when the run is not accepted".

— Claude

ui/src/features/agents/lib/stream/useLocalPromptQueue.ts
ramon-langchain: Looked at `multitaskStrategy: "enqueue"` / `useSubmissionQueue` before keeping this. It does not fit for three reasons:

1. **Mid-run steering, not a next run.** Open SWE's follow-ups are meant to reach the *current* run: the agent's `check_message_queue_before_model` middleware drains the store queue before its next model call, so a "also update the tests" lands in the running turn. The SDK queue holds the submission until the active run terminates and then starts a new run.
2. **Shared with other surfaces.** The same store item is written by Slack, Linear, GitHub webhooks and the dashboard's cloud path, and read by the agent. The SDK queue is client-local, so a follow-up from another surface (or another tab) would not see it and vice versa.
3. **Survives reload and disconnect.** The store copy outlives the page; the SDK queue is in-memory and is cleared on thread switch.

What would make the SDK queue usable here: a server-side queue the graph can drain mid-run (the SDK docs already mention a roadmap item for server-exposed queue cancel), or a `submit` mode that injects into the active run rather than waiting for it.

— Claude

PR #2606, opened 2026-09-10
refactor: standardize agent execution terminology on invocation

mdrxy:
Open SWE uses “run” for both agent executions and underlying runtime or tracing records. This PR calls one agent execution an **invocation** and retains **thread** for the durable conversation and work context. An initial request and a follow-up belong to one thread and produce two invocations.

Usage reporting adopts invocation terminology for execution counts and average duration. The backend introduces an explicit `invocation_id` and carries it through dispatch, usage recording, completion, and cost refresh. LangSmith cost lookup uses this identity to correlate tracing data with the invocation.

Existing storage keys and runtime identifiers remain unchanged. Legacy identity fields and usage API aliases remain supported for older workers, clients, queued jobs, and historical records. The UI also accepts legacy usage responses during the rollout.


mdrxy:
In a follow-up, I intend to make the usage leaderboard default to showing sessions instead of invocations.

ramon-langchain (approved):

PR #2629, opened 2026-09-10
fix(slack): interrupt for passive follow-ups

ramon-langchain:
Untagged messages accepted from active two-party Slack threads now use the same interrupt strategy as explicit mentions, so the agent consumes them immediately instead of enqueueing them behind active work.

Code-channel chatter and message edits retain their existing non-interrupting behavior. The busy-thread E2E now verifies that an accepted untagged follow-up interrupts the active run.

Made by [Open SWE](https://openswe.vercel.app/agents/7444b0a4-7906-5fd4-ac9e-795162497eaf) · openai:gpt-5.6-sol (high)

### DIFF
--- a/agent/slack/webhook.py
+++ b/agent/slack/webhook.py
@@ -77,20 +77,18 @@ def _interrupts_active_run(
     text: str,
     bot_user_id: str,
     *,
-    treat_all_messages_as_mentions: bool,
     code_channel: bool,
     message_update: bool,
     explicit_request: bool,
-    untagged_reply: bool,
 ) -> bool:
-    return (
-        explicit_request
-        or untagged_reply
+    return not message_update and (
+        not code_channel
+        or explicit_request
         or _is_explicit_slack_request(
             text,
             bot_user_id,
-            treat_all_messages_as_mentions=treat_all_messages_as_mentions and not code_channel,
-            message_update=message_update,
+            treat_all_messages_as_mentions=False,
+            message_update=False,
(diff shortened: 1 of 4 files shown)


agent/slack/webhook.py
```diff
@@ -81,12 +81,17 @@ def _interrupts_active_run(
...
 ) -> bool:
-    return explicit_request or _is_explicit_slack_request(
-        text,
-        bot_user_id,
-        treat_all_messages_as_mentions=treat_all_messages_as_mentions and not code_channel,
-        message_update=message_update,
+    return (
+        explicit_request
+        or untagged_reply
+        or _is_explicit_slack_request(
```
ramon-langchain, line 87: wording here is weird explicit or explicit

tests/e2e/tests/slack_debounce.spec.ts
```diff
@@ -1,16 +1,13 @@
...
-// follow-ups land mid-run.
+// Feature: while Open SWE is busy, an untagged two-party follow-up interrupts
+// the active run just like an explicit mention. Driven through the real webhook +
+// real agent; the LLM is faked and holds a run open so the follow-up lands mid-run.
 
 type SendResult = {
   thread_ts: string;
   thread_id: string;
-  webhook: { status: string };
+  webhook: { status: string; reason?: string };
```
ramon-langchain: do we really need to make this optional?

PR #2667, opened 2026-09-11
feat: replace the routing classifier with agent-driven pre-routed mode

ramon-langchain:
## Summary

Replaces the hidden model-routing classifier with **pre-routed mode**, a mini plan mode the agent exits itself. Design record: [OEP-0003](oeps/0003-pre-routed-mode.md).

- Every routed thread starts pre-routed: the model call runs on the `fast` profile and a new `exit_pre_routed_mode(model_route, title)` tool ends it. While pre-routed, only `execute`, `read_file`, `ls`, `glob`, and the exit tool may be called; other tool calls are rejected with an error result. The prompt tells the agent to size the task from the request and the code, not to edit anything, then exit. The decision is stored in the thread settings and is final for the thread.
- The tool list and system prompt are identical before and after the exit, so a `fast`→`fast` exit keeps the provider prompt cache warm.
- **Sizing is the first turn of the real conversation, not a side call.** `exit_pre_routed_mode` only appends its `ToolMessage`; nothing trims the history and the middleware only swaps the model. Every file read and shell result from the sizing turn is already in the context the chosen model continues from. On `main` the classifier is a separate `ainvoke` on its own prompt that sees the request text truncated to 8,000 chars, reads no code, and has both its input and its output discarded.
- `approve_plan` is renamed to `exit_plan_mode` and takes the same `model_route` and `title`, so a plan-mode thread is routed on approval. Plan mode still forces the `performance` profile.
- `save_plan` accepts an optional `title` so a thread that starts in plan mode is named when the plan is first published.
- With routing on, the agent-chosen title replaces the background first-message titler and the desktop branch-name model call. Both stay as the routing-off path.
- Auto model selection no longer forces routing on. A dashboard thread with no picked model inherits the user and org toggles; picking a model explicitly still turns routing off.

## Eval

`evals/routing/` runs the real agent on real models against 40 tasks, stopping each run at the routing decision and scoring the route, the title, and what the sizing turn cost. Each task carries the **set** of routes a competent agent could defensibly pick, so a task on a tier boundary does not punish either side.

Smoke subset, 6 tasks, about 4 minutes:

```
task                                accepted       actual       calls  tools  ktok   secs
fast-editorconfig                   fast           fast         5      5      94.7   14
none-hello                          none|fast      none         1      0      14.8   32
balanced-github-token-retry         balanced       balanced     7      8      129.7  30
fast-recent-commits                 fast|none      none         5      5      76.4   42
performance-planner-executor-split  performance    performance  24     25     532.3  122
balanced-save-plan-paths            fast|balanced  fast         30     36     786.1  236

route accuracy 6/6
```

```bash
uv run python -m evals.routing.run_eval --smoke --max-concurrency 6
```

## Known gap

The sizing turn is not cheap in wall-clock or in tokens. On open-ended tasks the agent has been observed spending 30 to 56 model calls and up to 1.8M input tokens before routing, against `main`'s single small classifier call. The prompt asks it to keep the turn short, which is not binding.

What bounds the cost is that none of it is thrown away. The exploration is not a preamble to the work — it is the work, carried forward verbatim to whichever model is picked, so the routed model resumes from a context it would otherwise have had to build for itself. The comparison against `main` is therefore not "one classifier call versus thirty", it is "one wasted call plus a cold start versus thirty calls the next model inherits". The residual waste is the difference between what the agent read to *decide* and what it needed to *finish*, which is why the prompt tells it to exit on the higher profile when torn rather than read further.

A hard cap in the middleware, collapsing the tool list to the exit tool after N tool calls, is the next step and is tracked in OEP-0003's unresolved questions rather than in this PR.

## Notes

Dashboard and Slack plan approval still clear `plan_mode` without a route, so the next run enters pre-routed mode with the approved plan as input. The E2E fake model leaves pre-routed mode before playing its script, the same way a real model would.


agent/middleware/model_selection.py
```diff
@@ -59,59 +37,61 @@ class ModelSelectionMiddleware(OpenSWEMiddleware[ModelSelectionState]):
...
     async def awrap_model_call(
         self,
         request: ModelRequest,
         handler: Callable[[ModelRequest], Awaitable[ModelResponse]],
     ) -> ModelResponse:
-        route = (
-            "performance"
-            if request.state.get("plan_mode")
-            else request.state.get("model_route", "balanced")
+        state = request.state
```
ramon-langchain: is it possible to strongly-type state?

agent/middleware/plan_mode.py
```diff
@@ -28,7 +28,7 @@ class PlanModeState(AgentState):
     plan_mode: NotRequired[bool]
 
 
-def _tool_name(tool: BaseTool | dict[str, Any] | Any) -> str | None:
+def tool_name(tool: BaseTool | dict[str, Any] | Any) -> str | None:
```
ramon-langchain: why do we need to accept weak types here?

agent/resources/prompts/system/pre-routed-mode.md
```diff
@@ -0,0 +1,9 @@
+---
+
+### Pre-routed Mode (ACTIVE)
```
ramon-langchain: replace with OEP

agent/middleware/model_selection.py
```diff
@@ -1,56 +1,49 @@
...
-                return authored
-            continue
-        text = message.text
-        if plain or not isinstance(text, str) or "<dynamic-context" in text:
-            continue
-        if "<input-message" not in text:
-            plain = text
-    return plain
 
+class ModelSelectionState(AgentState):
```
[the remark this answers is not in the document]
ramon-langchain: Confirmed, and this is the bug that decides the open design question on this PR.

The fix is the second of your two options: require an exit before any user-facing answer. `none` (answering while still pre-routed) stops being a valid outcome, and `exit_pre_routed_mode` gains a `response` argument carrying the acknowledgement, so routing and the first reply happen in one call. A greeting becomes: exit on `fast`, title the thread, send the reply, and the chosen model does nothing further.

Allowing the source reply tool during sizing was the alternative. It is worse here: it keeps a state where the thread can answer a user without ever recording a route, and it adds a per-source branch to the pre-routed allowlist.

Tracked in OEP-0003. Not yet implemented in this PR.

— Claude

agent/middleware/plan_mode.py
ramon-langchain: We do not — `| Any` is gone. `BaseTool | dict[str, Any]` is what `request.tools` actually holds, and `X | Any` collapses to `Any` anyway, so the old annotation was checking nothing.

I also reverted the `_tool_name` → `tool_name` rename I had made here. Nothing outside this module imports it, so the rename was unmotivated diff noise; this file's diff against main is now just the type tightening.

— Claude

agent/middleware/model_selection.py
ramon-langchain: Yes, and it is now. Two halves:

**Writes.** `abefore_agent` is typed by the base class as returning `dict[str, Any]`, so that signature cannot be narrowed. Instead the body builds a `ModelSelectionUpdate(TypedDict, total=False)` — a partial of `ModelSelectionState`, so a typo or a wrong value type fails the check — and widens with `dict(update)` only at the return.

**Reads.** The SDK types `request.state` as plain `AgentState`, so a `_state(request)` helper casts it to `ModelSelectionState` once and every read goes through that.

Separately, on these exact lines: the comment `Run state does not outlive the run` is **wrong**, and `initial_route` is dead weight because of it. Custom middleware state channels *do* survive across runs on a thread — `plan_mode.py`'s own `before_agent` exists specifically to reset `plan_mode` so that `a stale True left in the thread state by a previous run` does not leak forward. I have not cut `initial_route` yet because that is a design change rather than a comment fix; flagging it here so it is not lost.

— Claude

agent/resources/prompts/system/pre-routed-mode.md
ramon-langchain: I am not sure which of two things you meant, so here is what changed and what I did not touch.

The `(ACTIVE)` marker is gone from the header. It was a leftover from when this section was injected only while pre-routed mode was live; the section is now static and always in the prompt, so a state marker in the heading was actively misleading — the agent learns pre-routed mode is over from the `exit_pre_routed_mode` result in the transcript instead.

The design rationale is in [OEP-0003](https://github.com/langchain-ai/open-swe/blob/ramonn/model-routing-redesign-4f7c5f/oeps/0003-pre-routed-mode.md), which now also covers why history reuse makes the sizing turn affordable and what the sizing turn costs.

What I did **not** do is make this file defer to the OEP, because I do not think it should: this is the agent's operating instruction, rendered into every prompt, and the OEP is a design record written for humans. Pointing at it would cost a `read_file` on every thread and would hand the agent justification where it needs rules.

If you meant the second thing, say so and I will do it. If you meant the first, this is done.

— Claude

mdrxy:
!keep-open

PR #2678, opened 2026-09-11
fix: retain content when offload trimming empties summary

baskaryan:
LangChain's summarization helper returns a successful-looking fallback when newline-based partial trimming discards every oversized message. That caused Open SWE to persist the fallback as a completed offload.

Use character-level partial trimming so oversized messages still contribute bounded context to the summary. If trimming cannot retain a valid message sequence, summarize the original offload partition rather than replacing it with the fallback. The existing recent-message retention and offloaded-history pointer remain unchanged.

Trace inspection was unavailable because the configured LangSmith workspace returned 403; the defect and fix were verified against current source behavior with a regression test.

Made by [Open SWE](https://openswe.vercel.app/agents/ee17ea6b-1a3b-58b6-a2db-5925b86a6011) · openai:gpt-5.6-sol (high)

### DIFF
--- a/agent/middleware/conversation_offloading.py
+++ b/agent/middleware/conversation_offloading.py
@@ -5,10 +5,12 @@ from contextvars import ContextVar
 from typing import Any, NotRequired
 
 from deepagents.middleware.summarization import (
+    DEEPAGENTS_DEFAULT_SUMMARY_PROMPT,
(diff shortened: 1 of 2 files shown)


ramon-langchain (commented):
should this be fixed upstream?

PR #2692, opened 2026-09-11
feat(agent): preserve mid-conversation OpenAI tool declarations

ramon-langchain:
### Summary

- keep `load_integration_tools` as an ordinary callable tool while recording trusted declaration positions in graph state
- serialize loaded schemas as Responses `additional_tools` developer input items for direct OpenAI calls, leaving top-level tools and non-capable providers on the existing path
- apply schema injection after model selection, fallback, authorization, and plan-mode filtering so only currently advertised tools are introduced
- preserve declaration state across replay while resetting it at each run boundary

### Limitations

- the adapter is intentionally gated to direct API-key OpenAI Responses models; Gateway, desktop OAuth, Chat Completions, and fallback providers retain existing eager schema advertising
- validation covers generated request payloads and replay ordering, not live OpenAI prompt-cache behavior

Made by [Open SWE](https://openswe.vercel.app/agents/646bb065-2f32-5c28-b48e-5d9e52af7614) · openai:gpt-5.6-sol (high)

### DIFF
--- a/agent/middleware/__init__.py
+++ b/agent/middleware/__init__.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any
 
 _MIDDLEWARE_MODULES = {
     "check_message_queue_before_model": ".check_message_queue",
+    "DynamicToolDeclarationMiddleware": ".dynamic_tools",
     "DynamicToolMiddleware": ".dynamic_tools",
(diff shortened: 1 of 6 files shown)


ramon-langchain:
no point in doing this if it doesnt work in gateway

PR #2716, opened 2026-09-12
feat: add pull request dashboard and repair actions

ramon-langchain:
Make Pull Requests → Mine show the user's open GitHub PRs with diffstats, current CI and conflict status, review indicators, and creation/update times. Add repository and status multiselect filters, sortable column headers, URL-backed navigation, manual refresh, and pagination loading feedback; retain review history under All Reviews.

Add a Fix in Open SWE action for conflicted or failing PRs that reuses an accessible coding thread or creates one, then queues a repair request against the existing PR branch.


agent/github/pull_request_status.py
```diff
@@ -359,3 +361,177 @@ async def get_pull_request_statuses(records: Sequence[object], token: str) -> li
...
+                    else "passing"
+                    if runs or statuses
+                    else "none",
+                )
+                return result
+
+        items = await asyncio.gather(*(load(item) for item in payload["items"][:100]))
+    return {
+        "pullRequests": [item for item in items if item is not None],
+        "truncated": payload.get("total_count", 0) > 100,
```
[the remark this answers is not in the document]
ramon-langchain: Resolved in 5f2dd4d: the search now takes a `page` parameter (100 per page, capped at GitHub's 1000-result search limit) and returns `nextPage` instead of `truncated`. The dashboard fetches the next GitHub page automatically when the user pages past the loaded rows, and loads all pages when a status filter is active.

— Claude

ui/src/features/reviews/MyPullRequests.tsx
```diff
@@ -0,0 +1,545 @@
...
+                            <ul className="mt-1">
+                              {pr.pendingChecks.map((name, index) => (
+                                <li key={`${name}-${index}`}>{name}</li>
+                              ))}
+                            </ul>
+                          </details>
+                        )}
+                        {["Conflicted", "Failing"].includes(
+                          overallStatus(pr)
+                        ) && <FixPullRequest pr={pr} />}
```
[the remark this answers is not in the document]
ramon-langchain: Resolved in dcf3ac8: the repair action is now gated directly on merge conflicts and failing CI (`pr.mergeable === false || pr.mergeState === "dirty" || pr.ci === "failing"`), independently of the Draft display label, so a conflicted or failing draft can queue a repair. The same predicate now also gates the new bulk Fix action.

— Claude

agent/threads/pr_fixes.py
```diff
@@ -0,0 +1,203 @@
...
+from agent.utils.thread_pr_state import agent_thread_pr_state_lock
+
+
+class OpenPullRequestThreadRequest(BaseModel):
+    title: str = Field(min_length=1, max_length=1000)
+
+
+class PullRequestFixContext(BaseModel):
+    title: str = Field(max_length=1000)
+    headRef: str | None = Field(max_length=1000)
```
ramon-langchain: use snake_case

agent/dashboard/threads/pr_fixes.py
```diff
@@ -0,0 +1,203 @@
...
+    reviewDecision: Literal["approved", "changes_requested", "none"] | None
+
+
+async def _find_pr_threads(
+    owner: str, repo: str, number: int, login: str, email: str | None
+) -> list[Thread]:
+    client = langgraph_client()
+    url = f"https://github.com/{owner}/{repo}/pull/{number}"
+    candidates: dict[str, Thread] = {}
+    for query in ({"pr_url": url}, {"pr_urls": [url]}):
```
ramon-langchain: please centralize this kind of query somewhere ideally where we also write this value also why do we need to concervatively try both? also is pr_ufls: Array a contains?

ramon-langchain: I'm going to refactor a bit here

ramon-langchain:
Re: the inline review comment on `ui/src/features/reviews/MyPullRequests.tsx` about unknown CI enabling merge actions ([discussion_r4020749662](https://github.com/langchain-ai/open-swe/pull/2716#discussion_r4020749662)) — replying here because a pending review on this PR blocks inline replies from my account.

Good catch, and fixed in ad2648e9. `isMergeable()` now also requires `pr.ci === "passing" || pr.ci === "none"`, so an approved PR whose check-runs or commit-status fetch failed (`ci === "unknown"`) still shows Approved but offers no Merge, on either the row or the bulk bar. That restores the pre-change merge precondition exactly: `unknown` used to render "Status unavailable" and so was never mergeable, while `none` (no checks configured) already was.

The `&&` widening itself was deliberate and stays. A cold read of a pull request comes back `mergeable: null` because GitHub computes mergeability in the background and the first read only starts it — so the old condition discarded a known review decision and CI result to report "Status unavailable", which hid an approval on one PR and a conflict on another. A new test asserts Approved renders while both the row and bulk Merge stay unavailable under unknown CI.

— Claude

agent/github/close_pull_request.py
```diff
@@ -0,0 +1,38 @@
+"""Close a pull request using the signed-in user's GitHub permissions."""
+
+import httpx2
+from fastapi import HTTPException
+
+from agent.github.http import GITHUB_API_BASE, github_client, github_request
+from agent.github.pull_request_status import pull_request_identity
+
+
+async def close_pull_request(owner: str, repo: str, number: int, token: str) -> dict[str, bool]:
```
ramon-langchain: this return type is unacceptable

agent/github/merge_pull_request.py
```diff
@@ -0,0 +1,81 @@
...
+    ("rebase", "allow_rebase_merge"),
+)
+
+
+class MergePullRequestRequest(BaseModel):
+    sha: str = Field(pattern=r"^[0-9a-fA-F]{40,64}$")
+    merge_method: MergeMethod
+
+
+async def merge_pull_request(
```
ramon-langchain: idiotic return type

agent/github/merge_pull_request.py
```diff
@@ -0,0 +1,81 @@
...
+    owner: str, repo: str, number: int, body: MergePullRequestRequest, token: str
+) -> dict[str, bool]:
+    if pull_request_identity({"repo_full_name": f"{owner}/{repo}", "number": number}) is None:
+        raise HTTPException(422, "invalid pull request")
+    try:
+        async with github_client(token=token) as client:
+            response = await github_request(
```
ramon-langchain: what is "client"? why is this not something like `client.PUT` ?

agent/github/pull_request_status.py
```diff
@@ -359,3 +384,213 @@ async def get_pull_request_statuses(records: Sequence[object], token: str) -> li
...
+            else ""
+        )
+    identity = pull_request_identity({"repo_full_name": full_name, "number": item.get("number")})
+    if identity is None:
+        return None
+    owner, name, number = identity
+    pull = await _fetch_mergeable_pull_request(client, owner, name, number) if details else None
+    if pull is not None and _live_state(pull) != "open":
+        return None
+    result: dict[str, Any] = {
```
ramon-langchain: make a proper result type

agent/review/reviews.py
```diff
@@ -318,6 +319,28 @@ async def list_reviews(
     return page, len(summaries) > offset + limit
 
 
+async def get_review_summaries(identities: list[tuple[str, str, int]]) -> dict[str, Any]:
```
ramon-langchain: make real types

agent/review/routes.py
```diff
@@ -92,6 +101,35 @@ async def api_list_review_styles(
...
+
+class ReviewSummariesRequest(BaseModel):
+    pullRequests: list[ReviewSummaryRef] = Field(max_length=100)
+
+
+@router.post("/reviews/summaries")
+async def api_get_review_summaries(
+    payload: ReviewSummariesRequest,
+    session: dict[str, Any] = SESSION_DEP,
+) -> dict[str, Any]:
```
ramon-langchain: result type

agent/review/routes.py
```diff
@@ -114,6 +152,41 @@ async def is_accessible(summary: dict[str, Any]) -> bool:
     return {"reviews": reviews, "page": page, "has_more": has_more}
 
 
+@router.get("/reviews/{owner}/{repo}/{pr_number}/thread-status")
+async def api_pull_request_thread_status(
+    owner: str, repo: str, pr_number: int, session: dict[str, str] = SESSION_DEP
+) -> dict[str, bool]:
```
ramon-langchain: result type

agent/review/routes.py
```diff
@@ -114,6 +152,41 @@ async def is_accessible(summary: dict[str, Any]) -> bool:
...
+
+
+@router.post("/reviews/{owner}/{repo}/{pr_number}/thread")
+async def api_open_pull_request_thread(
+    owner: str,
+    repo: str,
+    pr_number: int,
+    body: OpenPullRequestThreadRequest,
+    session: dict[str, str] = SESSION_DEP,
+) -> dict[str, str]:
```
ramon-langchain: fix all result type issues - we want strong result types

agent/threads/pr_fixes.py
```diff
@@ -0,0 +1,203 @@
...
+    reviewDecision: Literal["approved", "changes_requested", "none"] | None
+
+
+async def _find_pr_threads(
+    owner: str, repo: str, number: int, login: str, email: str | None
+) -> list[Thread]:
+    client = langgraph_client()
+    url = f"https://github.com/{owner}/{repo}/pull/{number}"
+    candidates: dict[str, Thread] = {}
+    for query in ({"pr_url": url}, {"pr_urls": [url]}):
```
ramon-langchain: why can't we use the pull requests postgres table?

agent/threads/pr_fixes.py
```diff
@@ -0,0 +1,203 @@
...
+    context: PullRequestFixContext | None = None,
+) -> dict[str, str | bool]:
+    full_name = f"{owner}/{repo}"
+    if pull_request_identity({"repo_full_name": full_name, "number": number}) is None:
+        raise HTTPException(422, "invalid pull request")
+    await require_repo_access_for_user(login, full_name)
+    await _ensure_dashboard_github_token(login)
+    url = f"https://github.com/{full_name}/pull/{number}"
+    client = langgraph_client()
+    prompt = (
```
ramon-langchain: prompts should always go in prompt md files, update the root AGENTS.md to require this

agent/github/close_pull_request.py
ramon-langchain: Fixed in 4acee49c. `close_pull_request` now returns a `ClosePullRequestResult` model (`closed: bool`) instead of `dict[str, bool]`.

— Claude

agent/github/merge_pull_request.py
ramon-langchain: Fixed in 4acee49c. `merge_pull_request` returns `MergePullRequestResult` and `repository_merge_methods` returns `RepositoryMergeMethods` (`merge_methods: list[MergeMethod]`).

— Claude

agent/github/pull_request_status.py
ramon-langchain: Fixed in 4acee49c. The dict literal is now an `OpenPullRequest` model, and `list_open_pull_requests` returns `OpenPullRequests`. Added `CheckState` and `ReviewDecision` Literal aliases so `ci` and `reviewDecision` are no longer bare strings.

— Claude

agent/review/reviews.py
ramon-langchain: Fixed in 4acee49c. `get_review_summaries` returns `dict[str, ReviewSummary | None]` with real `ReviewSummary` and `ReviewCounts` models. A malformed summary now logs and yields `None` for that entry rather than failing the whole batch.

— Claude

agent/review/routes.py
ramon-langchain: Fixed in 4acee49c — the route is annotated with the real review-summaries type, and `ReviewSummariesRequest.pullRequests` became `pull_requests` with a camelCase validation alias.

— Claude

agent/review/routes.py
ramon-langchain: Fixed in 4acee49c. `api_pull_request_thread_status` returns a `PullRequestThreadStatus` model (`running: bool`) instead of `dict[str, bool]`.

— Claude

agent/review/routes.py
ramon-langchain: Fixed in 4acee49c across the surface this PR adds: close, merge, repository merge methods, the open-PR list and its rows, review summaries, and the thread-status / thread / fix results are all Pydantic models now. No `dict` return annotations remain in `pull_request_dashboard_routes.py`, and the wire format is unchanged — the models carry the existing JSON names as aliases, verified key-for-key against `ui/src/lib/api.ts` and end to end (11/11 Playwright green).

Deliberately left alone, because you also flagged this PR as too big: `list_reviews` / `api_list_reviews`, `_thread_review_summary`, `get_review`, `get_review_diff`, `list_review_comments`, `create_review_comment` and the `_fetch_*` GitHub helpers all predate this PR and returned loose dicts before it. `_thread_review_summary` in particular is shared with `list_reviews` and `get_review`, so typing it converts those too. Happy to do that sweep as its own PR.

— Claude

agent/threads/pr_fixes.py
ramon-langchain: Fixed in 4acee49c. All eight fields are snake_case now (`head_ref`, `head_sha`, `merge_state`, `failing_checks`, `pending_checks`, `status_available`, `updated_at`, `review_decision`), with camelCase *validation* aliases and `populate_by_name=True` so the browser's existing payload still validates and OpenAPI still documents the request body as camelCase. Validation-alias-only rather than a plain alias generator, because the latter makes the synthesized keyword the alias and then every snake_case construction site fails type checking.

— Claude

agent/threads/pr_fixes.py
ramon-langchain: We can, and now do — fixed in 4acee49c. `_find_pr_threads` no longer scans thread metadata: it loads the `pull_request` row and reads `linked_threads()`, which backfills once per PR for rows predating the table, and falls back to `discover_threads()` only if the registry itself errors. That leaves one hand-rolled scan in the codebase, inside the table.

One gap this exposed: the fix route *creates* threads, and once backfill has run for a PR the legacy scan never runs again — so a thread created by Fix or Open would have been invisible on the next call. It now links the thread it creates (`source="dashboard_pr_fix"`), non-fatally.

Authorization is unchanged. The table returns only thread ids, so each is fetched and still passes the same two gates in the same order: the `kind`/`graph_id` filter, then `_assert_thread_postable`. `discover_threads()` applies no auth of its own, which is exactly why the filtering stayed on our side. New test asserts an admin-only link and another user's private link both stay invisible and that Fix opens a fresh thread rather than hijacking them.

— Claude

agent/threads/pr_fixes.py
ramon-langchain: Fixed in 4acee49c. Three files under `agent/resources/prompts/runs/` — `pull-request-fix.md`, `pull-request-fix-context.md`, and `pull-request-thread.md` (that last one was another inline literal in `open_pull_request_thread` that the new rule would have violated) — loaded through the existing `render_prompt` helper rather than a new mechanism. Prompt text is byte-identical, which matters because the E2E fake model routes a fix dispatch by matching its opening sentence.

AGENTS.md now carries the rule in `## Conventions`: keep prompts in markdown under `agent/resources/prompts/` and load them with `load_prompt`/`render_prompt`; never write prompt text as an inline string literal.

— Claude

agent/github/merge_pull_request.py
ramon-langchain: Pushing back on this one rather than doing it here. You're right that it reads badly — `client` is an `httpx2.AsyncClient` and passing the verb as a string is worse than `client.put(...)`. The retry, backoff and secondary-rate-limit handling all live in `github_request`, which is why everything routes through it.

Fixing it properly means changing `agent/github/http.py`'s surface — most likely having `github_client()` return a thin wrapper exposing `.get/.post/.put/.patch/.delete` that applies the retry policy. That's **32 call sites across 10 files, only 3 of which this PR touches**: `review/publish.py` has 9, `github/ci.py` 6, `github/checks.py` 3, plus `review/diff.py`, `pull_request_context.py`, `pull_request_checks.py`.

You also flagged that this PR is already too big at +5,648 lines, so bolting a repo-wide refactor of a shared module onto it seems like the wrong call — and doing it for only my 3 call sites would leave two idioms side by side, which is worse than one ugly one. I'd rather land it as its own PR right after this, where it's mechanical and reviewable in isolation and improves all 32 sites.

Say the word if you'd rather have it in here and I'll do it now.

— Claude

ui/src/features/reviews/MyPullRequests.tsx
```diff
@@ -0,0 +1,1182 @@
...
+  if (pr.detailsError) return "Could not load status"
+  if (pr.draft) return "Draft"
+  if (pr.mergeable === false || pr.mergeState === "dirty") return "Conflicted"
+  if (pr.ci === "failing") return "Failing"
+  if (pr.ci === "pending") return "Pending"
+  if (
+    !pr.statusAvailable ||
+    (pr.ci === "unknown" && pr.reviewDecision === null)
+  )
+    return "Status unavailable"
```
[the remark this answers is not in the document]
ramon-langchain: Confirming the earlier answer now that inline replies work again (a pending review on this PR was blocking them, so this went to the issue thread at the time).

Fixed in ad2648e9: `isMergeable()` also requires `pr.ci === "passing" || pr.ci === "none"`, so an approved PR whose check-runs or commit-status fetch failed (`ci === "unknown"`) still shows Approved but offers no Merge, on either the row or the bulk bar. That restores the pre-change precondition exactly — `unknown` used to render "Status unavailable" and so was never mergeable, while `none` (no checks configured) already was.

The `&&` widening itself stays, deliberately: a cold read returns `mergeable: null` because GitHub computes mergeability in the background and the first read only starts it, so the old condition discarded a known review decision and CI result to report "Status unavailable" — hiding an approval on one PR and a conflict on another.

— Claude

agent/threads/pr_fixes.py
```diff
@@ -0,0 +1,306 @@
...
+    full_name = f"{owner}/{repo}"
+    if pull_request_identity({"repo_full_name": full_name, "number": number}) is None:
+        raise HTTPException(422, "invalid pull request")
+    await require_repo_access_for_user(login, full_name)
+    await _ensure_dashboard_github_token(login)
+    url = f"https://github.com/{full_name}/pull/{number}"
+    client = langgraph_client()
+    prompt = render_prompt("runs/pull-request-comments.md", url=url)
+    async with agent_thread_pr_state_lock(client, f"comments:{login}:{url}"):
+        thread_id = await _find_or_create_pr_thread(
```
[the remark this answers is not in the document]
ramon-langchain: Correct, and thank you — that was a real race. Both dispatchers call `_find_or_create_pr_thread` for the *same* thread, so a separate `comments:` key let two requests read an empty lookup and create competing threads; the per-thread lock afterwards could never catch it, exactly as you describe.

Fixed in 8ce330733: all three dispatchers (`open_pull_request_thread`, `fix_pull_request`, `address_pull_request_comments`) now take `_pr_thread_lock_key(login, url)`, a single helper whose docstring records why the key must be shared so it does not drift apart again.

— Claude

agent/github/merge_pull_request.py
```diff
@@ -0,0 +1,93 @@
...
+    merged: bool
+
+
+class RepositoryMergeMethods(BaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
+
+    merge_methods: list[MergeMethod]
+
+
+async def merge_pull_request(
```
ramon-langchain: lets unify these "do something on GitHub with a PR" into out endpoint called something generic like "action-pr"

agent/github/pull_request_dashboard_routes.py
```diff
@@ -0,0 +1,110 @@
...
+async def api_close_my_pull_request(
+    owner: str, repo: str, number: int, session: dict[str, Any] = SESSION_DEP
+) -> ClosePullRequestResult:
+    token = await get_valid_access_token(session["sub"])
+    if not token:
+        raise HTTPException(401, "GitHub token unavailable, re-login required")
+    return await close_pull_request(owner, repo, number, token)
+
+
+@router.post("/my-pull-requests/{owner}/{repo}/{number}/ready")
```
ramon-langchain: call it "mark-ready"

agent/github/ready_pull_request.py
```diff
@@ -0,0 +1,108 @@
...
+  repository(owner: $owner, name: $repo) {
+    pullRequest(number: $number) {
+      id
+      isDraft
+    }
+  }
+}
+"""
+
+_READY_MUTATION = """
```
ramon-langchain: if any of these can be done through REST, prefer REST

agent/github/merge_pull_request.py
ramon-langchain: Done in c6d3202cb — one endpoint for all three:

```
POST /my-pull-requests/{owner}/{repo}/{number}/action
  {"action":"merge","sha":"<hex>","merge_method":"squash|merge|rebase"}
  {"action":"close"}
  {"action":"mark-ready"}
→ {"action":"...","done":true}
```

The body is a pydantic discriminated union on `action`, so FastAPI still validates each variant's own fields and returns a 422 naming the bad tag. `merge_pull_request.py`, `close_pull_request.py` and `ready_pull_request.py` are gone, replaced by `agent/github/pull_request_actions.py`, where a base class holds the identity check, the client, the transport/JSON error handling and the status mapping once and each action carries only its own call and its own wording. Backend net −68 lines, and the three per-action test files collapsed into one.

I named the path segment `action` rather than `action-pr` — the path already reads `/my-pull-requests/{owner}/{repo}/{number}/`, so `action-pr` would say "pr" twice. Say the word if you want the literal name.

Two things I deliberately left out of the unification: `repository_merge_methods` (a GET of repo settings, not an action — it moved into the same module but kept its own route), and the fix / address-comments / thread routes in `agent/review/routes.py`, which dispatch an agent thread rather than mutating the PR on GitHub.

— Claude

agent/github/pull_request_dashboard_routes.py
ramon-langchain: Done — it is `mark-ready` now, as the action name in the unified endpoint rather than a path segment: `{"action": "mark-ready"}`. The `/ready` route is gone. The UI type is `PullRequestActionName = "merge" | "close" | "mark-ready"`, and the bulk path uses the same spelling, so `"ready"` no longer appears anywhere.

— Claude

agent/github/ready_pull_request.py
ramon-langchain: Audited every GitHub call this PR adds. One of the two GraphQL calls here was avoidable and is now REST; the other has no REST equivalent.

**Changed (c6d3202cb).** Marking ready used to make two GraphQL calls — a `PullRequestReadyNodeId` query for the node id and draft flag, then the mutation. `GET /repos/{owner}/{repo}/pulls/{number}` returns both `node_id` and `draft`, so that first query is gone and the lookup is REST. The E2E fake lost its `PullRequestReadyNodeId` branch with it.

**Still GraphQL, unavoidably.** `markPullRequestReadyForReview` — REST has no way to clear the draft flag. There is no `PATCH .../pulls/{n}` field for it and no dedicated endpoint; the mutation is the only route. That is the one line of comment left on it in the source.

**Also checked and already REST.** `_fetch_review_decision` reads `GET /pulls/{n}/reviews`. Merge is `PUT .../pulls/{n}/merge`, close is `PATCH .../pulls/{n}`, merge methods is `GET /repos/{owner}/{repo}`.

**The one other GraphQL query this PR adds** is `PullRequestThreadCount`, for the unresolved-conversation count. REST's review-comment endpoints expose no resolved/unresolved state at all — `reviewThreads { nodes { isResolved } }` is the only source, and `totalCount` cannot substitute since it counts resolved threads too. (`_REVIEW_THREADS_QUERY` above it predates this PR and is GraphQL for the same reason.)

— Claude

langchain-infra (approved):

PR #2719, opened 2026-09-13
feat(slack): stage uploaded Slack files in the thread sandbox

langchain-infra:
Slack file uploads never reached the agent. `agent/slack/client.py` rendered attachment *metadata* into the prompt and the webhook collected `url_private` only for images, so a request like "analyze this zip" had nothing to read and the agent had to ask for an external download URL instead.

## What changed

- `download_slack_file()` in `agent/slack/client.py` fetches a Slack-hosted `url_private` with the bot token. It reuses `request_with_safe_redirects`, so DNS pinning and per-hop host validation still apply, and the token is only attached while both the original and current URL are Slack-owned hosts — a redirect off `files.slack.com` drops it. Payloads are capped at 50 MB.
- The mention path in `agent/slack/webhook.py` collects non-image attachments from the messages already in context (capped at 10, deduped by URL), downloads them, and stages them in the thread's sandbox under `/workspace/.open-swe/slack-files/`. Filenames are sanitized against traversal and collisions.
- A `## Slack File Attachments` section is appended to the operational context listing each sandbox path, so the agent reads the files by path with its normal tools.

Staging is best-effort: an unreachable sandbox, a failed download, or a failed upload skips that file and logs it rather than failing the run, so a message whose text is actionable on its own still works.

## Required Slack app change

This needs the `files:read` bot scope, which the documented manifest did not request. It is added to the manifest in `docs/INSTALLATION.md` along with a note that existing installations must add the scope under **OAuth & Permissions** and reinstall the app before attachments reach the agent. Without the scope, behavior is unchanged from today: uploads stay invisible and only message text is used.

Made by [Open SWE](https://openswe.vercel.app/agents/ed84061a-461e-5858-a999-c6ae45672eb8) · openai:gpt-5.6-sol (high)


agent/slack/client.py
```diff
@@ -820,6 +820,61 @@ async def upload_slack_thread_file(
         return None, error
 
 
+SLACK_FILE_DOWNLOAD_MAX_BYTES = 50 * 1024 * 1024
+
+
+async def download_slack_file(url: str) -> tuple[bytes | None, str | None]:
```
ramon-langchain: why not throw? python doesn't generally use error returns

agent/slack/webhook.py
```diff
@@ -266,6 +268,114 @@ async def _format_slack_run_links_section(thread_id: str) -> str:
 
 _OPEN_SWE_SENDER_ID = "system:open-swe"
 
+_SLACK_FILE_DIR = "/workspace/.open-swe/slack-files"
+_MAX_SLACK_FILE_ATTACHMENTS = 10
+
+
+def _slack_file_entries(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
ramon-langchain: can we make strong types here instead of dict any?

agent/slack/webhook.py
```diff
@@ -266,6 +268,114 @@ async def _format_slack_run_links_section(thread_id: str) -> str:
...
+            if url in seen:
+                continue
+            seen.add(url)
+            entries.append(file_info)
+            if len(entries) >= _MAX_SLACK_FILE_ATTACHMENTS:
+                return entries
+    return entries
+
+
+def _sanitize_slack_filename(name: Any, url: str) -> str:
```
ramon-langchain: is name really Any?

ramon-langchain (approved):

PR #2723, opened 2026-09-13
refactor(incidents): drop the responder email allowlist

langchain-infra:
Stacked on #2497.

Removes `OBSERVABILITY_AUTHORIZED_EMAILS`. Incident channels are public internal channels and dashboard sign-in is already restricted to the allowed GitHub orgs and users, so the extra list mainly produced silent failures when an on-caller who was not on it tried to pause the bot.

## Access after this change

- **Dashboard**: any signed-in user can read incidents, postmortem summaries, and history. Incident settings stay admin-only.
- **Slack controls**: anyone in an incident channel can `pause`, `resume`, `complete`, or `reopen` it, and the native agent stop pauses it. Turning the bot off never requires a linked account.
- **Questions and manual starts**: mentioning the bot with a question, `@openswe incidents`, or `/openswe incidents` requires a connected Open SWE account, the same rule as mentioning Open SWE anywhere else, because an explicit request unlocks the agent's tools. Unlinked users get the standard account-link prompt in the thread.

The intro message now says anyone can turn it off. Docs updated accordingly.


langchain-infra:
Folded into #2497 so the incidents PR carries its own access rule. The separate PR off main only removes the dead references there.

PR #2726, opened 2026-09-13
fix: retry transient integration tool load failures

### LINKED ISSUE #2453: Transient integration-loader failures disable tools until process restart

rmnvg:
Fixes #2453

## Problem

`DynamicToolMiddleware._resolve` treated any loader exception as a final answer:

```python
except Exception:
    logger.warning("Failed to load %s integration tools", group, exc_info=True)
    tools = []
resolved.tools = {tool.name: tool for tool in tools}
resolved.done = True
```

`done = True` is the cache's "never ask again" flag, so the first failed handshake was stored as the permanent result for the process. Every later `_build` short-circuited on the cached empty set and the loader was never called again. A momentary MCP, credential, or network outage disabled that integration's tools for every thread handled by the process until it was restarted, with nothing but a single warning to show for it.

## Change

Only a successful load marks a group resolved. A failure now leaves the group unresolved and parks it behind an expiring cooldown, so the next request after the cooldown retries the handshake.

The cooldown is an exponential backoff — `RETRY_BASE_SECONDS` (5s) doubling to a `RETRY_MAX_SECONDS` (60s) ceiling — so a service that is genuinely down doesn't take a retry on every model call, while a blip recovers on its own within seconds. A successful load resets the failure count. The double-checked lock now tests the cooldown as well, so concurrent resolvers for the same group still make at most one attempt per cooldown window.

The warning log moved to a static message with the group, failure count, and retry delay in `extra`, matching the structured-logging convention in AGENTS.md.

`clock` is an injectable keyword-only constructor parameter defaulting to `time.monotonic`, which keeps the backoff tests deterministic and free of sleeps.

## Tests

Two regression tests in `tests/middleware/test_dynamic_tools.py`:

- `test_a_transient_load_failure_is_retried_after_a_backoff` — a loader that fails once then succeeds. Asserts the group is reported unavailable on the first call, that a call inside the cooldown does not re-invoke the loader, and that a call after the cooldown loads the tool successfully.
- `test_repeated_load_failures_back_off_up_to_a_ceiling` — asserts the cooldown grows after repeated failures, and that one ceiling-length wait always earns a fresh attempt however many times the group has failed, pinning the upper bound.

Both fail against the previous behavior (`assert 1 == 2` — the loader is never re-invoked) and pass with the change.


wreckavp:
### Triage Assessment

**Type:** Bug fix (integration resilience)
**Severity:** Medium-High

**Summary:** Fixes #2453 by changing the integration loader to use exponential backoff retry instead of permanently caching failures. A transient MCP/network/credential failure now retries with a cooldown (5s → 60s ceiling) rather than requiring a process restart.

**Duplicate check:** No duplicates — this is the fix for #2453, not a separate issue.

**Suggested labels:** bug, integration, priority-medium

**Priority:** Medium-High — prevents permanent tool outage from transient failures.

PR #2727, opened 2026-09-13
fix: drop conversation_offloading state key that crashed parallel subagents

ramon-langchain:
## Problem

Runs with parallel `task()` subagents could crash mid-step with:

```
langgraph.errors.InvalidUpdateError: At key 'conversation_offloading': Can receive only one value per step.
```

## Root cause

#2634 (Sep 10) added `conversation_offloading` to `OffloadingState` as a plain state key, so LangGraph backed it with a single-writer `LastValue` channel. The middleware is mounted on the main agent **and** the general-purpose subagent, and deepagents' `task()` returns the subagent's final state to the parent, filtering only `_EXCLUDED_STATE_KEYS` and `PrivateStateAttr` fields — `conversation_offloading` was neither. When two or more subagents finished in the same superstep, the channel received two values in one step and LangGraph raised `InvalidUpdateError`.

## Fix

Drop the state key rather than privatize it: nothing read `conversation_offloading` from graph state (agent, dashboard, tests, UI all confirmed — the UI consumes the custom stream event from `get_stream_writer()`, which is untouched). With no channel there is nothing for parallel subagent hand-off to collide on. This also deletes the unnecessary `OffloadingState` subclass and duplicated status plumbing in `awrap_model_call` / `abefore_model`; tests now assert the stream-event payload and that graph state carries no such key.

🤖 See my full work history and reasoning in [Open SWE](https://openswe.vercel.app/agents/fced141c-90c5-5f0b-b7cb-d19b1a539134)

Made by [Open SWE](https://openswe.vercel.app/agents/418abcae-7fa4-54ed-a9a3-c859fed3264b) · fireworks:accounts/fireworks/models/glm-5p3-flash (high)


wreckavp:
### Triage Assessment

**Type:** Bug (parallel subagents)
**Severity:** Medium

**Summary:** Parallel task() subagents crash with InvalidUpdateError because conversation_offloading is a plain state key backed by a single-writer channel. The fix drops the state key entirely since nothing reads it from graph state.

**Duplicate check:** No duplicates found. This is a regression from #2634.

**Suggested labels:** bug, parallel, priority-medium

**Priority:** Medium — breaks parallel subagent workflows.

PR #2730, opened 2026-09-13
feat(slack): deliver feedback prompts as DMs instead of ephemeral messages

mdrxy:
## Why

Slack feedback prompts were sent as ephemeral messages via `chat.postEphemeral`. Per Slack's docs, ephemeral messages never show as unread, don't persist across reloads/devices, and can't be retrieved via API — so users (e.g. Mason in #team-open-swe) were never notified and skipped the prompt entirely. Slack's own guidance says to DM the user in this situation.

## What

- Add `agent/slack/dm.py` with `open_slack_dm_channel` (`conversations.open`) and `post_slack_dm` (`chat.postMessage` to the DM channel).
- `post_slack_feedback_prompt` now DMs the Good/Bad prompt (same Block Kit buttons) to the thread initiator instead of posting an ephemeral message in the thread.
- The "✅ Feedback completed. Thanks!" confirmation is also delivered as a DM so it's actually seen.
- Updated `tests/slack/test_slack_thread_feedback.py` to mock/verify the DM calls.

## Notes

The prompt/confirmations are now visible in the user's DM with the app and produce normal unread badges/notifications. Existing dedup, lock, and dismiss logic is unchanged.

Made by [Open SWE](https://openswe.vercel.app/agents/e7510703-e6ea-5879-84a6-2436ac1c1caf) · fireworks:accounts/fireworks/models/glm-5p3-flash (high)


wreckavp:
### Triage Assessment

**Type:** Bug/Improvement (Slack integration)
**Severity:** Medium

**Summary:** Feedback prompts were sent as ephemeral messages that users never saw. The fix changes to DM delivery so users get proper notifications. This is a user experience fix.

**Duplicate check:** No duplicates found. Related to #2752 and #2738 (both about Slack reply reliability).

**Suggested labels:** bug, slack, good-first-issue

**Priority:** Medium — users were missing feedback prompts entirely.

PR #2733, opened 2026-09-14
perf(dashboard): let the search do the sidebar's filtering

langchain-infra:
Stacked on #2732. The thread list filters on "unresolved", "in this repo", "has no repo" and "not an automation". The platform's thread search matches metadata by JSONB containment only — no OR, no negation, no case folding, and no way to ask for an absent key — so none of those four is expressible as a query, and `_collect_thread_candidates` pages through threads and filters them in Python instead. That is the whole reason one sidebar load cost 295 sequential `POST /threads/search` calls and 27s of TTFB.

So make them expressible. `agent/utils/thread_filters.py` derives three keys from metadata a thread already has:

| key | value | replaces |
|---|---|---|
| `repo_key` | `owner/name`, lower-cased, `""` when there is no repo | `repo=` and `ownerless=true`, both post-filtered today |
| `is_automation` | bool | `scope=interactive`/`automation`, which is a negation over `thread_category`, `source` and `schedule_id` |
| `resolved` | bool, always present | `resolved=false`, which today means "key absent or false" |

`repo_key` is folded to lower case because repo metadata is compared case-insensitively everywhere else and containment cannot do that; `is_automation` is a boolean rather than a category because `scope=interactive` has to keep matching `issue` and `pull_request` threads.

`derive_filter_metadata()` is idempotent, so every path that writes thread metadata applies it to whatever view it holds: dashboard runs, private continuations, review chats, schedules, Slack `start_new_thread`, and `ensure_thread_metadata` — which covers every Slack, GitHub and Linear thread on both the create and update sides, so the keys follow a thread that later gains a repo or is resolved. The remaining `threads.create` calls are locks and bare stubs with no participants, which never reach a thread list.

**Rollout.** A filter pushed into the search would hide every thread written before the key existed, so the pushdown is off until `scripts/backfill_thread_filter_metadata.py` has stamped the existing threads and recorded that it finished. `filter_pushdown_enabled()` reads that record (cached for a minute, falling back to the scan on any store error) and `_search_metadata_filter` only pushes filters down when it is set. Run `--dry-run` first for a count. The Python post-filters stay in place either way, so the pushdown can only ever narrow what the search returns, never change what comes back.

After the backfill, each sidebar request is one indexed containment search with `LIMIT 11` instead of a paged scan, and `/threads/projects` walks only the caller's unresolved interactive threads instead of everything they have ever touched.

Made by [Open SWE](https://open-swe-staging-ff145e5fff2951c49ece1111c2175d16.us.langgraph.app/agents/7bee86b6-c4d2-50f7-870d-c5a89d08b200) · anthropic:claude-opus-5 (high)


wreckavp:
### Triage Assessment

**Type:** Performance improvement (dashboard)
**Severity:** Medium-High

**Summary:** Fixes severe sidebar performance by replacing Python-side filtering with actual search queries. The old approach paginated through all threads (295 POST calls, 27s TTFB). The fix adds derived filter metadata keys (repo_key, is_automation, resolved) and enables pushdown to the database search with a backfill script.

**Duplicate check:** No duplicates found. This is a significant performance improvement.

**Suggested labels:** performance, dashboard, enhancement

**Priority:** High — resolves 27s sidebar load times.

Issue #2735, opened 2026-09-14
Security Audit: Hardcoded Secrets, Input Handling, and Configuration Issues


wreckavp:
# Security Audit Findings

### Finding 1: `.env` file committed to repository
**Severity:** Critical  
**File:** `.env` (root)  
**Description:** The `.env` file contains production secrets including:
- `GITHUB_TOKEN` - GitHub personal access token
- `GITHUB_APP_CLIENT_SECRET` - GitHub App client secret  
- `GITHUB_APP_PRIVATE_KEY` - RSA private key (PEM format)
- `DASHBOARD_JWT_SECRET` - JWT signing secret
- `TOKEN_ENCRYPTION_KEY` - Fernet encryption key for OAuth tokens
- `LANGSMITH_API_KEY` - LangSmith API key
- Multiple API keys (Anthropic, OpenAI, etc.)

**Remediation:**
1. Remove `.env` from git immediately: `git rm --cached .env`
2. Rotate all exposed credentials
3. Add `.env` to `.gitignore` (already present, but verify)
4. Consider using a secrets manager instead of dotenv files

---

### Finding 2: Local sandbox has no isolation
**Severity:** High  
**File:** `agent/sandboxes/providers/local.py`  
**Description:** The local sandbox provider explicitly states:
```python
"""Create a local shell sandbox with no isolation.
WARNING: This runs commands directly on the host machine with no sandboxing.
Only use for local development with human-in-the-loop enabled.
"""
```
The sandbox inherits environment variables and runs without containerization, allowing potential:
- Host filesystem access
- Environment variable leakage
- Arbitrary command execution on host

**Remediation:**
1. Disable local sandbox in production configurations
2. Use isolated sandboxes (e2b, daytona, runloop) for all non-local deployments
3. Add runtime validation to prevent accidental production use

---

### Finding 3: Open redirect vulnerability (partial mitigation)
**Severity:** High  
**File:** `agent/dashboard/oauth.py` (lines 132-156)  
**Description:** The `sanitize_redirect_to()` function validates redirects against an allowlist, but:
- Blocked paths include `/login`, `/dashboard/api`, `/_serverFn` but could miss future endpoints
- No rate limiting on failed redirect attempts
- Could be exploited if new dashboard routes are added without updating the blocklist

**Remediation:**
1. Maintain explicit allowlist rather than blocklist where possible
2. Add monitoring/alerting for redirect failures
3. Regular security review of redirect handling when adding new routes

---

[...]

PR #2736, opened 2026-09-14
security: harden sandbox secret exclusion, redirect path validation, and CORS headers (#2735)

AlphaRex-pixel:
---

## Summary

Addresses three of the code-level findings from #2735 (secrets committed to the repo and JWT algorithm choice are process/architecture changes tracked separately, not included here).

## Changes

**`agent/sandboxes/providers/local.py`** — `LOCAL_SHELL_ENV_EXCLUDE` was a manually maintained set covering 10 keys, while `agent/config.py`'s `ENV` registry marks 27 variables `secret=True` (including `DASHBOARD_JWT_SECRET`, `TOKEN_ENCRYPTION_KEY`, `GITHUB_APP_CLIENT_SECRET`). Any secret not in the manual list leaked into the unsandboxed local shell's environment. Now derives the exclusion set from the registry directly, so a newly declared secret is excluded automatically instead of silently leaking until someone remembers to add it here.

**`agent/dashboard/oauth.py`** — `_is_blocked_redirect_path()` matched the raw, undecoded path. A percent-encoded or dotted variant (e.g. `/%2e%2e/login`, `/foo/../dashboard/api/x`) could bypass the blocklist while still resolving to a blocked path once a browser normalizes/decodes it client-side. Now decodes and normalizes the path before matching.

**`agent/api/app.py`** — `allow_headers=["*"]` combined with `allow_credentials=True` causes `CORSMiddleware` to reflect back whatever headers a request asks for in `Access-Control-Request-Headers`, rather than literally sending a wildcard. Scoped to `Content-Type` and `Authorization`, the only headers the dashboard frontend actually sends (per `ui/src/lib/api.ts`).

## Not included

- Path traversal in `output_iframe.py` (finding #6): investigated, already mitigated via `realpath` resolution + `commonpath` confinement in `resolve_sandbox_file()`. No change needed; existing tests (`test_create_sandbox_file_download_url.py`) cover this.
- JWT HS256→RS256 migration (finding #5): real architecture change requiring coordinated key rotation and backward compatibility for live sessions — worth its own issue/PR rather than bundling here.
- `.env` secrets and dependency scanning (findings #1, #8): process changes, not code.

## Testing

- `ruff check` / `ruff format --check` — clean
- `ty check agent tests` — clean (3 pre-existing warnings elsewhere, unrelated to this change)
- `tests/dashboard/test_dashboard_oauth_redirect.py` — 13/13 passed
- `tests/sandbox/test_local_integration.py`, `tests/tools/test_create_sandbox_file_download_url.py` — 22/22 passed
- Full suite (`pytest tests/`) — 2710 passed, 73 skipped, 0 failed

Ref #2735


wreckavp:
### Triage Assessment

**Type:** Security fix (follow-up to #2735)
**Severity:** High

**Summary:** Addresses three code-level findings from #2735:
1. Hardcodes a manual exclusion list for local shell env vars — now derives from the config registry so new secrets are auto-excluded
2. Fix percent-encoded bypass in redirect path validation
3. Scoped CORS headers from wildcard to just Content-Type and Authorization

**Duplicate check:** This is a follow-up PR to #2735, not a duplicate. It addresses specific remediation items from the broader audit.

**Suggested labels:** security, bug

**Priority:** High — directly addresses critical security findings from #2735.

PR #2737, opened 2026-09-14
feat: split fast routing 50/50 between GLM 5.3 Flash and Luna

ramon-langchain:
## Why

The fast routing tier has been running on GLM 5.3 Flash, which has been flaky. Luna is ranking at the top of the daily leaderboards, so this adds an A/B experiment: half of fast-classified turns go to GPT-5.6 Luna (direct OpenAI) so the two can be compared under real traffic before committing.

## What

- **Fast-route A/B split** (`agent/middleware/model_selection.py`): after the routing classifier picks `fast`, the middleware consults a deterministic SHA-256 bucket of the thread id and may return a new `fast_alt` route — each thread always lands on the same side, and the selection is tagged as `open_swe_model_route` on the LangSmith run metadata and every AI message's response metadata. The split only applies when a `fast_alt` model is actually configured, and the probability is injected from team settings.
- **Team settings** (`agent/dashboard/team_settings.py`): new `default_agent_routing_fast_alt_model` / `..._reasoning_effort` / `..._fast_alt_probability` settings. Defaults turn the experiment on: `openai:gpt-5.6-luna` (direct OpenAI) at high effort, matching GLM's effort with a 0.5 probability. Setting the probability to 0 (or clearing the alt model) disables the experiment with no code change. `get_team_agent_routing_models` only exposes the `fast_alt` route when the probability is > 0, and the swagger contract is regenerated.
- **UI/API types** (`ui/src/lib/api.ts`): the new settings fields are part of `TeamSettings`.

## Test plan

- New tests in `tests/middleware/test_model_selection.py`: fast route with `fast_alt_probability=1.0` routes to the alt model; no `fast_alt` entry in the model map keeps the plain fast route.
- 616 tests across `tests/middleware`, `tests/dashboard`, `tests/agent/test_agent_assembly_context.py`, and `tests/models` pass; ruff format/lint and `ty` are clean.

Made by [Open SWE](https://openswe.vercel.app/agents/d7194397-1c04-55b3-ac0b-fe7eb19c9b8e) · fireworks:accounts/fireworks/models/glm-5p3-flash (high)


langchain-infra (approved):

agent/middleware/record_run_usage.py
```diff
@@ -11,6 +12,24 @@
...
+
+        run_tree = get_current_run_tree()
+    except Exception:  # noqa: BLE001
+        run_tree = None
+    if run_tree is None:
+        return
+    try:
+        run_tree.metadata.update(extra_metadata)
+    except Exception:  # noqa: BLE001
+        logger.debug("Could not tag run metadata", exc_info=True)
```
eyurtsev, line 31: Feels like a .error

PR #2738, opened 2026-09-14
fix: require an explicit Slack response disposition

ramon-langchain:
## Summary
Require Slack invocations to finish with either a successful final `slack_thread_reply` or `no_slack_response_needed(reason)`, rather than silently completing with an undelivered assistant answer.

- Distinguish progress acknowledgements from final replies and track dispositions per invocation.
- Restrict missing-disposition repair to the response tools, with at most two repair calls before visibly failing the run.
- Respect trusted surface handoffs and recheck the active surface before sending; metadata lookup failures fail closed.
- Preserve intentional silence for Slack wakeups while requiring stop summaries to deliver a final reply.
- Cover delivery, explicit silence, repair bounds, tool failures, and surface transitions with focused regression cases.

Made by [Open SWE](https://openswe.vercel.app/agents/5bff850b-2b0f-5217-be34-0da0fce09fc2) · openai:gpt-6-astra (low)


wreckavp:
### Triage Assessment

**Severity:** Medium

**Summary:** Requires Slack invocations to finish with either a successful slack_thread_reply or no_slack_response_needed. Prevents silent completion with undelivered answers. Includes repair logic with bounds (max 2 repair calls) and respects surface handoffs.

**Duplicate check:** Related to #2752 (post-run safety net for missing replies) — #2752 is a reactive fix, this is a proactive enforcement. Both address the same root problem but from different angles. Could coordinate.

**Suggested labels:** bug, slack, priority-medium

**Priority:** Medium-High — prevents silent answer loss in Slack.

agent/middleware/slack_response_disposition.py
```diff
@@ -0,0 +1,276 @@
...
+from agent.utils.json_types import thread_metadata
+from agent.utils.thread_ops import langgraph_client
+
+logger = logging.getLogger(__name__)
+
+_FINAL_TOOLS = frozenset({"slack_thread_reply", "no_slack_response_needed"})
+_MAX_REPAIR_ATTEMPTS = 2
+_REPAIR_INVOCATION_KEY = "open_swe_slack_disposition_repair"
+_REPAIR_ATTEMPT_KEY = "open_swe_slack_disposition_repair_attempt"
+_REPAIR_PROMPT = (
```
ramon-langchain, line 31: I thought we said prompts should go into md files? please add this to root AGENTS.md

ramon-langchain:
Closing in favour of #2895, which carries the reply surface as thread state rather than re-reading thread metadata on every reply.

The `response_type` progress/final split from here is ported over — it caught a real hole: the Slack prompt orders an acknowledgement before any investigation, so a requirement satisfied by *any* successful reply is satisfied by the ack alone. The classification is slightly different there: a blocking question and an approval request are `final`, since they leave the asker holding the ball and the turn genuinely ends. The wakeup-silence and stop-summary prompt rules came over as well.

Not ported: the per-call `threads.get()` surface recheck (the state channel answers it without the round trip, and failing closed on an unreadable lookup reproduces the silence it is meant to prevent), and failing the run after the repair budget — #2895 posts the model's own final text instead, so the asker gets the answer.

PR #2750, opened 2026-09-14
fix: resolve scheduled-run actor from the automation's creator

baskaryan:
## Problem

Scheduled (cron) `agent` runs carry no triggering-user identity in their run config, so `_actor()` in `agent/tools/threads.py` resolves to `None` and every actor-gated thread tool (`list_threads`, `get_thread`, `manage_thread`) short-circuits with `{"success": false, "error": "No verified triggering user is available"}`. Scheduled automations cannot inspect their own thread, check prior runs for dedup, or manage plans.

## Root cause

#2647 moved scheduled runs to workspace credentials and deliberately dropped `github_login`/`user_email` from `_agent_run_config` (`agent/dashboard/schedules.py`). The thread tools' actor resolution was never given a scheduled-run identity to fall back on.

## Fix

When the run config has `source == "schedule"`, `_actor()` now resolves the actor from the schedule record's saved `created_by` login and `user_email` (still passing the GitHub login gate). Non-schedule sources are unchanged, and the schedule lookup is skipped entirely for them.

## Tests

- `tests/tools/test_threads.py`: schedule creator resolves as the actor; missing schedule record still denies; non-schedule sources never consult the schedule store.

```html
<!-- langsmith-issue {"id":"abc68b1a-5368-450a-a9da-6352abd9e0ed"} -->
```

Made by [Open SWE](https://openswe.vercel.app/agents/2b444a30-2ec9-5bb3-abb0-579d43bda59a) · openai:gpt-5.6-sol (medium)


wreckavp:
### Triage Assessment

**Type:** Bug (scheduling/automation)
**Severity:** Medium

**Summary:** Scheduled (cron) runs have no triggering-user identity, causing all actor-gated thread tools to short-circuit with "No verified triggering user." The fix resolves the actor from the schedule record's saved creator when source is "schedule."

**Duplicate check:** No duplicates found. This is a regression from #2647's scheduled-run credential changes.

**Suggested labels:** bug, scheduling, priority-medium

**Priority:** Medium — breaks all thread tool usage in scheduled automations.

PR #2751, opened 2026-09-14
fix: keep a visible sender-context block authoritative on follow-up turns

baskaryan:
## Problem

On a Slack follow-up turn from the same sender, `PrepareAgentRunMiddleware._sender_context_messages` intentionally suppresses a fresh `system:sender-context` injection when the earlier block is still visible in the thread (verified at `agent/server.py:735`). But the repository-setup prompt told the model sender context "applies to that turn only" — so the agent concluded it had no trusted git identity for the new turn, refused to implement, and blamed sender-context injection.

## Fix

One-line prompt change in `agent/resources/prompts/system/repository-setup.md`: a still-visible sender-context block for the triggering sender is now stated to be authoritative for the current turn, with the reason dedupe suppresses re-injection explained. The cross-participant safety rule ("never carry a participant's identity... to another participant's message, always use the most recent one") is preserved verbatim.

The dedupe logic itself is correct as designed — summarization-introspection via `visible_dynamic_context_hashes` already guards the case where the block fell behind a summarization cutoff. The defect was the prompt contradicting that behavior.

## Testing

- `tests/middleware/test_prepare_run_middleware.py` (15 passed) — dedupe behavior unchanged
- `tests/test_prompts.py` + `tests/agent/test_input_messages.py` (13 passed)
- `ruff check` / `ruff format --check` clean

Made by [Open SWE](https://openswe.vercel.app/agents/e4855feb-8240-5a56-b958-ceeb2d060de8) · fireworks:accounts/fireworks/models/glm-5p3-flash (high)


wreckavp:
### Triage Assessment

**Type:** Bug (prompt/agent behavior)
**Severity:** Low-Medium

**Summary:** A contradiction between the dedupe logic and the repository-setup prompt causes agents to refuse implementation on follow-up turns. The dedupe logic intentionally suppresses re-injection of sender-context when still visible, but the prompt told the model that context "applies to that turn only," leading the model to believe it had no trusted git identity.

**Duplicate check:** No duplicates found. This is a prompt-vs-logic alignment bug.

**Suggested labels:** bug, agent, good-first-issue

**Priority:** Low — narrow edge case affecting Slack follow-ups from the same sender.

mdrxy (approved):

PR #2752, opened 2026-09-14
fix(slack): post a pointer when a successful run's answer never reaches the thread

baskaryan:
## Problem

On Slack-sourced runs, the agent can finish **successfully** with its complete answer stranded in the final assistant message and never call `slack_thread_reply`. The user sees only the opening acknowledgement — a 9-minute, 48-tool-call investigation can be completely invisible. Delivery currently depends entirely on the model remembering to call the tool; nothing on the success path checks that the answer actually reached the thread. (`agent/completion.py` handles `error`/`timeout` with failure replies, but `success` only schedules cost refresh and feedback.)

## Fix

Add a post-run safety net in the run-completion webhook (`_handle_successful_run`): after a successful Slack run, if the run's state shows **no** `slack_thread_reply` tool call **and** the final AI message has non-empty text, post a short best-effort Slack notice pointing at the LangSmith trace and telling the user to re-ask to have the answer reposted.

Design notes:
- Detection reads tool calls from the run's state messages (`values.messages`) — tool calls survive artifact eviction, so they're the durable signal that the agent reached the thread. The tool's own result is deliberately not trusted: a failed post is indistinguishable from success once results are dropped.
- The notice is skipped for automated wakeups (`thread_wakeup`), runs that did call `slack_thread_reply`, and runs with empty terminal text — so normal silent operations (schedules, stop summaries, untagged observation) are unaffected.
- The webhook stays `ok` even if the notice fails to post; the safety net never breaks completion.

Verified against the two linked traces' failure mode and the current repo: the success path genuinely has no Slack-delivery check today.

Made by [Open SWE](https://openswe.vercel.app/agents/e915dd69-0f53-5b3a-82ba-4953104c93c4) · fireworks:accounts/fireworks/models/glm-5p3-flash (high)


wreckavp:
### Triage Assessment

**Severity:** Medium-High

**Summary:** On Slack-sourced runs, the agent can finish successfully but never call slack_thread_reply, leaving the answer invisible to the user. Currently the success path only schedules cost refresh and feedback — there's no check that the agent actually delivered its answer to Slack. The fix adds a post-run safety net in the completion webhook that posts a notice pointing to the LangSmith trace when delivery appears to have failed.

**Duplicate check:** No duplicates found. This is a unique reliability gap.

**Suggested labels:** bug, slack, priority-medium

**Priority:** Medium-High — users can lose complete answers without any error signal.

PR #2753, opened 2026-09-14
fix(slack): move a thread to the dashboard when its Slack thread is deleted

ramon-langchain:
Eugene asked Open SWE a question in #team-oss and deleted his message seconds later. The run kept its Slack location, and Slack accepted both replies against the deleted parent by posting them at the channel root — where they read as answers to nothing ([example](https://langchain.slack.com/archives/C09M16GJNNP/p1789397603915249)).

`chat.postMessage` does not error in that case, so nothing downstream noticed. Now a reply that comes back unthreaded is deleted, and the agent thread is detached from Slack: its Slack store mappings are cleared, `source_context.slack_thread` is dropped, and the thread becomes a dashboard thread. The reply tool tells the model the Slack thread is gone and to answer in its final response instead; `get_active_slack_thread` ignores the stale location the run config still carries, so nothing tries Slack again.

`_DASHBOARD_SOURCE` becomes `DASHBOARD_SOURCE` — it is now read outside `agent/dashboard/threads/`.


agent/slack/tools/thread_reply.py
```diff
@@ -95,6 +110,26 @@ async def slack_thread_reply(
     return {"success": True}
 
 
+async def _already_moved_to_dashboard(client: LangGraphClient, thread_id: str | None) -> bool:
+    if not thread_id:
+        return False
+    try:
+        thread = await client.threads.get(thread_id)
+    except Exception:
```
ramon-langchain, line 123: please do not just discard errors ever, add this rule to root AGENTS.md either propagate or log

agent/slack/client.py
```diff
@@ -372,6 +372,24 @@ def _log_automated_warning_sent_to_slack(
     logger.error("Sent automated warning message to Slack channel %s: %s", channel_id, text)
 
 
+def _threaded_under(data: Any, reply_ts: str) -> bool:
```
ramon-langchain: is Any really the best you can do here?

agent/slack/orphan.py
```diff
@@ -0,0 +1,84 @@
...
+from agent.store import now_iso
+from agent.utils.dashboard_links import dashboard_thread_url
+from agent.utils.json_types import thread_metadata
+
+logger = logging.getLogger(__name__)
+
+
+def slack_thread_detached(metadata: Mapping[str, object] | None) -> bool:
+    """Whether this thread has already been moved off Slack."""
+    if not isinstance(metadata, Mapping):
```
ramon-langchain: why is this needed this is the type

agent/slack/tools/thread_reply.py
ramon-langchain: Fixed in 37d436e9 — that `except` now logs with `logger.exception` before returning `False`.

Also added the rule to root `AGENTS.md`: "Never discard an error. Every `except` either propagates (re-raise, or raise a more useful error) or logs what it swallowed."

— Claude

agent/slack/client.py
ramon-langchain: No — it is `AsyncSlackResponse` from `slack_sdk.web.async_slack_response`, which is what `chat_postMessage` returns. Typed properly in 37d436e9.

— Claude

agent/slack/orphan.py
ramon-langchain: It is not. Dropped in 37d436e9 — the parameter is now `JsonObject` (what `thread_metadata()` always returns) and the `isinstance` guard is gone.

— Claude

wreckavp:
### Triage Assessment

**Severity:** Medium

**Summary:** When a Slack thread is deleted after a user asks Open SWE a question, replies get posted at the channel root instead of being dropped. The fix detaches the agent thread from Slack (clears store mappings, drops slack_thread from source_context) and falls back to dashboard-only operation. Also surfaces DASHBOARD_SOURCE outside its current module scope.

**Duplicate check:** No duplicates found. Distinct edge case in Slack lifecycle handling.

**Suggested labels:** bug, slack

**Priority:** Medium — affects user experience when Slack threads are accidentally deleted.

agent/slack/tools/thread_reply.py
```diff
@@ -85,6 +97,12 @@ async def slack_thread_reply(
             should_ask_for_feedback=should_ask_for_feedback and not options,
         )
     if message_ts is None:
+        if slack_error == "thread_not_found":
+            if thread_id:
+                await move_thread_to_dashboard(
+                    client, str(thread_id), str(channel_id), str(thread_ts)
+                )
+            return _dashboard_handoff(thread_id)
```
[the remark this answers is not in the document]
ramon-langchain: Good catch — fixed in the commit above. `slack_thread_reply` now uses the return value: a failed `move_thread_to_dashboard` returns `moved_to_dashboard: false` with `retry: true` and a hint saying the migration failed, instead of claiming the handoff completed. Added a test for that path.

— Claude

eyurtsev (approved):

PR #2761, opened 2026-09-14
feat: store pull requests and repositories in PostgreSQL

ramon-langchain:
A pull request had no representation of its own. The only way back from a PR to the threads working on it was a paginated `threads.search` over `pr_url`/`pr_urls` metadata followed by a heuristic pick (`update_agent_thread_pr_state` in `agent/webhooks/common.py`), which cannot answer *which thread is the thread* for a PR. There was no repository object either — only repo-keyed settings (`review_styles`, `enabled_review_repos`) and a pseudonymized analytics `repository_directory` that cannot be resolved back.

## Tables (migration `0011`, schema `open_swe`)

Entity rows carry a synthetic **UUIDv7 `id`**, generated client-side with Python 3.14's `uuid.uuid7()` and passed on every `INSERT` — no server-side generator, so it works on any PostgreSQL version. Natural keys stay unique so callers keep addressing a PR by `(owner, repo, number)`.

| Table | Primary key | Natural key | Holds |
|---|---|---|---|
| `repository` | `id` uuid | lowercased `owner/name` | `full_name` (GitHub casing), `private`, `default_branch`, activity timestamps |
| `pull_request` | `id` uuid | `(repository_id, number)` | state, title, head/base ref, author, `resolves_thread` |
| `pull_request_thread` | `(pull_request_id, thread_id)` | — link table | `role` primary/secondary, `source`, `linked_at` |
| `pull_request_review` | `id` uuid | GitHub review id, or `(reviewer thread, head sha)` | url, finding count, `published_at` |

## Domain objects

`Repository`, `PullRequest`, `ThreadLink`, `ReviewLink` are SQLAlchemy 2.0 mapped classes (`MappedAsDataclass`, keyword-only) and *are* the domain objects — no parallel DTO layer. They live next to the code that owns them (`agent/github/repositories.py`, `agent/github/pull_requests.py`); `agent/database/` holds only infrastructure: engine, `session()`, migrations, and the declarative `Base` (`agent/database/orm.py`). Pydantic remains only where untrusted input enters: `PullRequestEvent`/`PullRequestPayload` parse the GitHub webhook and build a `PullRequest`.

Two kinds of write, by who knows what:

- `save()` — for callers holding the PR as GitHub describes it (the opener, lifecycle webhooks). Writes the GitHub-owned columns; can set, never clear, `resolves_thread`.
- `link_thread()` / `link_review()` — for callers that know only the PR's identity. Create the row if missing, touch no other column.

[...]

PR #2779, opened 2026-09-15
feat: expedited Slack approval for tiny pull requests

ramon-langchain:
Implements [OEP-0002](oeps/0002-expedited-slack-review.md) (discussion: #2569). Built on the pull request, repository, and users tables from #2761 and #2781; both have merged, so this now targets `main` directly. The OEP draft is updated here to match what was built.

**Experimental and off by default.** An admin enables it per instance under Settings → Expedited Slack review. When off, the agent does not get the tool, new requests are refused, and open cards are withdrawn on the next evaluation.

## What it does

The agent calls `expedite_pr_approval` with a PR URL. The backend refuses anything over 10 changed lines, any file without a text patch, and any path under auth, secrets, CI workflows, dependency manifests, migrations, or key/env files. An accepted request becomes an `expedited_approval` row pinned to the head SHA and a fingerprint of the diff, watched by GitHub webhooks plus a 5-minute cron.

The card is posted in the Slack thread only when the revision is fully ready: PR open and not a draft, no conflicts, every check `success`, zero unresolved review threads, no standing `CHANGES_REQUESTED`, and, where Open SWE reviews the repo, an Open SWE review recorded for that exact SHA. Silence from a reviewer is not completion.

Voting: a Slack click resolves to a person whose GitHub identity has write access. The author may approve. Every non-author approval is submitted to GitHub as that person's own `APPROVE` review at `commit_id` before it counts. Two distinct people merge the PR with a GitHub App token scoped to `contents` and `pull_requests` write on that repo, conditional on the reviewed SHA. GitHub's refusal is final; the agent is told why. Reject opens a feedback modal, disables the card, and wakes the agent once with the feedback. A new commit, changed diff, or regressed check retires the vote and discards its approvals.

## Storage

Migration `0015`: `expedited_approval` (one active row per PR via a partial unique index) and `expedited_approval_vote` (one row per person, keyed by `users.id`). Votes never store a GitHub or Slack handle, so one person cannot vote twice through two identities; handles are resolved from their identities for display only.

## Identity resolution

`agent/users/resolve.py` is the one place a `PersonIdentity` becomes a `User`: Slack identity, then the legacy Slack→GitHub mapping (which links the Slack identity so the fallback retires itself), then GitHub numeric id, then login. Unknown people are never created. The Slack interactivity route builds the `PersonIdentity` for a click and the voting code receives a `User`. Wiring the same resolver into run dispatch (`open_swe_account`, `user_id` in config) is a follow-up.

## Typed Slack payloads

`agent/slack/blocks.py` declares the Block Kit objects Open SWE builds as TypedDicts, with builders that enforce Slack's text limits, so a malformed card is a type error rather than an `invalid_blocks` error at runtime. `block_payload` and `view_payload` are the single boundary where they widen to the dictionaries the SDK takes. `SlackViewSubmission` in `agent/slack/payloads.py` gives a modal submission a typed shape, replacing untyped digging through `view.state.values`.

## Wiring

- Tool `expedite_pr_approval`, Slack-only, excluded in plan mode, dropped when the setting is off.
- Slack interactivity: `expedited_review` button type and the reject modal submission.
- GitHub webhook: check, status, pull_request, and review events re-evaluate active approvals in that repo. The hook sits behind both gates it answers to — the repository must belong to a workspace and be allowlisted.
- Scheduler task `expedited_review` for the per-approval cron.
- Admin UI: new section with an Experimental badge.
- `aggregate_check_state` (baby-sit) and `fetch_unresolved_review_threads` (PR status) made public for reuse.

## End-to-end coverage

`tests/e2e/tests/expedited_review.spec.ts` drives the whole chain with only the LLM and the external SaaS boundaries faked: the agent opens the PR and starts watching it, a signed GitHub webhook reports a failing check, the agent fixes it, readiness is re-evaluated and the card posts, two people vote in fake Slack, each non-author approval is submitted as a real `APPROVE` review, and the PR merges conditional on the reviewed SHA.

Standing up that path needed some harness work, which is most of the `tests/e2e/` diff: real per-file `git diff` hunks behind `pulls/{n}/files`, review submit/list and a conditional merge endpoint, collaborator permission, recorded Slack ephemerals (so a refused vote is visible instead of silent), and a per-spec reset of the durable pull-request rows. `manage_baby_sit` was not reachable from the harness at all before this, because `parse_github_pr_url` only accepts `github.com`; that is patched at the test boundary rather than by loosening the production parser.

## Test Plan

- [x] `tests/slack/test_slack_blocks.py`, `tests/expedited_review/`, `tests/users/test_resolve.py`
- [x] `tests/e2e/tests/expedited_review.spec.ts`, plus the `workspaces` and `dashboard` specs it shares harness state with
- [x] `ruff check` + `ruff format --check` on `agent` and `tests`
- [x] `ty check agent tests`
- [x] `pnpm run format:check`, `pnpm typecheck` in `ui`


agent/expedited_review/card.py
```diff
@@ -0,0 +1,161 @@
...
+        _section(f"*{_escape(outcome)}*"),
+        _context(_vote_summary(approval)),
+    ]
+    return f"{outcome} — {pr.url}", blocks
+
+
+def reject_modal(approval: ExpeditedApproval, *, channel_id: str, thread_ts: str) -> dict[str, Any]:
+    pr = approval.pull_request
+    return {
+        "type": "modal",
```
ramon-langchain, line 137: can we please add proper types for any Slack things we add to the appropriate slack client module

ramon-langchain: Done in c56920986.

Added `agent/slack/blocks.py`: the Block Kit objects we build are now TypedDicts (`SectionBlock`, `ContextBlock`, `DividerBlock`, `ActionsBlock`, `InputBlock`, `ButtonElement`, `PlainTextInput`, `ModalView`, and the `PlainText`/`Mrkdwn` text objects) with builders that enforce Slack's limits — section and context text truncated to 3000 chars, button text to 75 — so an over-long card is impossible rather than an `invalid_blocks` error at runtime. `escape` and `code_block` moved there too, since neither was card-specific.

`block_payload` and `view_payload` are the single boundary where typed blocks widen to the plain dicts `agent/slack/client.py` and the SDK take. I left the existing client signatures alone: they `deepcopy` and `.get()` their way through blocks, so widening them is a bigger change than this PR should carry.

For the inbound half, `SlackViewSubmission` in `agent/slack/payloads.py` now models a modal submission (`callback_id`, `metadata` for parsed `private_metadata`, and `submitted(block_id, action_id)`), replacing the untyped digging through `view.state.values` in `agent/expedited_review/slack.py`.

`card.py` builds `list[Block]` throughout and no longer mentions `Any`. Covered by `tests/slack/test_slack_blocks.py` (truncation, fence escaping, optional-field omission, submission parsing).

— Claude

agent/expedited_review/readiness.py
```diff
@@ -0,0 +1,206 @@
...
+            logger.warning(
+                "Pull request registry unavailable while checking Open SWE review",
+                extra={"pr_repo_full_name": f"{owner}/{repo}", "pr_number": pr_number},
+                exc_info=True,
+            )
+            return None
+        reviewed_head = stored is not None and any(
+            review.head_sha == head_sha for review in stored.reviews
+        )
+
```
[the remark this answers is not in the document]
ramon-langchain: Changed, though not to a stricter aggregation — the opposite direction, per a product call from @ramon-langchain: a failing check that GitHub does not require should not block the vote, it should be named on the card so the voters approve with it in front of them.

So readiness now blocks on `pending` (a check still running) and on failures GitHub requires, and treats the rest as advisory. The required/not-required split comes from `mergeable_state`: GitHub reports `unstable` when the PR is mergeable and only checks it does not require are unhappy, and `blocked` when a required one is. Trusting its answer avoids reading branch protection, which needs admin, and avoids guessing at ruleset precedence.

That also covers the `neutral`/`skipped` case you raised, from the other side. Requiring literal `success` everywhere would have deadlocked the feature — `skipped` is simply how GitHub reports a path-filtered job, and this very repo produces both `skipped` shards and a `neutral` CodeQL run. Under the new rule a `neutral` or `skipped` check that GitHub actually requires still leaves `mergeable_state` at `blocked`, so it blocks; one it does not require no longer pretends to be a pass, it just is not a gate.

The advisory names are persisted on the approval row (`advisory_failures`) because `refresh_card` re-renders after every vote without a readiness pass of its own, and the warning has to survive those.

`readiness.py` + `card.py` in f8d062f8f, covered by three new cases in `tests/expedited_review/test_readiness.py`.

— Claude

agent/expedited_review/voting.py
```diff
@@ -0,0 +1,331 @@
...
+            await refresh_card(reopened)
+        return "open"
+
+    merge_token = await _merge_token(pr.owner, pr.repo) or token
+    methods = snapshot.allowed_merge_methods or ["merge"]
+    url = f"{GITHUB_API_BASE}/repos/{pr.owner}/{pr.repo}/pulls/{pr.number}/merge"
+    payload: dict[str, Any] = {"sha": approval.head_sha, "merge_method": methods[0]}
+    try:
+        async with github_client(token=merge_token) as client:
+            response = await github_request(client, "PUT", url, json=payload)
```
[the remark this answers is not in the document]
ramon-langchain: Correct, and fixed in f8d062f8f. Your reading of the control flow was exactly right: `evaluate_approval` returned into `complete_merge` at the `merging` branch *before* reaching the eligibility and fingerprint checks, so a `merging` row was the one state that never revalidated its diff.

Rather than duplicate the revalidation inside `complete_merge`, I hoisted the existing block — `_files_for` → `assess_eligibility` → `diff_fingerprint` comparison — above the `merging` branch, so every state now revalidates before anything acts on it. A retargeted base retires the vote as `superseded` and discards its approvals, the same as a changed diff anywhere else.

— Claude

agent/expedited_review/voting.py
```diff
@@ -0,0 +1,331 @@
...
+    if not readiness.ready:
+        reopened = await transition(
+            approval.id,
+            expected=("merging",),
+            state="open",
+            detail="; ".join(readiness.blockers),
+        )
+        if reopened is not None:
+            await refresh_card(reopened)
+        return "open"
```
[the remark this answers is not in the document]
ramon-langchain: Real, and fixed in f8d062f8f. The dead end was as you describe: `complete_merge` sends a transiently-blocked row back to `open` with both votes intact, and the `open` branch then just returned `open` forever, while both voters bounced off the "already approved" guard. It needed a third person to unstick a vote that already had quorum.

Evaluation now resumes it instead of waiting for a vote that cannot come. `_resume_quorum` runs on the `open`-and-ready path: if the row already has `REQUIRED_APPROVALS` approvals it transitions `open` → `merging` under the same row lock and completes the merge. The lock plus `expected=("open",)` keeps a concurrent webhook and cron from both resuming it, and a losing racer just returns `open`.

— Claude

agent/expedited_review/watch.py
```diff
@@ -0,0 +1,399 @@
...
+            extra={"approval_id": str(approval.id), "slack_error": error},
+        )
+        return "error"
+    updated = await transition(
+        approval.id, expected=("waiting",), state="open", slack_message_ts=message_ts
+    )
+    if updated is None:
+        await update_slack_message(
+            location[0], location[1], "This expedited review is no longer active.", blocks=None
+        )
```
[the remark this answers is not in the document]
ramon-langchain: Fixed in f8d062f8f — this was a straight bug. `location[1]` is the thread root timestamp, not the card, so the loser of a concurrent post could overwrite a root message (or fail with `cant_update_message` on a human-authored one) while leaving its own duplicate card live with working buttons. It now retracts `message_ts`, the card that pass had just posted, which is the message it actually owns. Comment on the line records why it must not be `location[1]`.

— Claude

ramon-langchain:
Noted, not acting on it for now. Both preview-conflict notices are accurate — this branch touches `tests/e2e/harness.py` and `tests/e2e/patches.py` (and earlier `agent/github/routes.py`, `agent/scheduler.py`, `agent/tools/__init__.py`), which is exactly the kind of overlap that keeps it out of a shared preview tree.

Building a `preview-manual/*` branch is the documented fix, but it means merging this PR together with whichever others it clashes with and pushing a shared branch, which is a call for @ramon-langchain rather than something to do autonomously — and nobody has asked for a preview of this yet. The feature is off by default behind an instance setting, so there is not much to see in a preview until an admin turns it on.

Leaving the `preview` label as-is; say the word and I will build the manual branch.

— Claude

agent/expedited_review/approvals.py
```diff
@@ -0,0 +1,194 @@
...
+    voted_at: Mapped[datetime | None] = mapped_column(server_default=NOW, init=False)
+    voter: Mapped[User] = relationship(init=False)
+
+    @property
+    def github_login(self) -> str:
+        """The voter's GitHub handle for display; never an identity key."""
+        return github_login_of(self.voter)
+
+
+def github_login_of(user: User) -> str:
```
ramon-langchain: move to User

ui/src/routes/admin.tsx
```diff
@@ -595,6 +598,55 @@ function FableSection() {
   )
 }
 
+function ExpeditedReviewSection() {
```
ramon-langchain: please break this out into a separate file

agent/expedited_review/approvals.py
ramon-langchain: Moved in 377b90b22. It is now `User.login_for(provider)` on the model in `agent/users/models.py`, so the handle lookup lives on the type that owns `identities` rather than in a free function in the expedited-review package.

Taking `provider` rather than hardcoding GitHub, since the same "prefer the provider handle, fall back to display name, then a short id" shape is what any caller wants for Slack too. `ApprovalVote.github_login` is now just `self.voter.login_for("github")`, and `voting.py` calls it directly instead of importing a helper.

— Claude

ui/src/routes/admin.tsx
ramon-langchain: Done in 377b90b22 — now `ui/src/features/settings/components/ExpeditedReviewSection.tsx`, next to `AllowedSlackBotsSection` and `MCPConnectionsSection`, which is where the other extracted settings sections live. `admin.tsx` is 49 lines shorter and just renders `<ExpeditedReviewSection />`; its now-unused `Badge` import is gone too.

— Claude

johannes117 (approved):

PR #2782, opened 2026-09-15
fix: propagate subagent cancellation

mdrxy:
Stopping a run now actually interrupts an in-flight subagent or other long-running tool call instead of letting it continue behind an error result.

---

### Context

Open SWE wraps tool calls with its own `ToolErrorMiddleware` so ordinary tool failures become `ToolMessage` results that the model can inspect and recover from. LangGraph signals an interrupted node with `NodeCancelledError`, which inherits from `Exception`; the middleware therefore caught the cancellation and converted it into a normal tool error. The graph no longer saw the cancellation signal, so the tool kept the run from stopping correctly. This is most visible with subagents because `task` calls commonly run much longer than other tools, but the bug affects every tool call.

### Before / after

| | Before | After |
|---|---|---|
| Stop during a tool call | `NodeCancelledError` becomes an error `ToolMessage`; execution can appear to continue | `NodeCancelledError` escapes the middleware and terminates the run |
| Ordinary tool failure | Returned to the model as a structured error | Unchanged |
| Dead sandbox | Notification is sent and the fatal error propagates | Unchanged |

### Change

- Re-raise `NodeCancelledError` before generic tool-error handling.
- Cover the cancellation boundary with a focused regression test.

Made by [Open SWE](https://openswe.vercel.app/agents/de7a163c-1649-5c23-8a73-9bd1c81b1d3f) · openai:gpt-5.6-sol (medium)


mdrxy:
!keep-open

johannes117 (approved):

PR #2799, opened 2026-09-15
feat(workspaces): store workspaces in PostgreSQL

langchain-infra:
Third and last PR of the OEP-0003 stack: workspaces move from the LangGraph Store into PostgreSQL.

- Migration 0013 adds `workspace`, `workspace_repository` (primary key on `repository_id`, referencing the `repository` table, so the database enforces single ownership) and `workspace_slack_channel` (primary key on `channel_id`); 0014 adds CHECK constraints for the refresh status and kind.
- `WorkspaceStore` keeps its public surface and the Pydantic `Workspace` shape but writes rows through `agent.database.postgres.session()` in one transaction per write; conflicts are translated by constraint name into the existing messages and answer 409, including a concurrent create of the same slug.
- Any LangGraph Store records under `["workspaces"]` or the legacy `["environments"]` namespace are imported once at startup and removed; until an import has succeeded on an empty table the GitHub route answers 503 rather than treating every repository as unowned.
- Routing looks up owners with indexed queries instead of a cached list; team settings and workspace MCP connections stay in the Store keyed by slug.
- Model options and the vision fallback resolve for the run's workspace; tests that need rows use the isolated-schema `registry_db` fixture (`TEST_ANALYTICS_POSTGRES_URI`, documented in DEVELOPMENT.md).
- The OEP records the remaining follow-ups (repository rename gap, optimistic concurrency, the public-workspace lockdown proposal).

Depends on the dashboard PR below it in the stack.


langchain-infra:
Superseded: the stack is being restructured into two PRs with PostgreSQL storage in the first one, so no Store-backed intermediate is reviewed. Replacement PR numbers follow in a comment.

langchain-infra:
Replacements: #2807 (backend with PostgreSQL storage) and #2808 (dashboard), stack #2809.

PR #2803, opened 2026-09-15
feat(threads): add a trimmed thread state view

johannes117:
## Summary

The transcript paints every user and assistant message at the top level and folds tool activity away, so the first paint needs the conversation text, not the tool outputs. On a 12 MB local thread the visible text and tool-call names were under 200 KB.

- `GET /dashboard/api/threads/{thread_id}/state?view=trimmed` returns the usual `ThreadState` shape, but tool messages whose content plus artifact exceeds 4 KB have their content blanked, and inline image blocks lose their `base64`. Each blank gets an `open_swe_stub` marker (`additional_kwargs` on the message, or on the image block) carrying the original size, and every message keeps its id so the client can fetch it individually later.
- `view=full` (the default) is byte-for-byte what the endpoint returned before.
- `Server-Timing` gains a `trim` phase plus `stubbed` and `saved_bytes` counts.

Nothing consumes the new view yet; the UI switch is a later PR in this stack.

-------------------------
- main
    - **feat(threads): add a trimmed thread state view** :point_left:
      - https://github.com/langchain-ai/open-swe/pull/2804
        - https://github.com/langchain-ai/open-swe/pull/2805


johannes117:

PR #2804, opened 2026-09-15
feat(threads): fetch a single thread message by id

johannes117:
## Summary

Companion to the trimmed state view: when the transcript paints from trimmed messages, expanding a stubbed tool output or image needs just that one message.

- `GET /dashboard/api/threads/{thread_id}/messages/{message_id}` reads the thread's latest state and returns the matching message dict in full. Unknown ids, empty ids and ids over 200 characters are 404, and an unreadable thread is 404 before any state is read, matching the rest of the threads API.
- `Server-Timing` reports `thread_get` and `get_state` phases so the per-message cost shows up in the same perf tooling as the state fetch.

-------------------------
- main
    - https://github.com/langchain-ai/open-swe/pull/2803
      - **feat(threads): fetch a single thread message by id** :point_left:
        - https://github.com/langchain-ai/open-swe/pull/2805


johannes117:

PR #2805, opened 2026-09-15
feat(ui): paint the transcript from the trimmed thread state

johannes117:
## Summary

Last PR in the stack: the web UI paints from the trimmed state view and hydrates the rest behind the first frame.

- `createDashboardClient` routes the SDK's hydration read (`GET …/threads/:id/state`) to `?view=trimmed`, so the initial payload is the conversation text and tool-call names rather than every tool output and screenshot. The desktop's local client is untouched.
- A zustand `messageContentStore` caches full message contents per thread. Right after the trimmed transcript paints, one deferred read of the full state fills the cache for every message the view blanked. Expanding a stubbed row before that lands calls the per-message endpoint for just that message and shows "Loading output…" in the meantime; blanked images show a placeholder and fetch the same way.
- `streamMessagesToUi` understands the `open_swe_stub` markers and consults the cache, so once content arrives the transcript rebuilds with the real output and the pending state disappears on its own.
- The head warmup script prefetches the exact trimmed URL so the in-flight response still hands over to the SDK's read.

On the 12 MB reference thread this takes the paint-critical payload from 12 MB to a few hundred kilobytes, with the same bytes still loaded in the background for expands.

-------------------------
- main
    - https://github.com/langchain-ai/open-swe/pull/2803
      - https://github.com/langchain-ai/open-swe/pull/2804
        - **feat(ui): paint the transcript from the trimmed thread state** :point_left:


johannes117:

PR #2807, opened 2026-09-15
feat(workspaces): introduce workspaces backed by PostgreSQL with routing and per-workspace settings

langchain-infra:
First of a two-PR stack implementing OEP-0003 (workspaces): the backend, stored in PostgreSQL from the start. The dashboard follows in the second PR; until it lands, the existing Workspaces page keeps working against the new routes.

The branch is layered so it can be read commit by commit: `swagger.json` regenerated against `main` (the committed document had drifted after the route-relocation refactor), the OEP, the rename, and then one feature commit.

**Design.** `oeps/0003-workspaces.md` (draft). A workspace owns one or more repositories (each repository belongs to exactly one workspace), zero or more Slack channels, its MCP connections, and its team settings, and absorbs the former environment (prompt, snapshot, scripts, sizing, refresh). Everyone signed in sees every workspace in this version; the OEP records the follow-ups.

**Rename.** `agent/environments` -> `agent/workspaces` and `/dashboard/api/environments*` -> `/workspaces*`; the Environments page, composer picker, and route files; route handlers, admin tools (`list_workspaces`, `publish_workspace`, `refresh_workspace_start`, `delete_workspace`), their results and prompts; the system prompt sections and the bundled skill; the refresh module, its scheduler task, and background-task handles (the previous task name and `env-` prefix are still accepted); and the config variables (`WORKSPACE_SNAPSHOT_PREFIX`, `WORKSPACE_*_TIMEOUT_SECONDS`, previous names kept as aliases). Snapshot names keep the `<prefix>-environment-<slug>` form and the script root path stays, so existing images remain valid. The rename commits leave the Store namespace alone; the storage move below imports it.

**Storage.** Migration 0013 adds `workspace`, `workspace_repository` (primary key on `repository_id`, referencing the `repository` table, so the database enforces single ownership) and `workspace_slack_channel` (primary key on `channel_id`); 0014 adds CHECK constraints. `WorkspaceStore` writes through `agent.database.postgres.session()` in one transaction per write; conflicts are translated by constraint name and answer 409. LangGraph Store records under `["workspaces"]` or the legacy `["environments"]` namespace are imported once at startup and removed; until an import has succeeded on an empty table the GitHub route answers 503 rather than treating every repository as unowned. A record the import cannot bring over stays in the Store for the next boot, the import is not marked complete, and GitHub deliveries for the repositories it names are answered 503 as well.

**Routing.** `agent/workspaces/routing.py` resolves a workspace in the OEP's order: existing thread (a GitHub PR thread keeps the workspace it started in even if its repository has since moved), `workspace:`/`env:` tag, repository owner, Slack channel owner (which outranks a repository that only came from a team default), the user's default, then `default`. `OPEN_SWE_UNASSIGNED_REPO_WORKSPACE` (`default` | `ignore`) governs GitHub events for repositories no workspace owns; only that route fails closed on lookup errors. Dashboard threads, Slack mentions, GitHub and Linear events, schedules, and sandbox selection stamp and read `workspace` (readers accept the legacy `environment` key; run `configurable` carries both for one release).

**Settings.** Team settings and every cross-run cache of them are keyed by workspace slug (`agent/dashboard/team_settings_cache.py`); `GET/PUT /team-settings?workspace=` and `GET /options?workspace=`. Workspace MCP connections live under `["workspace_mcps", <slug>]` with legacy records migrated into `default`; routes move to `/workspaces/{workspace}/mcps...`. Model defaults and the vision fallback resolve for the run's workspace.

Tests that need workspace rows use the isolated-schema `registry_db` fixture (`TEST_ANALYTICS_POSTGRES_URI`, documented in `docs/DEVELOPMENT.md`).


langchain-infra:
Restructured: the OpenAPI regeneration (#2812) and the complete environments-to-workspaces rename (#2813) now sit below this PR in the stack, so this PR carries only the backend feature. Its history was rewritten accordingly (force-pushed to 7121edd4d).

johannes117 (approved):

PR #2812, opened 2026-09-15
chore: regenerate swagger.json

langchain-infra:
Regenerates `swagger.json` from the app so it matches the routes on `main` again. The route-relocation refactor changed the tag on every dashboard operation, and the analytics and incidents routers were never added to the committed document.

No code changes. Bottom of the OEP-0003 (workspaces) stack so the feature PRs above it only carry their own OpenAPI changes.


langchain-infra:
Folded back into #2807: the stack stays at two PRs (backend, then dashboard). The commits are unchanged and sit at the bottom of that branch.

PR #2813, opened 2026-09-15
refactor: rename environments to workspaces

langchain-infra:
Second PR of the OEP-0003 (workspaces) stack: the rename, with no behavior change. `oeps/0003-workspaces.md` (draft) records the design; the workspace absorbs the former environment, so the environment components take the workspace name before the backend and dashboard PRs build on them.

**Renamed.** `agent/environments` -> `agent/workspaces` and `/dashboard/api/environments*` -> `/workspaces*`; the Environments dashboard page, composer picker, and route files; the route handlers, admin tools (`list_workspaces`, `publish_workspace`, `refresh_workspace_start`, `delete_workspace`), their results, and their prompts; the system prompt sections and the bundled skill; the refresh module, its scheduler task, and background-task handles; the config variables (`WORKSPACE_SNAPSHOT_PREFIX`, `WORKSPACE_*_TIMEOUT_SECONDS`); and the e2e fixtures.

**Kept on purpose.** The `environment` key in run configuration and thread metadata, which the backend PR migrates with a compatibility window. Snapshot names (`<prefix>-environment-<slug>`) and the script root path inside images, so existing images stay valid. The previous config variable names as aliases. The previous cron task name and task-handle prefix, which the scheduler and the refresh module still accept.


langchain-infra:
Folded back into #2807: the stack stays at two PRs (backend, then dashboard). The commits are unchanged and sit at the bottom of that branch.

PR #2817, opened 2026-09-15
fix(analytics): restore PR model attribution

mdrxy:
### Summary

- carry the opening invocation's resolved configured model into PR analytics, including inherited preferences/defaults and Auto runs
- keep routed, fallback, and subagent models out of opening-model attribution
- reconcile unavailable PR attribution from the exact recorded opening run regardless of run/PR event arrival order
- when GitHub rejects PR creation with HTTP 422 because that branch already has an open PR, continue refreshing the existing PR's stats and thread metadata without emitting another `pr.opened` event; otherwise a later run that merely encounters the PR could be mistaken for its opening run and overwrite the intended model association

### Migration `0013`

Fixing new `pr.opened` events does not repair rows already stored as unavailable: event IDs are deduplicated and the PR projection previously kept its first write, so replaying the same events cannot fill the missing model. Migration `0013_repair_pr_model_attribution.py` performs the one-time projection repair during the normal database migration process. It updates only unavailable rows whose stored `opening_run_id` exactly matches a run in the same workspace with a recorded non-null configured model, and it is safe to repeat because repaired rows no longer satisfy that condition.

It preserves existing valid attribution, PR outcomes, timestamps, workspace isolation, and reporting cutover metadata. Rows without an opening-run ID, without that run's configured model, or otherwise lacking a trustworthy exact association remain unavailable; no claim is made that all currently unavailable production rows are recoverable. No operator command or production repair is included.

Made by [Open SWE](https://openswe.vercel.app/agents/9998584e-a4a7-5783-a10b-a97480b3c55f) · openai:gpt-5.6-sol (medium)


johannes117 (approved):

agent/tools/open_pull_request.py
```diff
@@ -602,13 +603,14 @@ async def _record_pr_telemetry(
             created_at=details.get("created_at") or pr.get("created_at"),
             merged_at=details.get("merged_at") or pr.get("merged_at"),
             invocation_id=cfg.invocation_id,
-            model_id=cfg.agent_model_id,
+            model_id=cfg.resolved_agent_model_id,
```
eyurtsev, line 606: potentially incorrect due to model routing

PR #2824, opened 2026-09-16
feat(settings): tier workspace settings and MCP connections by instance, workspace, and user

langchain-infra:
Second backend PR of the OEP-0003 stack, ahead of the dashboard PR: settings and MCP connections resolve by tier instead of each workspace holding an isolated copy, and the pre-workspaces "team settings" name goes away.

- **Settings tiers.** Workspace settings (formerly team settings) now resolve from the hardcoded defaults, then the instance record, then a workspace's overrides. The instance record is the pre-workspaces `team_settings` record under its original key, so nothing migrates; a workspace holds only the fields an admin set (`workspace_settings/<slug>`), and anything unset inherits. Per-user profile settings and a thread's `configurable` keep layering on top, so model config is instance → workspace → user → thread.
- **API.** `GET`/`PUT /dashboard/api/settings` is the instance record; `/team-settings` still answers, hidden from the schema, until the dashboard PR stops calling it (the `workspace` query parameter from #2807 is gone). `GET`/`PUT /dashboard/api/workspaces/{slug}/settings` returns the effective values and the overrides separately; `PUT` replaces the overrides and a `null` field means inherit. Every field of the update model (`WorkspaceSettingsUpdate`, formerly `TeamSettingsUpdate`) is optional now, which keeps the instance defaults unchanged; the Fable rules run at write time against the toggle a record resolves to, so an inherited toggle is honoured. Deleting a workspace clears its overrides, and records #2807 wrote for other workspaces keep applying until they are saved again.
- **MCP tiers.** A new instance scope (`/dashboard/api/mcps`, admin) is loaded for every run before the workspace's connections and the user's own; a later tier's same-named connection replaces the earlier one, the way user connections already replaced workspace ones. The Linear notification loader includes instance connections too. The connections configured before workspaces existed are this tier: #2807 had moved them into the `default` workspace, and this PR adopts them into the instance scope instead, so they keep applying everywhere as they always did. (If #2807 was already running and someone opened the default workspace's MCP list, those records were moved into `default` at that point and would need recreating at the instance level.)
- **Ownership still holds.** A workspace never uses a default repository another workspace owns, even one it inherited from the instance record.
- **Rename and one entry point.** `agent/dashboard/team_settings.py` and its cache now say workspace settings, and `get_workspace_settings()` returns a `WorkspaceSettings` mapping whose properties and methods (`default_model(role)`, `agent_routing_models`, `fable_enabled`, `default_repo`, …) replace the former free getters; the cache module keeps a single `cached_workspace_settings()`. The Store namespace is unchanged.
- The OEP gains a "Settings tiers" section and retires the open question about inheritance.

Follow-up in the dashboard PR: the Admin page edits the instance tier, and each workspace page shows the same settings with inherit controls plus the instance MCPs it inherits.


agent/dashboard/options_routes.py
```diff
@@ -10,10 +10,10 @@
     gate_fable_model,
     models_with_profile_context_windows,
 )
-from agent.dashboard.team_settings import (
-    get_team_default_model,
-    get_team_default_subagent_model,
-    get_team_fable_enabled,
+from agent.dashboard.workspace_settings import (
+    get_workspace_default_model,
```
ramon-langchain: its kind of weird to have all these free functions I would expect one "get settings" function or class method

docs/INSTALLATION.md
```diff
@@ -111,7 +111,7 @@ Open SWE calls models through [LangChain](https://python.langchain.com/) chat mo
...
+**Which model runs.** The deployment default is `anthropic:claude-opus-5` when only an Anthropic key is configured and `openai:gpt-5.6-sol` otherwise, at `medium` reasoning effort; override it with `LLM_MODEL_ID` (`provider:model`) and `LLM_REASONING_EFFORT` (`low`, `medium`, `high`, `max`), and name a `LLM_FALLBACK_MODEL_ID` for when the primary provider fails. Admins set the instance default, which each workspace can override, under **Admin → Global defaults**, and each user can pick their own model and effort under **My settings**; the supported list lives in `agent/dashboard/options.py`. Model ids and their providers are described in [CUSTOMIZATION.md](CUSTOMIZATION.md).
```
ramon-langchain: we probably do not need to document these frequently changing things here (model names)

ramon-langchain (approved):

agent/dashboard/options_routes.py
langchain-infra: fixed

langchain-infra: yea best to have a single resolver

docs/INSTALLATION.md
langchain-infra: fixed

PR #2838, opened 2026-09-16
feat(analytics): record PR opened and merged as trace feedback

eyurtsev:
## Summary

Add langsmith feedback keeps for pr open and pr merged.

Make the keys easy to aggregate. 

If an agent opens multiple PRs, this aggregation fails -- this is fine, it's not the dominant use case and we just want to have a simple way to analyze the data.


agent/webhooks/common.py
```diff
@@ -1340,20 +1342,28 @@ def _pr_state_from_payload(payload: dict[str, Any]) -> str | None:
...
+                key,
+                score=1.0,
+                comment=comment,
+                source_info={
+                    "source": "github_pr_merged",
+                    "thread_id": thread_id,
+                    "pr_url": pr_url,
+                },
+            )
+        except Exception:  # noqa: BLE001
```
eyurtsev: why all the extra exception handling?

agent/webhooks/common.py
```diff
@@ -1340,20 +1342,28 @@ def _pr_state_from_payload(payload: dict[str, Any]) -> str | None:
...
-    except Exception:  # noqa: BLE001
-        logger.debug("Failed to record merged PR feedback for thread %s", thread_id, exc_info=True)
+    # Two keys: the per-URL one keeps every PR on a thread distinguishable, while
+    # the bare ``pr_merged`` stays averageable across threads in LangSmith.
+    for key, comment in (
+        (f"github_pr_merged:{pr_url}", f"Agent-authored pull request merged: {pr_url}"),
+        (PR_MERGED_FEEDBACK_KEY, pr_url),
+    ):
+        try:
+            await create_langsmith_thread_feedback(
```
eyurtsev: let's do asyncio.gather here

tests/github/test_open_pull_request.py
```diff
@@ -987,6 +987,53 @@ async def fake_bot() -> str | None:
...
+    ("record_opening", "expected_keys"),
+    [(True, ["pr_opened"]), (False, [])],
+)
+async def test_record_pr_telemetry_records_pr_opened_feedback(
+    monkeypatch: pytest.MonkeyPatch,
+    record_opening: bool,
+    expected_keys: list[str],
+) -> None:
+    """``pr_opened`` is the averageable denominator, so it fires once per new PR."""
+    _set_config(monkeypatch, {"source": "slack", "thread_id": "t1", "github_login": "octo"})
```
eyurtsev, line 1000: theese tests kinda suck, we can't do better than moneky patching? is that the general structure of the tests currently?

agent/tools/open_pull_request.py
```diff
@@ -550,6 +551,27 @@ async def _thread_pull_requests(thread_id: str) -> list[dict[str, Any]]:
     ]
 
 
+# Thread-scoped LangSmith feedback key; unsuffixed so it averages across threads.
+PR_OPENED_FEEDBACK_KEY = "pr_opened"
+
+
+async def _record_pr_opened_feedback(thread_id: str, *, pr_url: str) -> None:
```
eyurtsev, line 557: Please inline a doc-string here. You are writing code for people

agent/tools/open_pull_request.py
```diff
@@ -550,6 +551,27 @@ async def _thread_pull_requests(thread_id: str) -> list[dict[str, Any]]:
...
+            PR_OPENED_FEEDBACK_KEY,
+            score=1.0,
+            comment=pr_url,
+            source_info={
+                "source": "open_pull_request",
+                "thread_id": thread_id,
+                "pr_url": pr_url,
+            },
+        )
+    except Exception:  # noqa: BLE001
```
eyurtsev: kill the dumb exception catching

agent/webhooks/common.py
```diff
@@ -1340,20 +1342,28 @@ def _pr_state_from_payload(payload: dict[str, Any]) -> str | None:
 
 
 async def _record_pr_merge_feedback(thread_id: str, *, pr_url: str) -> None:
```
eyurtsev, line 1343: Inline documentation in python style plz

agent/tools/open_pull_request.py
```diff
@@ -550,6 +551,28 @@ async def _thread_pull_requests(thread_id: str) -> list[dict[str, Any]]:
...
 
 
+PR_OPENED_FEEDBACK_KEY = "pr_opened"
+
+
+async def _record_pr_opened_feedback(thread_id: str, *, pr_url: str) -> None:
+    """Record ``pr_opened`` feedback on the thread's trace.
+
+    The key is constant across threads so LangSmith can count PR-producing
+    threads against ``pr_merged``. Presence is the signal; the score is always 1.
```
eyurtsev: Could we leave a design note here, just say that this will probably not work if the agent make smultiple PRs (race conditions etc), but for now we're OK normalizing and just asuming simple case to make analysis easier

presencce is the "signal" really? could you make htis a bit more human written

agent/webhooks/common.py
eyurtsev: Removed. `create_langsmith_thread_feedback` already catches its own failures and returns a bool, so the wrapper only logged the same error a second time at `debug`.

agent/webhooks/common.py
eyurtsev: Done — both merge writes now go through `asyncio.gather`.

agent/tools/open_pull_request.py
eyurtsev: Added a Google-style docstring.

agent/tools/open_pull_request.py
eyurtsev: Gone, same reason as the other one — the helper reports its own failures.

agent/webhooks/common.py
eyurtsev: Added a Google-style docstring here too.

agent/tools/open_pull_request.py
eyurtsev: Both done. Dropped the "presence is the signal" phrasing and added the design note: a thread that opens several PRs collapses to one row and the writes would race, which we are accepting so the aggregate stays simple — with a pointer to the per-URL keys for the detail.

tests/github/test_open_pull_request.py
eyurtsev: Checked, and monkeypatching is the existing structure: 1763 `monkeypatch.setattr` calls across 183 of 267 test files, vs 106 uses of `patch(`. So this matches the surrounding tests rather than being worse than them.

The reason this one needs four patches is that `_record_pr_telemetry` reaches four module-global collaborators plus does its own HTTP. Making those injectable would be the real fix, but that is a production refactor to suit a test, so I left it out of this PR. Happy to take it on separately if you want it.

agent/webhooks/common.py
```diff
@@ -1340,20 +1341,32 @@ def _pr_state_from_payload(payload: dict[str, Any]) -> str | None:
...
+    Args:
+        thread_id: LangGraph thread the PR was opened from.
+        pr_url: Stored as the feedback comment.
+    """
+    source_info = {"source": "github_pr_merged", "thread_id": thread_id, "pr_url": pr_url}
+    await asyncio.gather(
+        create_langsmith_thread_feedback(
             thread_id,
             f"github_pr_merged:{pr_url}",
             score=1.0,
```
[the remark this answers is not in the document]
eyurtsev: yeah we're catching exceptions that shouldn't occur willy nilly, the feedback function is already all w/ exception handling. did you read the code carefully?

mdrxy (approved):

PR #2853, opened 2026-09-16
feat(routing): A/B test adaptive model routing

eyurtsev:
- Gate adaptive routing with a 50% die throw on the first model invocation
- Persist the experiment outcome so the classifier is never attempted later in the run
- Emit routing events and metadata only when the routing model actually made a decision
- Keep control-group invocations on the balanced default model

Made by [Open SWE](https://openswe.vercel.app/agents/c86fbb4e-7ae6-5ed7-9d9f-5d57dce3daba) · openai:gpt-5.6-sol (medium)


agent/server.py
```diff
@@ -1103,10 +1111,17 @@ async def reconnect_backend(
         model_id, profile_effort = routing_defaults["fast"]
         subagent_model_id, subagent_effort = routing_defaults["fast"]
 
-    config["metadata"] = {
-        **(config.get("metadata") or {}),
-        "model_routing_applied": adaptive_model_routing,
-    }
+    metadata = config.get("metadata") or {}
+    metadata.pop("model_routing_applied", None)
```
eyurtsev: we are we mutating metadata? let's keep the key as it was before.
We're going to introduce a new key called "model_routing_mode"

agent/server.py
```diff
@@ -1315,7 +1330,7 @@ async def reconnect_backend(
     agent_backend = CompositeBackend(default=backend, routes=skill_routes)
     main_model = _make_model_or_defer(model_id, use_gateway=use_gateway, **model_kwargs)
     model_selection: ModelSelectionMiddleware | None = None
-    if adaptive_model_routing:
+    if adaptive_model_routing and model_routing_mode is not None:
```
eyurtsev: do we need to make any changes here? we can just keep adaptive_model_routing path as before??

The model_routing_mode is just for a quick A/B test, we don't need to change everything

agent/server.py
```diff
@@ -1103,10 +1111,17 @@ async def reconnect_backend(
...
-        **(config.get("metadata") or {}),
-        "model_routing_applied": adaptive_model_routing,
-    }
+    metadata = config.get("metadata") or {}
+    metadata.pop("model_routing_applied", None)
+    if adaptive_model_routing:
+        model_routing_mode = _model_routing_mode(thread_id)
+        metadata["model_routing_mode"] = model_routing_mode
+        if model_routing_mode == "performant":
+            model_id, profile_effort = routing_defaults["performance"]
```
eyurtsev: Why do we need to unpack this information?

agent/middleware/model_selection.py
```diff
@@ -67,6 +68,21 @@ def fast_alt_bucket(thread_id: str | None) -> float:
     return int(digest[:8], 16) / float(0xFFFF_FFFF)
 
 
+def _mark_model_routing_applied() -> None:
```
eyurtsev: why is this change necessary?

agent/middleware/model_selection.py
```diff
@@ -17,6 +17,7 @@
 logger = logging.getLogger(__name__)
 
 Route = Literal["fast", "fast_alt", "balanced", "performance"]
```
sydney-runkle, line 19: let's not have fast alt?

agent/middleware/model_selection.py
```diff
@@ -122,6 +125,8 @@ async def select_route(
             return "performance"
         if model_route := state.get("model_route"):
             return model_route
+        if self._routing_mode == "performant":
+            return "performance"
```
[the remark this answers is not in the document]
eyurtsev: are you an idiot?

mdrxy (approved):

agent/middleware/model_selection.py
eyurtsev: will follow up in a few to remove fast alt

sydney-runkle (commented):
awesome

PR #2870, opened 2026-09-16
feat(slack): give /oswe channel context and a thread per invocation

ramon-langchain:
`/oswe` ran with nothing but the text of the slash command. Any request that pointed at the surrounding conversation — "this error", "that PR", "the thing we just discussed" — had nothing to resolve against, and while the prompt told the agent to go read the thread itself with `slack_read_thread_messages`, it never gave it a channel id to do it with.

## Channel context in the prompt

- The run prompt carries the channel id and name, plus a snapshot of the channel's recent top-level messages: 30 messages, capped at 5k tokens and trimmed oldest-first.
- The whole input message is wrapped in `<markdown>`. The snapshot is labelled as background, explicitly not as instructions.
- Channel history contains no thread replies, so a message that has them is marked `[thread: N replies, thread_ts=...]` for `slack_read_thread_messages` to follow.
- Join/leave/topic noise is dropped.

## A tool for reading further back

`slack_read_channel_messages(channel_id, limit)` reads past what the snapshot holds. Two things bound it:

- **The thread must be private**, since channel history lands in the transcript and everyone who can read the thread reads it. The factory resolves visibility once and does not register the tool at all otherwise, alongside the rest of the tool gating. `/oswe` threads are private by construction.
- **The channel must be public.** History is fetched with the deployment's bot token, which says nothing about who is asking, so a private channel, a DM, or one shared with another organization would leak its membership. This one is checked in the tool, because it depends on the argument rather than on the run, and it also lives in `fetch_slack_channel_messages` so the `/oswe` snapshot is bound by it too.

Existing manifest scopes already cover the reads, so no Slack app reinstall.

## One thread per invocation

Slash commands used to share a single thread per (channel, user). Unrelated questions accumulated in one history, and a command sent while another was still running was queued behind it instead of answered — which compounds badly now that every command also carries a channel snapshot.

Each invocation now gets its own thread, keyed by Slack's `trigger_id`, so two questions never share a thread and a redelivery of the same command still resolves back to the first. Threads stay private and `unlisted`, so one-off questions still do not fill anyone's thread list.

Related: `options` was silently dropped on the ephemeral `/oswe` path. Slack cannot route a click on an ephemeral message back to the run, so it is now refused with a hint to ask for the choice in the reply text instead.

## Test Plan

- [x] `make test` (2977 passed, 285 skipped)
- [x] `make lint`, `make typecheck`
- [x] New coverage: channel context reaches the prompt inside `<markdown>` with thread markers; the 5k budget holds and keeps the newest messages; the tool marks threads, skips joins, and refuses a private channel, and is registered only in a private thread; `options` is refused on the ephemeral path; two invocations get different threads while the same `trigger_id` gets the same one; the channel read survives the dashboard and subagent gates


langchain-infra (approved):

agent/slack/tools/read_channel_messages.py
```diff
@@ -1,19 +1,56 @@
...
+logger = logging.getLogger(__name__)
 
 DEFAULT_CHANNEL_MESSAGE_LIMIT = 30
 
+_NOT_PUBLIC = (
+    "Only public channels can be read this way. That channel is private, a direct message, "
+    "or shared with another organization, and reading it here would bypass its membership. "
+    "Ask the person for what you need instead."
+)
+_NOT_PRIVATE_THREAD = (
```
ramon-langchain: why is this needed? just make the tool unavailable in non-private threads?

PR #2890, opened 2026-09-17
feat: support explicit Slack model switches

ramon-langchain:
- Recognize exact, case-sensitive `/model:fast` and `/model:perf` tokens anywhere in the current Slack message; the last token wins and tokens are removed from the request.
- Resolve workspace-configured fast/performance models, override existing choices, and persist the explicit thread selection. DMs honor the current message’s switch while retaining their existing default-model behavior on subsequent messages.
- Record `slack_model_switch`, `slack_requested_model`, and `slack_requested_effort` in run/trace metadata, preserving existing image fallback behavior.

Made by [Open SWE](https://openswe.vercel.app/agents/a8b9c70c-1902-5515-af70-ce269df61fac) · openai:gpt-6-astra (low)


agent/slack/webhook.py
```diff
@@ -969,6 +980,14 @@ async def _process_slack_mention_impl(
     else:
         thread_workspace = await common.get_thread_workspace(thread_id)
 
+    model_switch_choice: tuple[str, str] | None = None
+    if model_switch:
+        routing_tier = "fast" if model_switch == "fast" else "performance"
```
ramon-langchain: do not default to performance, post an error message back to slack if unrecognized

PR #2896, opened 2026-09-17
feat(expedited-review): request from any thread and post to a named channel

ramon-langchain:
`expedite_pr_approval` required the run to be a Slack thread, so a dashboard thread could not ask for an expedited review, and in practice only the thread's own PR was nominated.

- New `channel` argument: a Slack channel name (`eng-reviews`, `#eng-reviews`, `<#C…|name>`) or id. The tool posts a root message there and threads the approval card under it. Public channels are joined on demand; private ones report that the bot needs an invite.
- Without `channel`, the current Slack thread is used as before. A `/oswe` question (channel, no thread) posts a new thread in that channel.
- Any open, eligible PR can be nominated, not only ones Open SWE authored. A PR without a row gets one from the GitHub payload so the card shows title and author and the author-vote rule applies.
- The tool is now gated on a configured Slack bot token and the workspace setting, not on Slack source context, so it appears in dashboard threads.


agent/tools/expedite_pr_approval.py
```diff
@@ -16,19 +16,59 @@
...
+        channel_id = cfg.slack_thread.channel_id.strip()
+    return channel_id, thread_ts
+
+
+async def _post_root_message(
+    channel_id: str, pr_ref: GitHubPrRef, title: str
+) -> tuple[str | None, str | None]:
+    """Open a thread in ``channel_id`` for the card; joins the channel if needed."""
+    label = f"{pr_ref.owner}/{pr_ref.repo}#{pr_ref.number}"
+    text = (
```
ramon-langchain: use a md file

mdrxy (approved):

ramon-langchain:
Replying here rather than inline: this account already has a pending (unsubmitted) review on the PR, and GitHub refuses a second pending review from the same user, which is what an inline reply would create. Both findings from **Open SWE Review** are addressed in f205dde.

### 🔵 Explicit same-channel requests reuse the current thread ([comment](https://github.com/langchain-ai/open-swe/pull/2896#discussion_r4038148931))

Correct, and the condition contradicted the documented contract, so it is gone. An explicit `channel` now always clears `thread_ts` and gets a thread of its own, including when it resolves to the channel the run is already talking in.

I originally wrote the guard to keep a redundant same-channel request threaded with the surrounding conversation, but that behavior was never stated anywhere, while the tool description does promise "pass `channel` ... to post it as a new thread in that channel instead". Honoring the explicit request is simpler and is what a caller would expect.

### 🟡 Check active approval before posting a new root ([comment](https://github.com/langchain-ai/open-swe/pull/2896#discussion_r4038148926))

Real bug, and the consequence was worse than a misreported channel: the orphaned root message tells everyone in that channel an approval card is coming, and nothing ever arrives under it.

The active approval is now resolved before anything is posted. When one exists for the same head SHA, the tool adopts its stored Slack location instead of posting a new root, so no message is left promising a card that belongs to another thread. The response reports where the card actually lives rather than the channel just requested, and `next` tells the agent to cancel and ask again to move it.

I chose report-and-refuse over relocating the approval. Relocating means retracting an already-posted card and re-posting it while votes may be in flight, which is a lot of moving parts for something the agent can resolve with `action="cancel"`.

One case the finding did not mention, covered as well: an active same-SHA approval whose stored location is empty. That is degenerate, but falling through would orphan a message there too, so a reused approval never posts a root regardless of whether it has a location.

Verified locally: `tests/expedited_review`, `tests/slack` and `tests/tools` pass, 703 passed and 25 skipped. Lint and `ty check` clean.

— Claude

langchain-infra (approved):

PR #2930, opened 2026-09-17
feat(transcript): append-only thread transcript event log

johannes117:
## Problem

The dashboard transcript hydrates from LangGraph `get_state` and follows runs through `useStream`. In production `get_state` carries a fixed 0.5–4 s cost regardless of thread size, the SDK fires extra `/history` fetches after every hydrate, and runs started outside the dashboard need `stream_resumable` plus a protocol marker just to be observable. Tuning the read path (skeleton state, transcript view, stream pooling) did not get us there.

## Approach

Adopt an append-only event log pattern, in our own Postgres, for **new cloud threads only**. Existing cloud threads without a transcript keep reading LangGraph state through a plain `useStream`; desktop-local threads do the same and are otherwise untouched. The multi-instance stream pool is removed.

**Storage** (`agent/transcript/`, migration `0023`): `thread`, `thread_event`, `thread_command_receipt`, `thread_turn`, `thread_message`, `thread_tool_call`, `thread_attachment`. One transaction appends events, applies SQL projections, writes attachment bytes, upserts receipts and `pg_notify`s. Versions are per thread and gapless, assigned under `pg_advisory_xact_lock(hashtext(thread_id))`. Messages and tool calls are keyed by `(thread_id, id)`, so a client-supplied id can never reach another thread; a settled turn is immutable. `thread_turn` carries `base_commit`/`head_commit`/`changed_files` for a future "last turn" picker (not populated yet).

**Emit path** (`agent/middleware/transcript.py`, wired into the main agent and the general-purpose subagent): turn start/end, paragraph-batched `message.appended` fragments (400 ms, markdown boundary, 24k hard flush) from a callback handler on the run's callback manager, canonical `message.completed` with token usage, `tool.started`/`tool.completed` with subagent namespaces via a context var, and `run.notice`. Writes go through a per-run queue so the model stream never waits on Postgres; under backpressure only streamed fragments are shed, which `message.completed` repairs. `run.start` enrichment appends `thread.created` and `turn.requested`, stores image attachments and stamps `metadata.transcript = "v2"`. The middleware is inert when Postgres is not configured. The completion webhook and the cancel endpoints settle turns as a safety net; a finished run can never settle the next waiting turn.

**Read API**: `GET /dashboard/api/threads/{id}/transcript` (newest 40 turns from the projections plus an opaque keyset `older_cursor`), `.../transcript/turns?before=<cursor>` (older pages), `.../transcript/events?after=N` (SSE: replay, `snapshot` fallback over 1000 events / 8 MiB, `synchronized` marker, `deleted` terminator, LISTEN/NOTIFY fan-out with resync after reconnect), `.../transcript/tool-calls/{id}/output`, `.../transcript/attachments/{id}`. Snapshot-style reads run under `REPEATABLE READ` so the projections and the returned version describe one committed state. Authorization reuses `_assert_thread_readable` against the mirrored thread metadata, which every write to an authorization or title key updates through `thread.meta_updated`; deleting a thread deletes its transcript. The read path never calls LangGraph.

**UI**: a `ThreadSource` seam with two implementations (`useStream` adapter, transcript). The thread page branches on `thread.transcript === "v2"` after the detail resolves, so a transcript thread never mounts an SDK stream. The reducer emits the existing `Message`/`Chunk` types with per-turn structural sharing, so everything from `renderItems.ts` down is unchanged; older turns load when the user scrolls to the top with the scroll position preserved. New threads mint a thread id client-side, post `run.start` to the commands endpoint, read the transcript flag from the response and navigate; Stop during creation cancels the run.

## Out of scope

- Desktop-local threads (separate refactor) and the reviewer chat stay on `useStream`.
- A skeleton outline with random-access hydration and list virtualization (Devin-style) can layer on the same cursor design later if long threads warrant it.


johannes117:
Superseded by a stacked set for review: #2993 (storage and read API) → #2994 (record runs) → #2995 (UI thread-source seam) → #2996 (serve new threads from the transcript). This branch stays as the reference for the combined diff.

PR #2931, opened 2026-09-18
feat: migrate message queue to the SDK's server-backed enqueue strategy

thushanth-bengre-langchain:
## Summary

Migrates how open-swe queues messages sent while a run is active. Moved it off its custom LangGraph store queue and onto the SDK's server-backed enqueue multitask strategy, so those follow-ups become real, durable pending runs instead of client-side state.

## Changes

- `useSubmitAgentMessage` now submits every message via `stream.submit(..., { multitaskStrategy: "enqueue" })` instead of `agentApi.queueMessage()`; enqueue failures surface through the per-submit `onError` callback.
- `AgentStreamProvider` opts into the SDK's `serverQueue: true` capability for both cloud and local transports.
- Removed the local-only queue implementation (`localMessageQueue`, `useLocalPromptQueue`); the desktop composer now reads pending entries from `useSubmissionQueue(stream)`.
- `AgentThreadView`'s queue display merges SDK-tracked entries with `thread.queuedMessages`, which still catches producers that bypass `stream.submit()` (Slack/incident relays, the `_send_message` agent tool).
- Removed the now-unused `queueMessage` API method and `ThreadMessageRequest` type.
- Temporary `pnpm-workspace.yaml`/lockfile overrides point `@langchain/langgraph-sdk`/`@langchain/react` at a local sibling `langgraphjs` checkout (flagged with a TODO) until [langchain-ai/langgraphjs#2848](https://github.com/langchain-ai/langgraphjs/pull/2848) ships.


pnpm-workspace.yaml
```diff
@@ -15,6 +15,12 @@ minimumReleaseAgeExclude:
   - '@langchain/react@1.1.0'
 
 overrides:
+  # TODO: local-only, points at a sibling langgraphjs checkout for the
+  # unreleased serverQueue feature. Breaks install for anyone without that
+  # checkout (CI, other contributors, Docker). Revert to real published
+  # versions once langchain-ai/langgraphjs#2848 ships.
+  '@langchain/langgraph-sdk': 'file:../langgraphjs/libs/sdk'
+  '@langchain/react': 'file:../langgraphjs/libs/sdk-react'
```
thushanth-bengre-langchain, line 23: Known. Temporary local-only override while langgraphjs #2848 is unpublished; will revert to published versions + regenerate the lockfile before merge.

agent/incidents/turns.py
```diff
@@ -137,13 +137,23 @@ async def cancel_active_runs(thread_id: str, *, keep_run_id: str = "") -> None:
...
     try:
         item = await store_client().store.get_item(("queue", thread_id), "pending_messages")
     except Exception:  # noqa: BLE001
-        return 0
+        item = None
     value = item.get("value") if isinstance(item, dict) else None
     messages = value.get("messages") if isinstance(value, dict) else None
-    return len(messages) if isinstance(messages, list) else 0
+    kv_count = len(messages) if isinstance(messages, list) else 0
+    return kv_count + len(await _runs(thread_id, "pending"))
```
hntrl, line 156: hmm do we have any way to expose this from sdk?

hntrl: like in theory after we move to useStream queueing, this store won't be populated with values anymore

agent/threads/proxy.py
```diff
@@ -294,6 +305,185 @@ async def proxy_dashboard_thread_history(
     return response.status_code, response.content, media_type
 
 
+async def proxy_dashboard_thread_runs_list(
```
hntrl, line 308: por que? why do we have to proxy? should be handled by LSD -- if it isn't then we should flag

ui/src/features/agents/components/messages/UserMessage.tsx
```diff
@@ -1,15 +1,33 @@
...
+  // takes precedence.
+  const optimisticStatus = useMessageMetadata(stream, message.id)
+    ?.optimisticStatus
+  const deliveryStatus =
+    message.deliveryStatus ??
+    (optimisticStatus === "pending"
+      ? "sending"
+      : optimisticStatus === "failed"
+        ? "failed"
+        : undefined)
```
hntrl: this ternary is kind of gnarly

ui/src/features/agents/components/AgentThreadView.tsx
```diff
@@ -232,18 +233,26 @@ export function AgentThreadView({
     () => ({ threadId: thread.id, running: thread.status === "running" }),
     [thread.id, thread.status]
   )
+  // `thread.queuedMessages` catches producers that bypass `stream.submit()`.
+  const queueEntries = useSubmissionQueue(stream).entries
```
hntrl, line 237: fire

ui/src/features/agents/lib/stream/AgentStreamProvider.tsx
```diff
@@ -84,6 +84,7 @@ function PooledStream({ entry }: { entry: StreamPoolEntry }) {
     assistantId: AGENT_ASSISTANT_ID,
     threadId: entry.threadId,
     fetch: dashboardFetch,
+    serverQueue: true,
```
hntrl: @/open-swe team reviewing this, we're going to rm this option

agent/incidents/turns.py
thushanth-bengre-langchain: We are moving the consumer messages to useStream. This one is for slack/incident/tool driven message queue.
So SDK cant expose them.

ui/src/features/agents/lib/stream/AgentStreamProvider.tsx
thushanth-bengre-langchain: It's now `queue:"server"`

agent/threads/proxy.py
thushanth-bengre-langchain: The browser never talks to langGraph directly. This proxy is used by `history`/`cancel`/`stream`/`events` already and has nothing to do with LSD. 
Bowser->dashboard backend-> langgraph
So we need browser->dashboard backend->runsAPI. That's why we need this proxy

ui/src/features/agents/components/messages/UserMessage.tsx
thushanth-bengre-langchain: Used a function with a `switch`

hntrl (commented):
stamping UI changes, I want to better understand the proxy changes

thushanth-bengre-langchain:
Superseded by #3163  and rebuilt on top of #3054's shipped queue/steer UI instead of duplicating it. 
Closing in favor of that.

PR #2932, opened 2026-09-18
fix(slack): render agent replies with native Markdown

pjrule:
Uses native Slack Markdown blocks for agent replies up to 12,000 characters, including language-tagged code fences, ephemeral answers, and option prompts. (Long ordinary replies fall back to mrkdwn.) 

Motivated by a bug I saw when using Open SWE with Slack: code blocks in mrkdwn mode are rendered with non-breaking space characters, which generally produce code blocks that cannot be copy-and-pasted.


langchain-infra (approved):

agent/slack/markdown.py
```diff
@@ -0,0 +1,89 @@
...
+_parser = MarkdownIt("commonmark", {"html": False}).enable("strikethrough")
+
+
+def markdown_to_mrkdwn(markdown: str) -> str:
+    """Convert standard Markdown to Slack mrkdwn."""
+    output: list[str] = []
+    list_stack: list[tuple[bool, int]] = []
+    quote_depth = 0
+    item_prefix = ""
+    in_heading = False
```
pjrule, line 15: fyi I am not reading or testing this carefully, since this is just rough fallback handling for a pretty extreme edge case. if we actually encounter rendering issues on this path, we can revisit

PR #2965, opened 2026-09-18
feat(middleware): guard delegated subagent PR creation

vladislav-nechakhin:
## Summary

Subagents compile into their own graphs, so the parent agent's middleware never wraps their tool calls. The general-purpose subagent already carries the workflow-push guard for this reason; this extends the same treatment to `PullRequestCreationGuardMiddleware` so delegated shell commands meet the same bar as the parent's.

## Changes

- Add `_subagent_guard_middleware`, mirroring the parent's shell guards and skipping the PR-creation guard on local desktop runs exactly as the parent stack does.
- Thread the guards into `_general_purpose_subagent` via a `guard_middleware` parameter (defaulting to empty, so existing callers are unchanged) and place them after `_subagent_middleware`, before conversation offloading.
- Extend the existing subagent-guard assembly test to cover both guards.

## Testing

- `pytest tests/agent/test_agent_assembly_context.py tests/agent/test_agent_incidents.py tests/middleware/test_dynamic_tools.py tests/middleware/test_model_call_timeout.py tests/github/test_pr_creation_guard.py tests/agent/test_workflow_push_guard.py` — 71 passed
- `ruff format --check`, `ruff check`, and `ty check` on both touched files — clean

Made by [Open SWE](https://github.com/langchain-ai/open-swe) · [view thread](https://openswe.vercel.app/agents/1ae14249-44a5-596f-ba58-fcd23a9878cb) · fireworks:accounts/fireworks/models/glm-5p3-flash (max)


agent/server.py
```diff
@@ -425,6 +425,17 @@ def _subagent_middleware(
     return middleware
 
 
+def _subagent_guard_middleware(local_run: bool) -> list[AgentMiddleware[Any, Any, Any]]:
+    """Shell guards mirroring the parent stack for delegated tool calls.
+
+    Local desktop runs skip the PR-creation guard the same way the parent does.
+    """
+    return cast(
```
ramon-langchain: this is an odd cast I think you can do early return to avoid

ramon-langchain (approved):

agent/server.py
vladislav-nechakhin: done

